PR #120, opened 2018-05-07
Disable pull ui while switching connection

barryceelen:
On slower sites and/or connections switching the connection on the pull list view can take some time. This may lead to the user interacting with the page, unaware that the context has not been switched yet.

This pull request disables the search input and list table after using the connection dropdown.


barryceelen:
@tlovett1 I've added the required styles to `admin-pull-table.css` and replaced setting styles on the elements via jQuery with setting a body class with vanilla JS.

tlovett1:
Awesome! For future reference, you can just use `classList` for adding the CSS class.

Issue #143, opened 2018-06-26
Preserve blocks when distributing posts created in Gutenberg to site with Gutenberg enabled


adamsilverstein:
### Preserving supported blocks
When distributing content from Site A to Site B that was created in Gutenberg (and where site B has Gutenberg enabled), we would like to be able to fork the distributed post on Site B and have its blocks editable.

### Remove the_content filtering that renders Gutenberg blocks
Removing this filter when distributing Gutenberg content will ensure raw blocks are distributed instead of rendered content.

### Matching Blocks Caveat
If a block type from site a is not supported on site b, the unknown block will wind up as a  Classic Block

adamsilverstein:
* add block A not on site B caveat to the readme

adamsilverstein:
I'm going to work on this issue.

PR #145, opened 2018-06-29
Network connections: use raw content when distributing gutenberg->gutenberg

adamsilverstein:
* doesn't check for specific post type gutenberg support, only whether gutenberg is enabled on the remote site

Partially fixes https://github.com/10up/distributor/issues/143

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_script:
   - composer install
 
 script:
-  - ./vendor/bin/phpcs
+  - ./vendor/bin/phpcs **/*.php *.php
   - ./vendor/bin/phpunit
 
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.6
   - 7.0
   - 7.2
 
--- a/includes/rest-api.php
+++ b/includes/rest-api.php
@@ -41,8 +41,6 @@ function filter_distributor_content( $prepared_post, $request ) {
 
 	if ( \Distributor\Utils\is_using_gutenberg() && isset( $request['distributor_raw_content'] ) ) {
 		$prepared_post->post_content = $request['distributor_raw_content'];
-					error_log( json_encode( $prepared_post, JSON_PRETTY_PRINT ) );
-
 	}
 	return $prepared_post;
 }
--- a/includes/rest-api.php
+++ b/includes/rest-api.php
@@ -1,4 +1,9 @@
 <?php
+/**
+ * REST API functionality
+ *
+ * @package  distributor
+ */
 
 namespace Distributor\RestApi;
 
@@ -50,9 +55,9 @@ function filter_distributor_content( $prepared_post, $request ) {
 /**
  * When an API push is being received, handle Distributor specific attributes
  *
- * @param WP_Post         $post
- * @param WP_REST_Request $request
- * @param bool            $update
+ * @param WP_Post         $post Post object.
+ * @param WP_REST_Request $request Request object.
+ * @param bool            $update Update or create.
  * @since 1.0
  */
 function process_distributor_attributes( $post, $request, $update ) {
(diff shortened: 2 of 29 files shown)


adamsilverstein:
* supports network connects and external connections. when both original and remote site support gutenberg, raw content is distributed
* stores the gutenberg active status of an external connection when you save it (resave to activate feature for existing connections)
* does not handle subscriptions yet, or post types that may or may not support gutenberg (eg rest not enabled)
* add a new 'dt_the_content' filter applied to final content so plugins can change the default behavior
* will work to get unit tests running

barryceelen:
@adamsilverstein Although I love the explicit nature of the `dt_the_content` filter, in the spirit of frugality one might argue that the content can already be filtered via the `dt_push_post_args` filter.

adamsilverstein:
good point @barryceelen - thanks for the feedback! I also noticed this and later removed the filters in https://github.com/10up/distributor/pull/145/commits/22732adf9696d95807befe2f46672ef4a2c8bec2

I did add a new filter to the subscription updates: https://github.com/10up/distributor/pull/145/files#diff-12f37ea7c389dfd5a6a70cc9992d8935R278

adamsilverstein:
* subscriptions are now supported, although when i tested updating a subscribed post using the default demo post, gutenberg lost the header block. 

It looked fine on the initial distribution, however after an update the header was corrupted. Comparing html, on initial distribution:


After an update:


Note the background style is missing, looking into why.

adamsilverstein:
Subscriptions now work correctly after https://github.com/10up/distributor/pull/145/commits/7b454858da33767d02d58003fb80b2dcd0d9501d. I'd still like to get unit tests running again, although their failure is not related to this PR. Also, the current PR is missing any post_type support - all post types are assumed to support G utenberg

includes/rest-api.php
```diff
@@ -20,11 +20,33 @@ function setup() {
...
+ * @param stdClass        $prepared_post An object representing a single post prepared.
+ * @param WP_REST_Request $request       Request object.
+ *
+ * @return stdClass $prepared_post The filtered post object.
+ */
+function filter_distributor_content( $prepared_post, $request ) {
+
+	if ( \Distributor\Utils\is_using_gutenberg() && isset( $request['distributor_raw_content'] ) ) {
+		$prepared_post->post_content = $request['distributor_raw_content'];
+					error_log( json_encode( $prepared_post, JSON_PRETTY_PRINT ) );
```
barryceelen: @adamsilverstein A lonely error_log is missing its parents and would like to be picked up at Småland!

tlovett1: @adamsilverstein I think this got left in by accident.

adamsilverstein: removed

adamsilverstein:
* added support for post types, via `gutenberg_can_edit_post_type`, needs testing
* realized i need to test/address pull as well

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.6
```
adamsilverstein: wp-mock doesn't work on 5.6

.travis.yml
```diff
@@ -14,7 +13,7 @@ before_script:
   - composer install
 
 script:
-  - ./vendor/bin/phpcs
+  - ./vendor/bin/phpcs **/*.php *.php
```
adamsilverstein: phpcs complained no files were being scanned

tests/php/includes/TestCase.php
```diff
@@ -2,7 +2,7 @@
 /**
  * We will extend this test case to make WP_Mock set up easier
  */
-class TestCase extends \PHPUnit_Framework_TestCase {
+class TestCase extends \PHPUnit\Framework\TestCase {
```
adamsilverstein: changed in latest PHPUnit

adamsilverstein:
* got tests and phpcs running, seeing many phpcs issues - are these already being addressed in a separate issue?

barryceelen:
There is no issue yet for WPCS compliance @adamsilverstein.

PR #148, opened 2018-07-03
unset post_parent before inserting pulled post

### LINKED ISSUE #146: Pulled post in internal network retains original post_parent ID
When pushing a post from an internal connection, the post array is explicitly defined, and the post_parent is not included. (See @jakemgold in #44.) But when pulling, the entire original post array is copied as-is:

`$post_props      = get_object_vars( $post );

/* ... */

foreach ( $post_props as $key => $value ) {
				$post_array[ $key ] = $value;
			}
`

This results in the external copies having a post_parent ID set to a post that doesn't exist on that network site -- or a completely different post on the external that happens to have the same ID.

post_parent needs to be unset before the pulled post is inserted.

sillybean:
Fixes #146

### DIFF
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -164,6 +164,10 @@ class NetworkSiteConnection extends Connection {
 			} else {
 				unset( $post_array['ID'] );
 			}
+			
+			if ( isset( $post_array['post_parent'] ) ) {
+				unset( $post_array['post_parent'] );
+			}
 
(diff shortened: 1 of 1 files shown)


adamsilverstein:
Hey @sillybean, thanks for the PR! 

Can we add/check for similar funcitonality to [`WordPressExternalConnection`](https://github.com/10up/distributor/blob/master/includes/classes/ExternalConnections/WordPressExternalConnection.php) and also in [https://github.com/10up/distributor/blob/master/includes/classes/API/SubscriptionsController.php#L220](`SubscriptionsController`), which handles updates of linked posts.

adamsilverstein:
Thanks @sillybean!

PR #149, opened 2018-07-03
Doc corrections: $new_post is ID, not obj or array

sillybean:
Doc blocks for push() and pull() specified $new_post as object or array. It should be the ID, which is what wp_insert_post() returns.


barryceelen:
Awesome! Perhaps we want to rename the variables to `$new_post_id` as well. The naming has been  throwing me off already before...

sillybean:
I agree; I keep expecting to have the full object at my disposal in those filters.

tlovett1:
@sillybean mind renaming the variables?

sillybean:
Let me know if I fouled up the tabs there.

tlovett1:
Awesome. Thank you!

Issue #44, opened 2017-12-10
Refine handling of hierarchical post types


jakemgold:
While hierarchical post types aren't the primary use case, there are plenty of conditions – from custom hierarchical post types, to staging pages, to generic about pages shared across sites – where this is useful. Right now, I don't think we "think" about this at all - it just ignores the hierarchy attribute.

Near term, I think we just need to refine the UX to make it _obvious_ that we're not going to do anything with hierarchy. I think that can just mean that – if the current post has a parent or child (in the literal sense, don't want to include things like media or drafts) – the "Distributor" toolbar menu explains / warns that parent / child relationships are not distributed.

I can see a potential future iteration that adds an option to "Include all child posts" (for a use case like a generic about page with subpages or learning module with several child items coming over).

I'm also not clear how - if at all - we handle showing hierarchy in the pull UI?

sillybean:
This is an issue when trying to integrate Distributor with Pressbooks, which organizes book chapters into parts using a post parent/child relationship. The distributed chapters end up orphaned, so they never appear in the book's table of contents.

It would help tremendously to add dt_original_post_parent to the post meta. From there, I can use various hooks to check whether there's a distributed post with the corresponding dt_original_post_id and manipulate the relationship as needed.

sillybean:
[Sample plugin to preserve parent/child relationships](https://gist.github.com/sillybean/0dd7f00a6ef686b938f7cf652e9746d5)

(the first two functions deal with Pressbooks CPTs in particular)

PR #150, opened 2018-07-03
Add 'dt_original_post_parent' to post meta

sillybean:
Store the original post_parent ID in post meta to make it easier to restore parent/child relationships after push/pull. See #44


includes/classes/ExternalConnections/WordPressExternalConnection.php
```diff
@@ -300,6 +300,7 @@ public function pull( $items ) {
 
 			$post_props = get_object_vars( $post );
 			$post_array = array();
+			$original_post_parent = wp_get_post_parent_id( $item_array['post_id'] );
```
barryceelen: I _think_ this is not referring to the correct post in this context. The original post is the `remote_post_id`. The WP_Post object for this post is stored in the `$post` variable on line 294, so we could use `$post->post_parent`.

includes/classes/ExternalConnections/WordPressExternalConnection.php
```diff
@@ -335,6 +336,7 @@ public function pull( $items ) {
 			update_post_meta( $new_post, 'dt_original_post_url', esc_url_raw( $post_array['link'] ) );
 			update_post_meta( $new_post, 'dt_original_site_name', sanitize_text_field( $post_array['original_site_name'] ) );
 			update_post_meta( $new_post, 'dt_original_site_url', sanitize_text_field( $post_array['original_site_url'] ) );
+			update_post_meta( $new_post, 'dt_original_post_parent', (int) $original_post_parent );
```
barryceelen: A check for `$original_post_parent !== 0` before adding the custom field would be good, so as not to store unnecessary data.  The field not being present would imply that the origin post did not have a parent.

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -176,6 +179,7 @@ public function pull( $items ) {
 				update_post_meta( $new_post, 'dt_original_blog_id', (int) $this->site->blog_id );
 				update_post_meta( $new_post, 'dt_syndicate_time', time() );
 				update_post_meta( $new_post, 'dt_original_post_url', esc_url_raw( $post->link ) );
+				update_post_meta( $new_post, 'dt_original_post_parent', (int) $original_post_parent );
```
barryceelen: Same here: only store `$original_post_parent` if it is not `0`.

includes/classes/ExternalConnections/WordPressExternalConnection.php
```diff
@@ -443,6 +445,7 @@ public function push( $post_id, $args = array() ) {
 			'distributor_original_site_name' => get_bloginfo( 'name' ),
 			'distributor_original_site_url'  => home_url(),
 			'distributor_original_post_url'  => get_permalink( $post_id ),
+			'distributor_original_post_parent'  => wp_get_post_parent_id( $post_id ),
```
barryceelen: Only save `$original_post_parent` if it is not 0.

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -85,6 +86,7 @@ public function push( $post_id, $args = array() ) {
 			update_post_meta( $new_post, 'dt_original_blog_id', (int) $original_blog_id );
 			update_post_meta( $new_post, 'dt_syndicate_time', time() );
 			update_post_meta( $new_post, 'dt_original_post_url', esc_url_raw( $original_post_url ) );
+			update_post_meta( $new_post, 'dt_original_post_parent', (int) $original_post_parent );
```
barryceelen: Same here: only store $original_post_parent if it is not 0.

barryceelen:
Awesome, love keeping the parent post info preserved when distributing!

tlovett1:
@sillybean do you have time to address @barryceelen's feedback? If not, we are happy to take over. Thank you!

sillybean:
Please do! I’m leaving again today for another ten days.

helen:
Could use some help getting this PR refreshed if we want it in 1.3.

Issue #83, opened 2018-03-09
Site B pushes post 1 to Site A. Post 1 shows in "New" pull list in Site A.


tlovett1:
We need to address this scenario.

PR #151, opened 2018-07-04
Remove deleted post from sync log

### LINKED ISSUE #93: When deleting a distributed post, it should become available for redistribution to that site again.
## Description

When deleting a distributed post, it should become available for redistribution to that site again. Currently that does not appear to be the case.

I have distributed my post 'Hello World' from site A to site B. I delete the post from site B. When I view the original post on site A and open the Distributor menu, the connection to site B is disabled - site A won't let me distribute the same post to site B twice, even though I have deleted the distributed copy from site B.

## Steps to reproduce

* Create a post on site A
* Distribute post to site B
* On site B, trash and delete the distributed post
* On site A, open the post again (and refresh)

**Expected behavior:**
The connection to site B should be enabled and I should be able to distribute the post to site B again.

**Actual behavior:**
The connection to site B remains disabled.

barryceelen:
Makes deleted posts available in the pull view on network connections. Fixes #93 for internal connections.

### DIFF
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -359,18 +359,18 @@ class NetworkSiteConnection extends Connection {
 		add_action( 'wp_ajax_dt_auth_check', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'auth_check' ) );
 		add_action( 'save_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'update_syndicated' ) );
 		add_action( 'before_delete_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'separate_syndicated_on_delete' ) );
-		add_action( 'before_delete_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'remove_distributor_post_form_original' ) );
+		add_action( 'before_delete_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'remove_distributor_post_from_original' ) );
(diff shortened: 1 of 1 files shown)


barryceelen:
The connection ID context in the `NetworkSiteConnection` and `WordPressExternal` classes is different depending on whether we're pushing or pulling a post.

To fix #83 as well, this pull request also adds a `$connection_id` parameter to the `log_sync()` methods and uses that in the the push and pull ui functionality for each connection class.

*Note:* I haven't tested this for external connections!

tlovett1:
@barryceelen I think this one necessitates a larger discussion. Ultimately, I think you are right, but I can see a world where someone doesn't want to see a post in the push/pull list that they've deleted.

tlovett1:
@barryceelen would you mind tweaking this a bit so if the post is deleted, it becomes available again but is marked as "skipped"?

helen:
Seems like this still needs:

> tweaking this a bit so if the post is deleted, it becomes available again but is marked as "skipped"?

dkotter:
@helen This is ready for review. I've merged in the latest changes from `master`, cleaning up some merge conflicts, and also removed some code that didn't work anymore, since we changed up how the sync log works for internal connections a release or two ago.

So the way this works now, if you Pull an item from an internal connection and then later delete that item, we mark that item as being skipped in the sync log. This allows you to Pull this again if you want, though it shows in the Skipped screen, not New screen. Also note that an item has to be truly deleted before this triggers (so sending to trash doesn't trigger it, emptying the trash does).

helen:
Just so I'm clear here - this fixes #83 and is a partial fix for #93 in that it fixes this for internal connections but not external? cc @tlovett1 @barryceelen

dkotter:
@helen Yes, that is correct (though I had broken the fixes Barry had put in place for #83, when cleaning this up, so added that piece back in)

PR #157, opened 2018-07-06
Make distributable post statuses filterable

barryceelen:
The changes in #110 resulted in only allowing published posts to be distributed, with a possible exception for drafts via a filter.

This pull request adds a filterable `distributable_post_statuses()` utility function similar to `distributable_post_types()`. It is used in the `syndicatable()` function and replaces the previously added `dt_drafts_can_be_distributed` filter.

### DIFF
--- a/includes/push-ui.php
+++ b/includes/push-ui.php
@@ -20,7 +20,6 @@ function setup() {
 	);
 }
 
-
 /**
  * Check if post is syndicatable
  *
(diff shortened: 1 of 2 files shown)


includes/push-ui.php
```diff
@@ -20,7 +20,6 @@ function setup() {
 	);
 }
 
-
```
adamsilverstein: unrelated change :)

barryceelen: Cleanin' up dem whitespaces 😄

PR #158, opened 2018-07-06
Prefix distributable_post_types filter

barryceelen:
The `distributable_post_types` should prefixed to `dt_distributable_post_types`. This could be a breaking change for some users.


barryceelen:
Lumped in the unprefixed `oauth_admin_notices` action as well.

Not sure if we'd want to retain the renamed action and filter and use `apply_filters_deprecated()` and `do_action_deprecated()` to mark them such.

barryceelen:
@tlovett1 Somehow the renaming to `dt_distributable_post_types` did not make it into master (I'm assuming due to my changing and reverting in [another pull request](https://github.com/10up/distributor/pull/159)).

PR #163, opened 2018-07-13
Blacklist _wp_old_slug and _wp_old_date

barryceelen:
No need to to distribute ` _wp_old_slug` and `_wp_old_date`.


tlovett1:
Thanks, Barry!

PR #169, opened 2018-07-23
Internal distribution should prepare posts the same for push or pull

### LINKED ISSUE #168: Serialized array values in post meta are broken when distributing to internal connections
The `\Distributor\Utils\prepare_meta()` function is used to put the post's meta into the `$meta` var. It uses the WP core function `maybe_unserialize()` to set the values to the unserialized version.

`\Distributor\Utils\set_meta()` is used to set the meta on the new post. However, when it encounters an array for the value, it loops through the array adding post meta with the same key for each value. It doesn't set the whole array to the meta value. The original value is then lost.

I understand that it is difficult to make distribution work for every scenario so custom hooks should be used to control the distribution process. In `\Distributor\InternalConnections\NetworkSiteConnection`, the `push()` method sets up the posts slightly different than the `pull()` method. With the pull method, the meta, taxonomies and media are prepared as properties of the `$post` object. In the push method, they are just storied in local variables:

https://github.com/10up/distributor/blob/6923bcd2e664a8f9198bd25d72af3dc64bb7e4d9/includes/classes/InternalConnections/NetworkSiteConnection.php#L68-L70

I think the push method should be updated to store them in the same properties as the pull method uses. Like this:

```php
		$post->media = \Distributor\Utils\prepare_media( $post_id );
		$post->terms = \Distributor\Utils\prepare_taxonomy_terms( $post_id );
		$post->meta  = \Distributor\Utils\prepare_meta( $post_id );
```

And then later, use those properties when setting the meta, taxonomy and media. That way, the dt_push_post_args filter can be used to manipulate the meta values.

It seems that the push and pull methods could potentially be refactored to use a shared method that distributes the post. I'll submit a pull request that fixes this.

Here are the filters I used to fix my use case:

```php
// Fixes an issue that Distributor has with serialized arrays in post meta
function serialize_post_meta_arrays($post) {
	if (!$post || !$post->meta) {
		echo "not an expected call of the post meta filter?";
		return $post;
	}
	$meta = $post->meta;
	$serialized_meta = [];
	foreach ($meta as $key => $value) {
		$serialized_meta[$key] = $value;
		if (is_array($value)) {
			$serialized_meta[$key] = serialize($value);
		}
	}
	$post->meta = $serialized_meta;
	return $post;
}
add_filter( 'dt_push_post_args', function($post_array, $post, $args, $connection) {
	serialize_post_meta_arrays($post);
	return $post_array;
}, 10, 4 );
add_filter( 'dt_pull_post_args', function($post_array, $remote_post_id, $post, $connection) {
	serialize_post_meta_arrays($post);
	return $post_array;
}, 10, 4 );
```

rmarscher:
Switches the internal connection push method to prepare WP_Post objects in the same way as done by the pull method. This enables the post meta to be filtered when pushing which allows for a workaround to #168 .

I added a `@since` tag to the new `\Distributor\Utils\prepare_post()` function that will probably need to be edited. This may need to be part of a minor version bump instead of a patch bump since it slightly affects people already using the dt_push_post_args filter - but it seems pretty unlikely that it would cause an issue.


rmarscher:
I will fix whatever is causing the build to fail and push it again.

rmarscher:
Build is passing now. Unit tests needed to return an ID property from the get_post mock to avoid warnings.

dinhtungdu:
Thank @rmarscher! Could you please fix the conflict in `NetworkSiteConnection.php` so we can review your PR?

dinhtungdu (approved):

dinhtungdu:
Please hold the merging, I'm waiting for the 570 to be merged to have final testing.

dkotter:
@dinhtungdu #570 has been merged now, so this is ready for testing again.

dinhtungdu (dismissed):
Both push and pull worked in my test. No error or warning. Tests are passed here: https://travis-ci.org/github/dinhtungdu/distributor/builds/674211646

rmarscher:
Thanks for working on merging this. Sorry I was unable to look at it when requested.  🏆

dkotter (dismissed):

dkotter (approved):

PR #178, opened 2018-08-06
Fixes #177: Adds additional error checking when parsing media.

### LINKED ISSUE #177: Media Sideloading Stores WP_Errors
Hey there,

I discovered I believe what's a bug in the system. In the `set_media` function in `utils.php`, it runs `process_media` before storing the result in the `_thumbnail_id` post meta.

Unfortunately, `process_media` internally uses [`wp_handle_sideload`](https://developer.wordpress.org/reference/functions/wp_handle_sideload/), which can return a `WP_Error` object. That outcome is not checked by the code before it stores the media, so the error object is stored in the post meta and causes fatal errors when featured image functions are used in the site.

Recommended fix: Instead of directly setting the post meta, use [`set_post_thumbnail`](https://developer.wordpress.org/reference/functions/set_post_thumbnail/) instead to safely set the post's featured image.

ChaosExAnima:
This PR resolves #177 by adding additional error checking when ingesting media files.

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -415,12 +415,17 @@ function set_media( $post_id, $media ) {
 			}
 		}
 
+		// Exit if the image ID is not valid.
+		if ( ! $image_id ) {
+			continue;
+		}
+
 		update_post_meta( $image_id, 'dt_original_media_url', $media_item['source_url'] );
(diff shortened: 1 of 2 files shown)


tlovett1:
Awesome!

Issue #161, opened 2018-07-13
Option to skip distributing attached media


helen:
Attachments are frequently a point of failure in terms of time outs and in many situations don't need to be distributed because the original source uses third-party hosting for media or can continue to serve as the canonical source for media. There are also situations where it's not appropriate for the media to be copied elsewhere. There should be a way to skip distributing media, both in terms of code and in UI.

Related: #152

barryceelen:
For people looking to prevent or manipulate the distribution of images: the `dt_push_post_media` [filter](https://github.com/10up/distributor/blob/master/includes/classes/InternalConnections/NetworkSiteConnection.php#L125) allows you to disable or replace the handling of images for internal connections.

It would be great to add something similar to [prepare_media()](https://github.com/10up/distributor/blob/master/includes/utils.php#L230) to be able to filter the attachments to be distributed. Either to shortcut it or to add other media to it, eg, from a gallery custom field.

dkotter:
After discussing this issue (and #152) with the team, here's a proposal on how to proceed here.

**Short Term**
Currently we distribute the featured image and any images attached to the post object. But the only image source we actually update on the receiving end is the featured image, meaning the processing of any other images will typically be unneeded overhead.

In the short term, propose that we change up how we handle media, to distribute **only** the featured image by default and have an option to distribute all attached images. This option can be both a filter in the code, as well as an option in the admin, both of which would be turned off by default. If either of those options are used, we will send over attached images, but note that we still won't update image URLs in the content. Also propose standardizing the filters we have, so you can choose to not distribute images for both internal and external connections, instead of just the internal like it is now.

**Long Term**
Looking towards the future and what we can do better for media handling, propose that we add new options in the admin. So in addition to having the option to distribute the featured image and all attached images, we will add a new option to distribute all images associated with a post.

Would need to figure out the exact logic here, but the idea is to parse through the post_content and find any image referenced there, sending those images along. This solves the issue of a post that has images in the content, but those images aren't actually attached. Currently we only send attached images, so any other images are never sent along.

In addition to this change, we will also parse through the distributed post and update the image URLs to point to the new site, instead of leaving those pointing to the origin site. Would need to figure out how we handle different image sizes between sites, as well as if we want to look through post meta and update images there.

barryceelen:
Also: Image handling in Distributor has a bit of a batch-like nature. It would be awesome if we could pluck the various functions apart to make it easier to distribute/handle/assign-to-an-async-task a single image.

[...]

PR #184, opened 2018-08-17
Change media handling to only process featured image by default

dkotter:
Currently, when an item is distributed, we process all attached images to the post object and download those to the receiving site. But only the featured image is actually replaced. All other images (i.e. in content images) still reference the original site URL.

Typically the downloading of these images will be the biggest performance issue when distributing content, and if those images aren't actually being used, as is the case currently, it's an unnecessary performance burden.

Based off discussion in #161, this PR changes the default behavior to only process the featured image by default. It adds a new setting to the plugins settings page where you can change that to be featured image and all attached images, which is the current default.

I've also gone ahead and added some additional filters around media handling, both to standardize the filter we already had in place for internal connections and allow more areas where media handling can be turned off entirely.


dkotter:
Screenshot of new setting:
screen shot 2018-08-28 at 1 51 17 pm

dkotter:
@tlovett1 Just tested this again (both internal and external Push and Pull) and I did run into a few issues, which I've fixed (issue where if there was no featured image, it still processes all images, even if you don't want. And an issue where if the featured image was the first image in the media array, it wouldn't process that), but never saw the issue of only getting the featured image, when I have the setting for featured and attached selected.

One thing to note here though is that to properly test this, these changes need to be installed on both sites that are doing the distribution. The way this is set up, the site that is ingesting the content determines what to do with images, so whatever setting is on that site is what impacts this.

For instance, if I push from Site A to Site B, Site A will send all images across, like it does now. Site B then determines what to do with those images, based on this media handling setting. It will either download just the featured image, if there is one, or all images that were sent.

If I pull from Site B into Site A, Site B will send all images and Site A will determine what to do with those images, based on the media settings it has in place.

So depending on how you're testing this, you'll need to make sure these changes are installed on both sides and the settings for both are set the way you want to test. Also note, we currently pull attached images from a post, but an image can't be attached to more then one post. This caught me a few times in testing, where I was using the same image across multiple posts but WordPress attached that image to the first post I inserted it into, so the other posts weren't sending that as an attached image. That's an issue we discussed as a phase 2 approach here, to make sure we send all images a post references, attached or not.

Also, for the wording, happy to update that as needed but again, those settings don't change how images are distributed, it changes how images are downloaded/ingested/processed.

includes/settings.php
```diff
@@ -217,6 +219,33 @@ function license_key_callback() {
...
+ * @since 1.2.2
+ */
+function media_handling_callback() {
+	$settings = Utils\get_settings();
+	?>
+
+	<ul class="media-handling">
+		<li>
+			<input <?php checked( $settings['media_handling'], 'featured' ); ?> type="radio" value="featured" name="dt_settings[media_handling]">
+			<span class="description">
```
helen: I would not wrap these in a span with the `.description` class, as it makes the text italic. The class is really meant for further explanatory text for text entry fields. They do need to be `label`s though. The label can be wrapped around the `input` so you don't have to add an `id`.

Issue #174, opened 2018-07-30
Support pulling more than one post type from external connections.


tlovett1:
External connection pull screen only shows the "post" post type.

tlovett1:
We discussed this today and decided on the following solution:

Instead of showing all post types in one list, we will have a dropdown above the list table with each available post type on the remote site. You will then have to switch between post types. Post types unavailable on the current site will be greyed out.

We will extend this functionality to internal connections as well.

PR #187, opened 2018-08-24
Choose post type to pull

dkotter:
This addresses the issues mentioned in #174.

Previously, if you were trying to Pull content from an External Connection, you'd only get Posts, no other content types would be shown. If pulling from an Internal Connection, it would show all public post types, but it was using the post types from the site you're on, not the site you're pulling from.

This PR changes up how pulling works. When the Pull screen is loaded, we get an array of available post types. This will utilize a new `get_post_types` method that is extended on each connection type (internal and external). This will get the available post types from the remote site. We then get the public post types from the site we're on and we compare the two, returning only post types that exist on both sides.

These post types are then used in a new dropdown, next to the Connection dropdown, allowing you to choose what post type to see.
screen shot 2018-08-31 at 11 09 03 am

On initial load, whatever post type is first in the list will be selected and used in the query to populate the view. You can then change the post type in the dropdown and this will reload the page, the same as the Connection dropdown works, adding a query variable for the post type we want.

One thing I wasn't sure of was the Content Type column. For now I've left that but it does seem kind of pointless, now that every item on the page should be the same type. If we don't want that anymore, I can remove any code around that.


helen:
This works well for me and makes sense. The only issue I've found is that attempting to change/filter post type when viewing pulled or skipped sends me back to the `New` list, seems like it's losing that part of the URL.

dkotter:
@helen Nice catch! That should be working now.

PR #189, opened 2018-08-29
Fix implementation of set-screen-option filter

rebeccahum:
Hi folks!

We’ve noticed this filter is not implemented correctly. This PR ensures correct implementation so that no issues arise.

Thanks so much,

### DIFF
--- a/includes/external-connection-cpt.php
+++ b/includes/external-connection-cpt.php
@@ -140,7 +140,7 @@ function ajax_begin_authorization() {
  * @return mixed
  */
 function set_screen_option( $status, $option, $value ) {
-	return $value;
+	return 'per_page' === $option ? $value : $status;
 }
 
 /**
--- a/includes/external-connection-cpt.php
+++ b/includes/external-connection-cpt.php
@@ -140,7 +140,7 @@ function ajax_begin_authorization() {
  * @return mixed
  */
 function set_screen_option( $status, $option, $value ) {
-	return 'per_page' === $option ? $value : $status;
+	return 'connections_per_page' === $option ? $value : $status;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


adamsilverstein:
Hey @rebeccahum,

I really appreciate your proactive fix for what looks like a clear issue at first glance. Looking into this a bit further, however I think the original code is correct. Can you review and see if you agree or if I am missing something?

Even though the normal pattern is to return the first parameter passed from a filter, in this case doing so short circuits saving the value. Instead, we need to return the third parameter which contains the actual value. The filter documentation could be improved to make this clearer.

It makes more sense if you take a look at the core source code: 
https://cl.ly/f442d99fbe45 (https://core.trac.wordpress.org/browser/tags/4.9/src/wp-admin/includes/misc.php#L614)

The filter will pass three parameter - false, $option, and $value. Returning the first parameter, false, will cause the meta to not be saved. Returning the third parameter, '$value' will cause that value to be saved.

You can test this after you commit by going to the Distributor->Pull Content or External Connections screens, opening screen options, changing the value and saving.

rebeccahum:
@adamsilverstein Hiya!  So, multiple filters may be in place and thus, the purpose of the filter is both whitelisting and sanitizing of the values, but much of the docs around it are indeed unclear. By returning `$status` if your option key doesn’t match, you leave it to other filters to do their thing when their option matches. If the option does match, you can then do range-checking and enforce a maximum if needed.  Here's an example: https://github.com/buddypress/BuddyPress/blob/master/src/bp-members/classes/class-bp-members-admin.php#L1355

adamsilverstein:
Ah, multiple filters! Thanks for the details and clarification - that is a very confusing filter! Looking at your changes in the PR this looks good to me.

Thanks again!

adamsilverstein:
Thanks @rebeccahum - I'll test the latest version in our preprod environment,

adamsilverstein:
* Fixes an issue where screen options on distributor pages did not save correctly.

adamsilverstein:
Retested - looks good, merging. Thanks @rebeccahum!

Issue #191, opened 2018-09-04
Meta value not pushed correctly if the value is an array


ravichandrach1:
For Internal connections within the same multisite, If meta value is an array, only the last value is being saved for a pushed post.

Steps to reproduce:

1. Create a new post on the source site and create a custom meta with value `a:3:{i:0;i:1;i:1;i:2;i:2;i:3;}`.
2. Publish the post and Distribute it to any site on the network.
3. Check the meta value for the distributed post on the target site, the meta value will be `3`

The issue seems to be at https://github.com/10up/distributor/blob/master/includes/utils.php#LC116. As we are looping through all the values in the meta array, we are updating the meta with the current value in the array, so only the last value in the meta array will be persisted. Not sure on why we are looping through the values if the meta value is an array.

```
$meta_array = (array) $meta_value;
			foreach ( $meta_array as $meta_item_value ) {
				if ( ! in_array( $meta_key, $blacklisted_meta, true ) ) {
					$meta_item_value = maybe_unserialize( $meta_item_value );
					update_post_meta( $post_id, $meta_key, $meta_item_value );
				}
			}
```

jeremyfelt:
I ran into this as well and started poking at a solution (#193). With that branch, I'm able to distribute featured media and meta array values as expected.

dkotter:
@ravichandrach1 Can you test this again with PR #193? I've just tested following your instructions and I do see the meta values being distributed correctly now if they are arrays. But would love if you could test as well and confirm this fixes the issues you're seeing. Thanks!

ravichandrach1:
@dkotter Tested PR#193 on multiple sites and networks, it fixes the meta value array issue. This is good to merge. Thanks!

PR #193, opened 2018-09-05
Restore support for arrays when storing meta

jeremyfelt:
I ran into #191 this morning and started poking around. It looks like the meta handling changed in #67 because of how featured media meta was breaking during storage.

https://github.com/10up/distributor/blob/30474c9c34834b43e3c51ed422b5a1bd5fbf108b/includes/utils.php#L484

I think that can be traced to the above line, which retrieves all meta for the media item and returns each key as an array by default. This mimics `get_post_meta( $post_id, 'my_key', false );` rather than `get_post_meta( $post_id, 'my_key', true );` and I'm guessing led to issue reported and fixed in #67.

This PR:
* Loops through the media meta and pulls only the first array value for each meta key. This probably needs to also `count()` the existing size just in case somebody is truly storing multiple keys/values.
* Restores support for arrays when setting meta.

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -125,7 +125,7 @@ function set_meta( $post_id, $meta ) {
 				update_post_meta( $post_id, $meta_key, $meta_value );
 			}
 		} else {
-			$meta_array = (array) $meta_value;
+			$meta_array  = (array) $meta_value;
 			$meta_values = array();
 			foreach ( $meta_array as $meta_item_value ) {
 				if ( ! in_array( $meta_key, $blacklisted_meta, true ) ) {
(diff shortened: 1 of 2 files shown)


dkotter:
@jeremyfelt Code here looks great, so thanks for this PR! I've also tested it out in a few scenarios and everything seems to work fine.

Couple questions:

1. Could you update your branch with the latest changes from `master`? We had some PHPCS issues popup with the update to the WordPress Coding Standards, but those should be fixed in `master` now. I think that will clear up the failed checks here
2. You mention:

> This probably needs to also count() the existing size just in case somebody is truly storing multiple keys/values

Is this something that you feel should be added to this PR before it gets merged in? Not sure how big of an issue that is, but happy to review more if you wanted to add that.

Thanks!

jeremyfelt:
@dkotter Thanks! Just rebased the branch on top of the latest master. I didn't spend much time poking at the `WP_Mock` stuff in the test failures. Let me know if I need to do anything there.

And we just chatted elsewhere about the multiple key/values, pasting that for the record:

I think the PR solves the initial issue, though still leaves a sub-bug open. If anybody stored values with `add_post_meta()` and relied on multiple rows being stored in the DB with the same meta key, then the distribution of that meta will still be broken.

IMO, the handling of multiple meta key/values can be skipped for now, though should probably be addressed with tests in the future.

PR #217, opened 2018-10-02
WP Assure Tests

tlovett1:
[WP Assure](https://github.com/10up/wpassure) is new project, currently in beta, for acceptance testing based on snapshots of the database/environment. WP Assure, using Docker, actually spins up a WordPress environment and via Selenium interacts with the websites testing real functionality in the browser.

This PR adds acceptance tests that test the following:

Tests that settings actually save
Test push menu - opens properly, connections cam be selected and marked for distribution, connections cross out after being pushed to, make sure selected connections are removable.
Test post linking and unlink
Test we can push a post as draft/published
Test distributed posts contain all data from original post, including media
Test distributed post UI - marked as distributed on post table, shows unlink message, test connection number is shown on original post
WP Assure is integrated with Travis and will run automatically on each PR.

WP Assure is currently in beta so there will be little bugs here and there that we have to work around, but it will still provide valuable testing to improve plugin stability. We'll want to continue to add WP Assure tests as we add new features; there are also many features that still need WP Assure tests.


tlovett1:
@helen this is ready.

PR #225, opened 2018-10-09
Handle multiple meta fields that use the same key

### LINKED ISSUE #200: Distributing multiple meta fields with the same key
WordPress allows storing multiple pieces of meta using the same meta key but in different rows in the database. You can then pull out all this meta into separate arrays under the same meta key. I don't think this is a common use case in WordPress but currently Distributor does not support this.

If I have two pieces of meta stored in the database separately, both using the same key, when Distributor prepares that meta (in the `prepare_meta` function), whatever piece of meta it processes last will be the one it sends along.

Steps to test:

1. Create a new post
2. Add some meta to this post using `add_post_meta`. Example `add_post_meta( 1, 'test_meta', 'First Test' )`
3. Add more meta using the same key. Example `add_post_meta( 1, 'test_meta', 'Second Test' )`
4. Distribute this post to another connection, external or internal
5. Check the database for that post and you'll only see one entry for `test_meta`, with the value being `Second Test`. The original post has two entries for `test_meta`

I think we'll need to adjust our logic in the `prepare_meta` function and the `set_meta` function. This is related to PR #193

dkotter:
In the current approach to processing meta, if an item has multiple pieces of meta that happen to use the same key (which WordPress allows), we don't pass all of that meta along, only whatever meta ends up being processed last (as we were using the meta key as the key in our array, so each one would overwrite the previous).

This attempts to fix that by setting our prepared meta in a numeric array, so if multiple pieces of meta exist with the same key, the first will be set at index 0, the next at index 1 and so on. For the majority of meta, only one will exist for each key, so only index 0 will be set. But this allows the use case of multiple pieces of meta with the same key to be sent across.

The actual setting of the meta needed to be modified as well, to take into account this new array structure as well as to take into account previous values (if any) and the difference between `add_post_meta` and `update_post_meta`. We grab all existing meta from the distributed post (if any, there won't be the first time a post is distributed) and when processing the meta we were given, we check if we already have existing meta. If we do, we use `update_post_meta` and we pass in the previous value, which allows WordPress to determine how best to update that meta. If we don't have a previous value, we use `add_post_meta`.

I also fixed up the unit tests for this change, though I do believe these tests could be more robust in the various meta scenarios.

Fixes #200.


dkotter:
I've added a few more tests to this so I think this is ready for review now.

Issue #228, opened 2018-10-09
Add Distributor plugin icon to plugin update UI


helen:
There's already an SVG, so it should be a pretty straight-forward PR to do this. See https://github.com/10up/10up-experience/blob/develop/10up-experience.php#L35-L43

helen:
How to test: change the version number in `distributor.php` to something lower (like `1.0.0`) and head to the update screen. If it's not there, you may need to go to the installed plugins list and force an update check from there before going back to the update screen, where you should see the icon in place.

helen:
Fixed in #232

PR #232, opened 2018-10-11
Add Distributor plugin icon to plugin update UI

s3rgiosan:
Implements #228


helen:
Thanks for jumping on this - I would suggest using the Distributor icon that's already in the `assets/img` folder and not adding the 10up one.

s3rgiosan:
Sorry, my bad. I already pushed a fix.

helen:
For reference, here's what it looks like in context:

screen shot 2018-10-15 at 1 32 18 pm

Issue #244, opened 2018-10-26
Timeout on ajax_verify_external_connection


madmax3365:
When trying to create an external connection with the site, which has over 10 post types  ajax_verify_external_connection can't establish a connection because of a timeout. 
The core problem in WordPressExternalConnection's check_connections method. It gets all post types available in REST and does for each of them remote_get and remote_post to check user's permissions.
<code><pre>
foreach ( $types as $type_key => $type ) {
...
$type_response = wp_remote_get( $link, $this->auth_handler->format_get_args( array( 'timeout' => self::$timeout ) ) );
					 }
					 if ( ! is_wp_error( $type_response ) ) {
					 $code = (int) wp_remote_retrieve_response_code( $type_response );
					 if ( 401 !== $code ) {
					 $can_get[] = $type_key;
					}
...
		if ( in_array( 'POST', $routes[ $route ]['methods'], true ) ) {
					$type_response = wp_remote_post(
					$link,
					$this->auth_handler->format_post_args(
					array(
					'timeout' => self::$timeout,
					'body'    => array( 'test' => 1 ),
					)
					)
					);
					if ( ! is_wp_error( $type_response ) ) {
					$code = (int) wp_remote_retrieve_response_code( $type_response );
					if ( 401 !== $code ) {
					$can_post[] = $type_key;
					}
					}
					 }
...
}
</pre></code>

arsendovlatyan:
Hey @helen ,
In our case we have ~15 post types and it's a problem to get external connection verified, usually, therefore, we had to change external connection verification process.
It will be very nice if we can include this change in next releases, for sure happy to make any improvement/fix that you will point out.

helen:
Thanks for the diagnosis and PR - we will take a look at your approach and consider it as we figure out our ideal scenario for how this should work.

Issue #45, opened 2017-12-10
Improve warning messages when a remote connection cannot be established


jakemgold:
Setting up remote connections is causing some confusion with early testers; I think our warning messages might be a bit too vague. 

Can we be a bit more specific in catching and explaining common failure use cases? Like the REST API not being available, not finding a secure connection (even linking to a recommended auth plugin), Distributor not being detected on the other end... ?

helen:
I'm gonna connect this with #368, moving to 1.6.

peterwilsoncc:
To get some recent notes on testing external connections:

**Automatic setup**

These seem pretty good. Somewhat geek oriented with the curl error IDs but I don't this that's entirely a bad thing.

* _DNS error_
  Unable to connect to site.
  cURL error 28: Connection timed out after 5001 milliseconds (http_request_failed)
* _Not WordPress_
  Unable to connect to site. The external site is private or not a WordPress site. (rest_api_uri_not_found)
* _Distributor not installed_
  Distributor not installed on remote site.
* _REST API Requires Authentication -- error message is inaccurate_
  Distributor not installed on remote site.
* _HTTPS certificate error (self signed in testing)_
  Unable to connect to site.
  cURL error 60: SSL certificate problem: self-signed certificate (http_request_failed) 

**Manual setup**

These could still do with some work but I suspect they're not seen very frequently as manual set up is behind an additional link.

* _DNS error_
  No connection found.
* _Not WordPress_
  No connection found.
* _Distributor not installed_
  Limited connection established. Distributor not installed on remote site.
  -Push distribution unavailable.
  -Pull distribution limited to basic content, i.e. title and content body.
* _REST API Requires Authentication -- no auth_
  As above, limited connection
* _HTTPS certificate error (self signed in testing)_
  No connection found.

PR #245, opened 2018-10-29
Fix timeout on ajax_verify_external_connection

madmax3365:
Related to #244. We can use these methods to check connections, it gives us a 4x performance boost.


jeffpaul:
@dkotter does your work touch this area of the code?  If not, then I'm inclined to see if we can resolve the build error and look to land this alongside the performance improvements you've opened for v1.5.0.

dkotter:
@jeffpaul Nope, the performance changes I've been working on have only touched how we get internal connections, not external.

jeffpaul:
@madmax3365 it looks like the build failed on these changes, mind giving those a review to try and ensure this build passes?

madmax3365:
@jeffpaul  It seems that problem is in Travis. We have another pull request #262  that failed due to the same reason 
`No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`
I think you had some problems with Travis back then, so to avoid unnecessarily commits you can rerun the build.

jeffpaul:

helen:
At first glance this makes a lot of sense to me - I'd like to group this together with #45 and #368 (not dependent on #368) so we can also use this to display a more granular status check for a given connection. Moving to 1.6.

includes/rest-api.php
```diff
@@ -223,4 +223,37 @@ function register_endpoints() {
...
 	);
+	register_rest_route(
+		'wp/v2',
+		'/distributor/permissions',
+		array(
+			'methods'  => 'GET',
+			'callback' => __NAMESPACE__ . '\check_permissions',
+		)
+	);
+}
```
dinhtungdu: this call should be in a separate function, the current function name is `register_endpoints`, and route and endpoint are different. The function name should be clear and explicit about what function does.

includes/rest-api.php
```diff
@@ -223,4 +223,37 @@ function register_endpoints() {
...
+
+/**
+ * Check user permissions for available post types
+ */
+function check_permissions() {
```
dinhtungdu: This can be more explicit, how about `check_post_types_permissions` or `get_post_types_permissions`? so does the route above: `distributor/post-types-permissions`?

includes/classes/ExternalConnections/WordPressExternalConnection.php
```diff
@@ -792,64 +792,28 @@ public function check_connections() {
...
+				$permission_auth_handler = $this->auth_handler->format_get_args(
+					array(
+						'timeout' => self::$timeout,
+					)
+				);
 
-					$link = $this->parse_type_items_link( $type );
-					if ( empty( $link ) ) {
-						continue;
+				$permission_response = wp_remote_get( $permission_url, $permission_auth_handler );
```
dinhtungdu: We should support VIP Go here as other parts of our codebase by checking and using `vip_safe_wp_remote_get`.

includes/classes/ExternalConnections/WordPressExternalConnection.php
```diff
@@ -792,64 +792,28 @@ public function check_connections() {
...
+				$permission_body     = wp_remote_retrieve_body( $permission_response );
+				if ( is_wp_error( $permission_response ) || empty( $permission_body ) ) {
+					$output['errors']['no_permissions'] = 'no_permissions';
+				} else {
+					$permissions          = json_decode( $permission_body );
+					$can_get              = $permissions->can_get;
+					$can_post             = $permissions->can_post;
```
dinhtungdu: Using `array_filter` here will make the code easier to read and may save us some LOC.

dinhtungdu (commented):
Tested with 100 CPT and this PR handle it very well while couldn't do the same with the current `develop` branch. Great work! Here are just some suggestions to make the codebase more maintainable.

madmax3365:
@dinhtungdu  Applied code changes as you suggested.

includes/classes/ExternalConnections/WordPressExternalConnection.php
```diff
@@ -792,65 +792,48 @@ public function check_connections() {
...
+				if ( is_wp_error( $permission_response ) || empty( $permission_body ) ) {
+					$output['errors']['no_permissions'] = 'no_permissions';
+				} else {
+					$permissions = json_decode( $permission_body );
+					$can_get     = array_filter(
+						$permissions->can_get,
+						function ( $key ) {
+							return 'dt_subscription' !== $key;
+						},
+						ARRAY_FILTER_USE_KEY
```
dinhtungdu: I'm wondering why you use `ARRAY_FILTER_USE_KEY` here?

includes/classes/ExternalConnections/WordPressExternalConnection.php
```diff
@@ -804,65 +804,48 @@ public function check_connections() {
...
+					);
+					$can_post    = array_filter(
+						$permissions->can_post,
+						function ( $key ) {
+							return 'dt_subscription' !== $key;
+						},
+						ARRAY_FILTER_USE_KEY
+					);
```
dinhtungdu, line 842: This actually makes the code harder to read 😅. Using a private method will help increase the readability here.
```suggestion
$can_post = array_filter(
    $permissions->can_post,
    [$this, 'remove_distributor_internal_post_type']
);
```

dinhtungdu (commented):
@madmax3365 Thanks! I added some comments on your updates :)

includes/classes/ExternalConnections/WordPressExternalConnection.php
madmax3365: Ahh, you're right. I haven't work on this code for a very long time, and don't know why I thought there's an associative array in `can_get` ))

Issue #252, opened 2018-11-15
Issue with Notices in Gutenberg


JasonHoffmann:
I looked around and didn't see an issue for this, but it seems that recent updates to Gutenberg broke the notices being used in the plugin. The notices currently don't show at all with the latest version of Gutenberg. The issue appears to arise from this PR: https://github.com/WordPress/gutenberg/pull/9617.

The first issue is that the notice method from `gutenberg-syndicated-post.js` is no longer part of the `core/editor` data package, but `core/notices`. That's a fairly simple fix though.

The larger issue seems to be that the `createWarningNotice` method no longer accepts an element as it's first argument, but instead accepts a string, followed by a series of actions, the best example of which I found here: https://github.com/WordPress/gutenberg/blob/8cbfb3cca07ff7e517cd908315cfb75ab92a4eb7/packages/editor/src/components/provider/index.js#L34-L47

That means the `gutenberg-syndicated-post.js` file would need to be changed to reflect this, and some of the language would likely need to change as well. I may be wrong, and there may be a way to pass a WP Element to the `createWarningNotice` function, but I haven't found one yet. The only thing I've gotten working is refactoring the notices to use plain text strings and a series of actions after, for instance:

![screen shot 2018-11-15 at 12 19 05 pm](https://user-images.githubusercontent.com/1718298/48569802-b4440b80-e8d0-11e8-9cdb-eb09d4845ace.png)

From a UX perspective, however, you may want to change the messaging, or try and figure out a way to add plain HTML or WP Elements to the `createWarningNotice` method. I have a branch in progress I'd be happy to submit so you can see how I've got it working, but I'm betting you would want to make changes to the way the messages are structured. Let me know and I can post some more screenshots of the other messages, or submit the PR so y'all can take a look.

Thanks!

adamsilverstein:
Thanks for the bug report @JasonHoffmann - would be happy to review a branch in progress! can you please open a PR and prefix the title with `WIP:` (which explicitly prevents it from being merged)?

PR #254, opened 2018-11-16
WIP: Update Notices to latest Gutenberg Package

JasonHoffmann:
Brought up in #252 

This changes the way notices are generated, passing plain text strings and a series of actions, instead of creating a dynamic element and passing it to `createWarningNotice`. Most of the text remains the same in the notices, but some of  the nomenclature has to be shifted.

Here are the three states:
screen shot 2018-11-16 at 8 27 31 am

screen shot 2018-11-16 at 8 28 03 am

screen shot 2018-11-16 at 8 36 21 am

### DIFF
--- a/assets/js/gutenberg-syndicated-post.js
+++ b/assets/js/gutenberg-syndicated-post.js
@@ -14,7 +14,7 @@ if ( '0' !== dtGutenberg.originalSourceId || '0' !== dtGutenberg.originalBlogId 
 		message = wp.i18n.sprintf( wp.i18n.__( 'Distributed from %s. The original will update this unless you', 'distributor' ), dtGutenberg.originalLocationName );
 
 		actions.push( {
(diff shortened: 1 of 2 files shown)


helen:
Looks like we've lost the link to the original post in the `Originally distributed from X` string. Otherwise the PR looks good, the internationalization is done incorrectly but that is not your fault (see #267).

JasonHoffmann:
@helen thanks for reviewing. I've added that original post link back into the actions. Are there any steps to take in terms of proper internationalization yet, or is that still being sorted out?

helen:
@JasonHoffmann Thanks for the update, this looks good now so I'll get it merged! I haven't yet outlined specific steps for fixing up internationalization, is that something you'd be interested in?

JasonHoffmann:
@helen I think I could help out with that. I could at least help sort out the build process and fix up some strings, once you've decided what the best way to approach the issue is.

Issue #258, opened 2018-11-26
Duplication of post meta keys


arsendovlatyan:
I was testing Distributor with Woo-commerce (using external connection) and all of the sudden noticed that meta keys getting duplicated in destination, with every update i have more and more duplicated meta keys.

I'm not sure if this is related to 1.3.4 release or #200 , but couldn't find similar bug-report and want to make sure that you are aware about it.

P.C. i have used fresh WordPress install (4.9.8) with fresh WooCommerce (3.5.1) and last version of Distributor 1.3.4

@helen @dkotter want to make sure that you are aware about this issue, we are not using Distributor in production yet, but if other do, it can add a mess in post meta table.

@madmax3365 to follow

jakemgold:
Likely related to 1.3.4 - we tried to handle a situation where the same meta key is used multiple times on the source side (intentionally - WordPress supports that use case)  - and it seems likely that this is unintended side effect with WooCommerce. We'll look into, but in the mean time, you can always roll back to 1.3.3 via Github.

helen:
@arsendovlatyan Looking into this - when you say "with every update", what kind of update are you talking about?

jakemgold:
Quick clarification  - we only see the empty "_test_meta" field being duplicated in your example; all the other fields seem fine. Can you confirm that all or some other fields are also duplicating? Is it just empty key values?

helen:
Commented on the PR about this too but also want to summarize here:

> The only thing that i have noticed, is when you have 2 or more meta values, with the same keys and one of them is empty ('' or 0), in that case, all meta values in distention will eventually be reset to empty.

I think this is the same cause as what I saw in testing #259, which is that all those duplicated meta keys get set to the same value because of the way `update_post_meta()` works (doesn't take an ID). We also don't actually delete any meta at any point, so they just kind of linger there. I think we need to address post meta deletion for sure, but carefully and separately. For now, this is a good enough fix for 1.3.5 and I think will actually also alleviate a back-compat issue I noticed (although I didn't test to make sure it truly works correctly, it just won't fully break).

PR #259, opened 2018-11-26
fix of meta duplication, when meta value is empty or null

arsendovlatyan:
fix of meta duplication, when meta value is empty or null, issue #258
refactoring of `set_meta` function

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -120,36 +120,28 @@ function is_dt_debug() {
 function set_meta( $post_id, $meta ) {
 	$existing_meta    = get_post_meta( $post_id );
 	$blacklisted_meta = blacklisted_meta();
-
+	
 	foreach ( $meta as $meta_key => $meta_values ) {
+		if ( in_array( $meta_key, $blacklisted_meta, true ) ) {
+			continue;
+		}
+		
 		foreach ( $meta_values as $meta_placement => $meta_value ) {
+			/* Even if previous value is NULL, we don't need to add new one. */
+			$has_prev_value = is_array( $existing_meta[ $meta_key ] ) && array_key_exists( $meta_placement, $existing_meta[ $meta_key ] ) ? true : false;
+			if( $has_prev_value ) {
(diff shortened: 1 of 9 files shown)


helen:
I think this is a good start - there are a couple things I have on my mind here and want to try to get this done and released today:

1. We should have tests for empty and serialized meta that fail before this PR and succeed afterward. I don't know if it truly has to come in with this PR but it's important that we set a standard.
2. I think we've caused a mismatch between meta as provided in older versions and how it's processed in Distributor 1.3.4+. Namely, that older versions don't supply all meta as arrays and I think we could end up with a situation where a meta value that's an array gets each array item stored as a new piece of meta with the same meta key. Discussing this with @jakemgold I think we should clearly state that interoperability between different versions of Distributor is not guaranteed and block it in 1.4.

helen:
I made a couple of tweaks to the PR mostly for formatting and after some more testing I think we're good to go. However! I believe I see what you were saying in #258 about all values being changed for those sites that have been affected by the duplicated meta so we still have the duplicated meta problem, it just doesn't keep growing. This actually points me to a bigger problem which is that meta is never actually deleted. So while it stinks that some sites will have some duplicated meta for a while, I'd rather address deletion separately and extremely carefully. I actually don't know just how much impact this has had so far but I actually think it's relatively minimal and thankfully shouldn't have much practical effect, if at all.

includes/utils.php
```diff
@@ -126,7 +126,7 @@ function set_meta( $post_id, $meta ) {
 			continue;
 		}
 
-		foreach ( $meta_values as $meta_placement => $meta_value ) {
+		foreach ( (array) $meta_values as $meta_placement => $meta_value ) {
```
arsendovlatyan: I think we are sending meta value as array anyway, why do we need to cast them here ? Maybe because of backward compatibility ?
https://github.com/NovemBit/distributor/blob/895feb9bf21a259bf4d7d26510c38dbf0b4bf24b/includes/utils.php#L294

Issue #261, opened 2018-11-27
Duplication of terms


arsendovlatyan:
If term slug is custom set in source website and term creation is not disabled/skipped with hooks in destination, every distribution/update of a post, with this kind of term, will add a new term in destination.

arsendovlatyan:
please consider my PR #262

PR #262, opened 2018-11-28
Fix duplication of terms

arsendovlatyan:
Fix duplication of terms with every update, in case if term slug is custom set in source website and term creation is not disabled/skipped with hooks in destination. #261

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -404,8 +404,8 @@ function set_taxonomy_terms( $post_id, $taxonomy_terms ) {
 							$term_array['name'], 
 							$taxonomy,
 							[
-							'slug'        => $term_array['slug'],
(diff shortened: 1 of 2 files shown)


jeffpaul:
@arsendovlatyan it looks like the build failed on these changes, mind giving those a review to try and ensure this build passes?

jeffpaul:
Resolves #325.

jeffpaul:

helen:
I cannot figure out why it's reporting whitespace at the end of line 447 and not reporting the warnings at line 490 that I get locally. Merging, will deal with it separately if necessary.

arsendovlatyan:
Thank you very much @helen , will try to be more accurate with code and cover with tests as well.

PR #285, opened 2018-12-09
Remove autoembed filter before pushing content

### LINKED ISSUE #265: Using oEmbeds pulls the full markup instead of URLs
When pulling in a post that has social URLs (Twitter, Instagram etc) the embed code is pulled instead of the actual URLs. This results in messy HTML on the new post.

Ideally the URLs should be used and the oEmbeds retrieved on the other end, after being pulled.

To replicate -

1. Publish a post with a social URL `https://twitter.com/10up/status/1067517868441387008`
2. Pull in the post on an external site
3. The post content is stored as the following -

```
<div class="oembed-wrapper oembed--twitter">
<blockquote class="twitter-tweet" data-width="500" data-dnt="true">
<p lang="en" dir="ltr">We&#39;re looking for a Creative Director in Seattle, Washington to work closely with clients in the area &amp; others around the world. Come join us!</p>
<p>Role description: <a href="https://t.co/RZ3hPUzeGM">https://t.co/RZ3hPUzeGM</a> <a href="https://t.co/ZNvVNRy6dz">pic.twitter.com/ZNvVNRy6dz</a></p>
<p>&mdash; 10up (@10up) <a href="https://twitter.com/10up/status/1067517868441387008?ref_src=twsrc%5Etfw">November 27, 2018</a></p></blockquote>
<p><script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script></div>
```

Ritesh-patel:
Fix #265: Remove autoembed filter so that actual URL will be pushed and not the full markup of URL.


helen:
In addition to tests, I want to make sure we cover the following in code and both automated and manual testing:

|          | Push | Pull |
|----------|------|------|
| Internal |  [ ]  |   [ ]    |
| External |  [ ]     |  [ ]     |

Ritesh-patel:
@helen I have updated PR and have checked with all the four scenarios.

includes/classes/ExternalConnections/WordPressExternalConnection.php
```diff
@@ -265,6 +265,9 @@ public function remote_get( $args = array() ) {
 			$posts_url = untrailingslashit( $types_urls[ $post_type ] ) . '/?' . $args_str;
 		}
 
+		// Add request parameter to specify Distributor request
+		$posts_url = add_query_arg( 'distributor_request', '1', $posts_url );
```
adamsilverstein: @Ritesh-patel won't this cause distributor to use raw content for all external connection pulls? i'm not sure we want to do that - for sites that aren't using gutenberg, we typically want the _rendered_ content because we can't be sure both sites use the same theme/plugins eg support the same embeds/shortcodes.

adamsilverstein:
## Testing
I tested this manually, before and after the patch. I'm working on automating what I did next :) 

Here are the results so far:

### Using Gutenberg - 

### Before PR/Develop:
#### External connection
Push - no problem (original block preserved)
Pull - problem as described on the issue

#### Network connection
Push - no problem (original block preserved)
Pull -  no problem (original block preserved)

### After PR:
#### External connection
Pull - no problem (original block preserved) - the issue seems resolved, however looking at the code it needs to only use raw data when gutenberg is active

Testing non-GB next

includes/classes/ExternalConnections/WordPressExternalConnection.php
adamsilverstein: See https://github.com/10up/distributor/blob/d7a1d90bc8ac15c9296f237524939b558ef8b030/includes/classes/InternalConnections/NetworkSiteConnection.php#L88 for how we do this on push

includes/utils.php
```diff
@@ -652,3 +652,23 @@ function dt_use_block_editor_for_post_type( $post_type ) {
...
 }
+
+/**
+ * Helper function to process post content
+ *
+ * @param $post_content
+ *
+ * @return mixed|void
+ */
+function get_processed_content( $post_content ) {
```
adamsilverstein: nice!

includes/utils.php
```diff
@@ -652,3 +652,23 @@ function dt_use_block_editor_for_post_type( $post_type ) {
 	 */
 	return apply_filters( 'use_block_editor_for_post_type', true, $post_type );
 }
+
+/**
+ * Helper function to process post content
```
adamsilverstein:
```suggestion
 * Helper function to process post content.
```

includes/rest-api.php
```diff
@@ -124,6 +127,28 @@ function process_distributor_attributes( $post, $request, $update ) {
...
+ * Filter the data requested over REST API when a post is pulled
+ *
+ * @param $response \WP_REST_Response Response object
+ * @param $post     \WP_Post Post object
+ * @param $request  \WP_REST_Request Request object.
+ *
+ * @return \WP_REST_Response $response
+ */
+function prepare_distributor_content( $response, $post, $request ) {
+	if ( '1' === $request->get_param( 'distributor_request' ) && intval( $request->get_param( 'id' ) ) ) {
```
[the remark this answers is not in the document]
Ritesh-patel: Yes, it is to make sure we have valid integer id.

includes/classes/ExternalConnections/WordPressExternalConnection.php
Ritesh-patel: @adamsilverstein thanks for checking this, I was under assumption that this needs to happen for both Classic and Gutenberg editor.

includes/rest-api.php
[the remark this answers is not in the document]
adamsilverstein: `intval` will always return a truthy value, i think you want `is_int` here.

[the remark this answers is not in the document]
Ritesh-patel: `is_int` will pass for value 5 but will fail for '5' while `intval` will pass for both. I know `id` will always be an integer value but using intval for a safe side.

adamsilverstein:
@Ritesh-patel I fixed a few phpunit test fixes in https://github.com/10up/distributor/pull/285/commits/84a7b4069e07bd29be9038b12b9b38492cf64d6a and fixed the doc blocks (linting error) in https://github.com/10up/distributor/pull/285/commits/6d1655087f99024ca872c9e6008dfc5dd1f3c436. All tests will now pass after combining with https://github.com/10up/distributor/pull/304, see https://github.com/10up/distributor/pull/305 where all tests pass.

adamsilverstein:
@helen this is ready for additional review

Issue #237, opened 2018-10-16
Image sizes reference incorrect/non-existent files (attachment meta is being overwritten)


helen:
When attachments are distributed, they are sideloaded, which stores in the uploads folder and generates image sizes. The location of the original (potentially a year/month subfolder) and filenames of other image sizes are stored in the serialized `_wp_attachment_metadata` piece of post meta.  *Then* post meta for that attachment is transferred over so it can be correctly associated with a post, thus overwriting the correctly-generated `_wp_attachment_metadata`. This is an issue when image sizes or uploads folder settings aren't the same between sites.

The likely best fix for this is to blacklist `_wp_attachment_metadata` in `set_meta()` but we need to ensure that no data inside that array needs to be updated separately.

helen:
Sample contents of that meta - seems pretty safe to just leave this piece of meta alone after it's been generated on the receiving site. See #240.

```
Array
(
    [width] => 1600
    [height] => 1200
    [file] => 2011/07/100_5540.jpg
    [sizes] => Array
        (
            [thumbnail] => Array
                (
                    [file] => 100_5540-150x150.jpg
                    [width] => 150
                    [height] => 150
                    [mime-type] => image/jpeg
                )

            [medium] => Array
                (
                    [file] => 100_5540-300x225.jpg
                    [width] => 300
                    [height] => 225
                    [mime-type] => image/jpeg
                )

            [medium_large] => Array
                (
                    [file] => 100_5540-768x576.jpg
                    [width] => 768
                    [height] => 576
                    [mime-type] => image/jpeg
                )

            [large] => Array
                (
                    [file] => 100_5540-1024x768.jpg
                    [width] => 1024
                    [height] => 768
                    [mime-type] => image/jpeg
                )

            [post-thumbnail] => Array
                (
                    [file] => 100_5540-1200x900.jpg
                    [width] => 1200
                    [height] => 900
                    [mime-type] => image/jpeg
                )

        )

    [image_meta] => Array
        (
            [aperture] => 4
            [credit] => 
            [camera] => KODAK DX7590 ZOOM DIGITAL CAMERA
            [caption] => 
            [created_timestamp] => 1145697622
            [copyright] => 
            [focal_length] => 6.3
            [iso] => 80
            [shutter_speed] => 0.00125
            [title] => 
            [orientation] => 1
            [keywords] => Array
                (
                )

        )

)
```

PR #289, opened 2018-12-17
Pass Media Meta Through `prepare_meta`

gthayer:
When an image is created on a site, then distributed to another site, the original `_wp_attachment_metadata` is sent with the image thumbnail sizes. If the receiving site does not have the same size rules, the `srcset` attribute can cause 404ed images based on browser size.

It appears this was already [addressed here](https://github.com/10up/distributor/issues/237), however the media meta is not actually compared against the blacklist.

For anyone reading this who needs a patch now, you can add the following to your functions.php file:

```/**
 * If the _wp_attachment_metadata is set on an image, remove it. Let the receving site build its own data.
 *
 * @param array $media_item An array of media values.
 * @param int   $media_id The media ID on the originating site.
 * @return array
 */
function remove_attachment_metadata( $media_item, $media_id ) {

	if ( isset( $media_item['meta']['_wp_attachment_metadata'] ) ) {
		unset( $media_item['meta']['_wp_attachment_metadata'] );
	}
	return $media_item;
}
add_filter( 'dt_media_item_formatted', 'remove_attachment_metadata', 20, 2 );```

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -574,7 +574,7 @@ function format_media_post( $media_post ) {
 	$media_item['media_details'] = apply_filters( 'dt_get_media_details', wp_get_attachment_metadata( $media_post->ID ), $media_post->ID );
 	$media_item['post']          = $media_post->post_parent;
 	$media_item['source_url']    = wp_get_attachment_url( $media_post->ID );
-	$media_item['meta']          = \Distributor\Utils\prepare_meta( $post_id );
+	$media_item['meta']          = \Distributor\Utils\prepare_meta( $media_post->ID );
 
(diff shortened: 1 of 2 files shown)


gthayer:
I see the CI error here, but it is not clear to my yet why this is creating an error.

helen:
@gthayer Just so I'm fully understanding here - the blacklist doesn't actually work because `prepare_meta()` is not actually called? We do need to make sure we have test coverage for this since I guess we missed this both in automated and human testing previously :(

gthayer:
@helen Yes, that is correct. It appears the blacklist is being applied to the posts when they are distributed, but not to the images.

PR #303, opened 2019-01-11
Add context to get_available_authorized_sites

aaemnnosttv:
This PR extends the fetching of available authorized sites within a multisite network to include the context of the authorization; push or pull.

Currently, this method does not distinguish between push or pull contexts.
The proposed changes updates the method signature to accept the context as a parameter and passes it to its filters to allow for more granular control.

A simple use-case for this might be that you might want to prevent sub-sites from pushing content to the main site, but you still want to allow those sites to pull from the main site.

The added parameter is nullable for backwards compatibility.

### DIFF
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -656,9 +656,12 @@ class NetworkSiteConnection extends Connection {
 	 * Find out which sites user can create post type on
 	 *
 	 * @since  0.8
+	 *
+	 * @param string $context The context of the authorization.
+	 *
 	 * @return array
 	 */
-	public static function get_available_authorized_sites() {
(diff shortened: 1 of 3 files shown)


helen:
This makes sense to me - just so I can track for the future, were there any other spots you noticed needed or could use push/pull context?

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -656,9 +656,12 @@ public static function auth_check() {
 	 * Find out which sites user can create post type on
 	 *
 	 * @since  0.8
```
helen:
```suggestion
	 * @since  0.8
	 * @since  1.3.7 Added the `$context` parameter.
```

helen: I would add this to the PHPDoc here and also to the docblocks for the filters.

PR #304, opened 2019-01-12
Tests for oEmbed distribution

adamsilverstein:
### Description
Tests for #265

These tests fail currently because oEmbeds are expanded when distributed.

* New snapshot includes a post with an oembed for testing.
* Add new tests: testOembedNetworkPushedContent, testOembedNetworkPulledContent, testOembedExternalPushedContent, testOembedExternalPulledContent.

These tests pass after applying https://github.com/10up/distributor/pull/285 and can be used to ensure we don't have regressions.

This PR also fixes a couple other tests that broke with the new snapshot.


adamsilverstein:
Added tests for external push/pull

tests/wpacceptance/OembedTest.php
```diff
@@ -0,0 +1,163 @@
...
+		$I->waitUntilElementVisible( '.notice-success' );
+
+		$I->moveTo( 'wp-admin/admin.php?page=distributor' );
+
+		$post_info = $this->pushPost( $I, 48, 51, '', 'publish', true );
+		$I->moveTo( 'two/wp-admin/edit.php' );
+
+		// Switch to the distributed post.
+		$I->waitUntilElementVisible( '#the-list' );
+		$I->jsClick( 'a.row-title' );
```
tlovett1: Any reason you are using `jsClick`?

Issue #284, opened 2018-12-05
Post meta is never actually deleted on the receiving end


helen:
Noticed in #258 / #259:

`set_meta()` never actually deletes any post meta. This appears to always have been an issue but was exacerbated by a bug we introduced in 1.3.4 that was duplicating empty meta values that are now kind of stuck in place because of the way `update_post_meta()` works with previous values.

I don't yet know what the right thing to do is. We could:
* Be less performant and just wipe out everything not blacklisted before adding again
* Try to do some trickery with slicing the existing meta arrays down and matching up against the length of the new data and deleting the ones that are no longer represented. The issue with this is that for these situations where we have duplicated values (I don't know if there's ever a time where that's a valid thing people want to do??) it will wipe out all of the post meta with that same value regardless of how many items there are supposed to be. They will get added back, presumably, but that still seems like a weird routine.
* Something else I haven't managed to think of yet

I guess the biggest question to start is: what kind of situations would there be where the distributed post has meta that the origin post did not have and need to keep?

@tlovett1

avag-novembit:
Created PR where implemented the 'delete meta' functionality. Doing checks during post redistribute (in the `SubscriptionsController::receive_item(..)` method) by comparing stored previous request meta data (stored in `dt_subscription_update` post meta key) versus new one and deleting post meta which were existing in previous request but aren't in the current one.

arsendovlatyan:
We have discussed this issue with @avag-novembit and I would like to add my thoughts, hopefully it will give more info about our approach and explain PR sent by @avag-novembit.

The basis is following - It's a must and it's safe to delete part of the post (in this case post meta), if it has been added/modified by  Distributor. In other words, if will delete part of the post, that has been generated by Distributor, it's OK and expected, but if we will delete non-distributed data, somehow generated in directly destination  - it's not OK and unexpected.

For example, we may have a plugin in source and destination, and that plugin can add/update a meta on certain events, it may also delete this meta from all posts on deactivation, now, when we will deactivate this plugin in source and then update a post, Distributor will delete this meta in destination as well and this can generate unpredictable issues, but this issues are expected for destination site maintainer, the same way we could have issues when we were adding meta data from source. 

Deletion will be unexpected for maintainer if Distributor was never distributed that data, but all of sudden will decide to delete it. That's  why we want to make sure that the most recent update from source brought this data, that we want to delete now.

[...]

PR #310, opened 2019-01-18
Fix featured image delete push issue

ravichdev:
If we delete the featured image for a post, it does not delete for the distributed post. This happens only if the featured image is selected from Media Library, works fine if the featured image is uploaded to the actual post.

Steps to reproduce:

- Create a new post and select a feature image from the Media library, make sure to use an existing image
- Distribute the post
- The distributed post will show the featured image
- Remove the featured image in the original/source post and update
- The distributed post will still show the featured image

This PR contains an `else` statement to delete featured image if `distributor_media` is empty. There are still other places (NetworkSiteConnection.php) which would need a similar condition, I am happy to add those as well if this approach looks good to you.


helen:
Meta deletion is broadly a problem, see #284. Happy to keep this PR open and milestone it if media should be handled separately though, I haven't dug into the code to find out.

dkotter:
So we do handle media and meta separately at the moment (as far as processing and setting), so I think we will probably handle the deletion separately as well. That said, in my mind it makes sense to deal with deletion as one larger task, so we keep in mind all scenarios when implementing this.

I'm not opposed to the approach in this PR but a couple things come to my mind:

1. Some posts might have media set but not have an actual featured image set. If that post then removes those media items, this approach would delete the featured image meta, even though that meta isn't actually set. Not a huge deal but I like the idea of only deleting things that should be deleted.
2. This approach only deals with deleting the featured image meta. What about other media meta that might get changed or deleted? Or what about the media items themselves? If the original post removes those, should we delete those entirely from the receiving site?

adamsilverstein:
> What about other media meta that might get changed or deleted? Or what about the media items themselves? If the original post removes those, should we delete those entirely from the receiving site?

This feels dangerous: what if the images themselves had been used in another post on the destination site?

helen:
Doesn't seem like we're quite at a consensus about approach/result yet, punting pending further dedicated discussion.

jeffpaul:
First, let's separate the terms `removal` and `deletion` as they seem to be used interchangably in this PR when they should be considered separate cases.  Removal, to me, is that the image is unattached from the post but still available in the Media Library.  Deletion, to me, is that the image is deleted/trashed from the site and no longer available on the post or in the Media Library.

My feeling on this is that we could limit this PR to the following use case:
An origin post removes the feature image and that removal is handled on distributed remote posts as well, however it merely removes the featured image from those posts, the image is still available in the remote Medial Libraries, and does not delete the image from those remote sites.

For all other cases, we can cover them via a separate PR in a future release that more fully handles meta and other media removals/deletions.

arsendovlatyan:
Side note: I think it will be good to blacklist `_thumbnail_id` meta, the value that we are pushing with this meta never works as expected, it's the attachment ID in source, but in destination, it's something else and we are updating this value during media handling process.

For deletion we developed our solution, which is based on `dt_original_media_id` . Every time, before applying updates on a post in destination, we retrieve all original media IDs of all attachments, that are connected with the post being updated, then we  look up this IDs in new update, if  an ID is not found, we delete that attachment.  For example, we a have post in source website, which has attachments with following IDs (1, 2, 3), when you distribute this post, a new post in destination will have this attachments with different IDs, like ( 5, 6, 7), but their original media ID is saved in `dt_original_media_id` , if at some point destination receives an update, where attachment IDs are (1, 3), it's means 2 has been deleted in the source, so, we can delete 6 in destination. 

I'm not sure if this approach is acceptable or not, but you can find our solution here https://github.com/NovemBit/distributor/blob/develop/includes/utils.php ( I know that it needs improvements before PR)

I think this behavior is safe as long as  `dt_original_media_id` exists only if media was created by Distributor, so, you won't delete attachments created and attached in destination.

jeffpaul:
As this PR still needs discussion and we're working to clear the milestone to get nearer a release, I'm going to punt this to the next release in hopes we can get through our discussion and to an agreed approach by then.

includes/classes/API/SubscriptionsController.php
```diff
@@ -287,6 +287,9 @@ public function receive_item( $request ) {
 
 			if ( ! empty( $request['post_data']['distributor_media'] ) ) {
 				\Distributor\Utils\set_media( $request['post_id'], $request['post_data']['distributor_media'] );
+			} else {
+				// remove any previously set featured image
```
peterwilsoncc:
```suggestion
				// Remove any previously set featured image.
```

includes/rest-api.php
```diff
@@ -117,6 +117,9 @@ function process_distributor_attributes( $post, $request, $update ) {
 
 	if ( isset( $request['distributor_media'] ) ) {
 		\Distributor\Utils\set_media( $post->ID, $request['distributor_media'] );
+	} else {
+		// remove any previously set featured image
```
peterwilsoncc:
```suggestion
		// Remove any previously set featured image.
```

peterwilsoncc (commented):
This PR looks good to me and worked while testing with:

* internal push
* internal pull
* external push
* external pull

It looks like the `set_media()` utility function was intended to handle this situation but the `empty()` check means the relevant code doesn't run:

https://github.com/10up/distributor/blob/f769e290954a17868692de893be5b5c46161383b/includes/utils.php#L790-L792

I've put in a couple of coding style nitpicks but otherwise consider this PR as approved. ✅ 

This PR is on the 2.1.0 milestone so I'm going to hold of clicking the approve button for now so I don't accidentally merge it prior to the release of 2.0.0 in the coming weeks.

peterwilsoncc (approved):
Hitting the approve button and merging now that 2.0.0 is released.

This has been added to the milestone for 2.0.1 so should be in the next release.

PR #314, opened 2019-01-23
Add action hook after meta set

Ritesh-patel:


adamsilverstein:
@Ritesh-patel I noticed that this action is firing for the entire set of distributed meta, not for meta that is actually set (for example, blackisted meta is excluded https://github.com/10up/distributor/blob/88b2bfaad7c6093e25e779ab41dff89893b94561/includes/utils.php#L126). 

Can we fire this inside the loop for each meta that is set, eg around here: https://github.com/10up/distributor/blob/88b2bfaad7c6093e25e779ab41dff89893b94561/includes/utils.php#L148 and if not, can you explain why we need the full set of meta passed here? 

If we do keep it this way, maybe filter description can be updated to make it clear what is being passed

adamsilverstein:
also: can we add a phpunit test for this, verifying the action is triggered?

Ritesh-patel:
@adamsilverstein this hook is for after Distributor sets the post meta.

We need to pass full set of meta because we need details about which meta were processed and we are passing existing meta because we need previous value of a meta after it gets updated in this function.

Ritesh-patel:
@adamsilverstein added tests in https://github.com/10up/distributor/pull/314/commits/97a6b6a20524c70decff3bf8369248fc3235f100

helen:
@adamsilverstein does make a good point, although in the context of what this specifically enables I think a per-meta action would be a separate enhancement to make. The one thing that might make sense would be to name the action `dt_after_set_meta` to leave room for said per-meta action, since the existing WP hooks inside of meta functions don't seem to have enough context to be able to target these.

Ritesh-patel:
Renamed hook in https://github.com/10up/distributor/pull/314/commits/78db8b3037c2811d296f17a4f196b3f46068c85d. Totally agreed to leaving the room for future enhancement.

PR #316, opened 2019-01-29
Avoid null post excerpts causing pulls to fail

mphillips:
Avoid a null post excerpt when syndicating a post from an external WordPress connection when the source post type doesn't support excerpts. A null post excerpt will prevent WordPress from being able
to insert the post and it fails silently.


helen:
I'm putting this into 1.3.8 since it seems like a sane change overall. Two more things I'd like to take a look into for 2.0 though: whether there are other fields that have this same kind of routine in PHP (assignment that assumes presence), and making sure we catch these kinds of failures and report them both in code and UI (one of the broader focuses for 2.0). /cc @dkotter

PR #318, opened 2019-01-31
Leverage is_dt_debug() for logging image sideload issues.

JayWood:
In short, the image handler doesn't bubble up to the top and say "I couldn't get this image" so posts will be processed normally, but in some edge cases, the image simply isn't processed when using an external connection.

Some scenarios include CDN issues, Firewall blocks, etc... 

This PR is really to open a conversation about this issue, in an ideal world we would give feedback to the client when distributing to an external connection, but this will at least allow developers to review their logs to see what happened.

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -651,7 +651,7 @@ function process_media( $url, $post_id ) {
 
 		// Distributor is in debug mode, display the issue, could be storage related.
 		if ( is_dt_debug() ) {
-			error_log( sprintf( 'Distributor: %s', $file_array['tmp_name']->get_error_message() ) );
+			error_log( sprintf( 'Distributor: %s', $file_array['tmp_name']->get_error_message() ) ); // @codingStandardsIgnoreLine
 		}
 
 		return false;
@@ -663,7 +663,7 @@ function process_media( $url, $post_id ) {
 
 		// Distributor is in debug mode, display the issue, could be storage related.
 		if ( is_dt_debug() ) {
-			error_log( sprintf( 'Distributor: %s', $file_array['tmp_name']->get_error_message() ) );
+			error_log( sprintf( 'Distributor: %s', $file_array['tmp_name']->get_error_message() ) ); // @codingStandardsIgnoreLine
 		}
 
 		return false;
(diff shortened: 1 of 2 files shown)


includes/utils.php
```diff
@@ -648,12 +648,24 @@ function process_media( $url, $post_id ) {
 
 	// If error storing temporarily, return the error.
 	if ( is_wp_error( $file_array['tmp_name'] ) ) {
+
+		// Distributor is in debug mode, display the issue, could be storage related.
+		if ( is_dt_debug() ) {
+			error_log( sprintf( 'Distributor: %s', $file_array['tmp_name']->get_error_message() ) );
```
dkotter: We'll want the coding standards ignore line here, or this will be flagged

includes/utils.php
```diff
@@ -648,12 +648,24 @@ function process_media( $url, $post_id ) {
...
 		return false;
 	}
 
 	// Do the validation and storage stuff.
 	$result = media_handle_sideload( $file_array, $post_id );
 	if ( is_wp_error( $result ) ) {
+
+		// Distributor is in debug mode, display the issue, could be storage related.
+		if ( is_dt_debug() ) {
+			error_log( sprintf( 'Distributor: %s', $file_array['tmp_name']->get_error_message() ) );
```
dkotter: We'll want the coding standards ignore line here, or this will be flagged

dkotter:
@JayWood Left a small comment. Overall, I think this is fine, I can see this being useful to help debug image issues.

We have talked about putting in better logging and error handling as part of a larger 2.0 release. I don't know exactly what that will look like but I'd imagine that would probably change the way the logging here is done (hopefully something a little more elegant then just `error_log`, though again, not sure what that is).

So I guess we have to decide if this is worth putting in temporarily, knowing we'll probably change up how we're doing logging here soon? Or if we keep it out for now, keeping this in mind as an area we want to add logging to as part of 2.0? /cc @helen

JayWood:
Completely agree, I dislike `error_log()` in production environments, but honestly I needed something to catch the issue. I feel like the issue we dealt with was an edge case. 

Maybe we could try to run the full import and if something fails on the external connection, we delete that post/data entirely and _then_ send the reply to the remote?

helen:
I think it's okay to have `error_log()` as a temporary measure especially with concrete plans to change it up later, it's really difficult to chase down bugs right now.

PR #322, opened 2019-02-01
Fix pagination and search on pull screens

### LINKED ISSUE #273: Pull: Switching views, post types, or doing a search should reset pagination
Right now the pagination is retained which can a) be unexpected, and b) return an error when the pagination is out of bounds.

dkotter:
On the Pull screen, if you paginate through results and then switch to a different view (New, Pulled or Skipped), the pagination number is brought over which is unexpected and can break things. This fixes that by removing unnecessary query args from the URLs we use for those views.

Also, searching on these views doesn't work as expected. If you are on the Pulled screen and you search for something, it will send you to the New screen. This fixes that with some javascript to change the destination URL when you search, so we keep the proper query args.

Fixes #273.

### DIFF
--- a/assets/js/admin-pull.js
+++ b/assets/js/admin-pull.js
@@ -3,6 +3,8 @@ import jQuery from 'jquery';
 const chooseConnection = document.getElementById( 'pull_connections' );
 const choosePostType = document.getElementById( 'pull_post_type' );
 const choosePostTypeBtn = document.getElementById( 'pull_post_type_submit' );
+const searchField = document.getElementById( 'post-search-input' );
+const searchBtn = document.getElementById( 'search-submit' );
 const form = document.getElementById( 'posts-filter' );
 
 jQuery( chooseConnection ).on( 'change', ( event ) => {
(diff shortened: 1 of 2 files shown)


jeffpaul:
@dkotter can you please look into resolving the build failure issue before this moves to review?

dkotter:
@jeffpaul None of the code changes here broke the tests. I re-triggered the tests and they all pass this time, so must have been a hiccup in the testing platform. Should be good to go now.

helen:
This is a great start and the JS is easier to follow now IMO. The search issue still seems to persist however; to reproduce:

1. Go to the last page of the pull list.
2. Search for some term that only returns a few items.
3. Note that it says "no items found" and that the `paged` query var is still set.
4. Manually remove `paged` from the URL and note that it works.

Switching between sites and post types works great!

dkotter:
@helen Not seeing that issue myself locally. If I switch to the `develop` branch and then run `npm run build`, I can reproduce that right away. But then switching to `feature/pull-pagination` and running `npm run build` again, the issue goes away. I think I'm following your steps exactly but is there something else I should be doing to reproduce this?

helen:
I must have mistested somewhere, I can't reproduce now. Looks great!

PR #323, opened 2019-02-03
Fix #280: Allow media pull from local IP addresses

### LINKED ISSUE #280: Attachments cannot be loaded from local environments
Due to the `http_request_host_is_external` filter in core, remote get requests to domains that resolve to IP addresses that appear to be internal (e.g. start with `127` or `192`, among others) fail out and somewhat non-descriptively. I am not sure if this can actually impact production environments since network connections are processed differently from external connections, but I think we should probably proactively set that to `true` for Distributor requests.

We also need to consider handling and reporting the failure to distribute media items but that is broader and needs its own issue.

Ritesh-patel:


jeffpaul:
Fixes #280

helen:
I have a minor qualm about what if we end up removing it for anybody else who's put the same filter in place for their entire site, perhaps it would be best if we called it with a specific priority so we have a better chance of only unhooking our addition (not foolproof but somewhat better).

PR #328, opened 2019-02-12
Resolve risky tests.

JayWood:
This updates the unit tests to do the following:
* Extends `WP_Mock\Tools\TestCase` and not `\TestCase` to add functionality.
* Drops in the `$this->assertConditionsMet()` to the risky tests to force all Mockery conditions to be met if no assertions are present.

### DIFF
--- a/tests/php/ConnectionsTest.php
+++ b/tests/php/ConnectionsTest.php
@@ -2,7 +2,9 @@
 
 namespace Distributor;
 
-class ConnectionsTest extends \TestCase {
+use WP_Mock\Tools\TestCase;
+
+class ConnectionsTest extends TestCase {
 	/**
 	 * Test connection registration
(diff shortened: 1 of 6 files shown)


jeffpaul:
@helen are you ok with pulling this into 1.4.0?

helen:
@JayWood Do any of those tests marked as risky need actual assertions added at any point?

JayWood:
@helen Honestly I didn't look. The use of `assertConditionsMet()` ensures that the function mocks are there and removes the `risky` aspect of it all, let me take a look sometime this week and get back to you on that.

helen:
Maybe we can open an issue at another point to review anything that calls `assertConditionsMet()` and ensure that it is, in fact, met. Gonna merge this for now.

Issue #330, opened 2019-02-18
PHP Warning in set_meta function


madmax3365:
Distribution and post update generates PHP warning in the destination. <br> `Warning: Invalid argument supplied for foreach() in distributor/includes/utils.php on line 125` <br> Warning occurs because in `utils.php` `function format_media_post` line `592` plugin assigns <br> `$media_item['meta'] = \Distributor\Utils\prepare_meta( $media_post->ID );` <br> And inside  `prepare_meta` it filters meta_keys excluding all keys that are `in blacklisted_meta`. And because there's case that media_item has only metas with blacklisted keys, so in `utils.php` line `540` it passes empty array to function <br> `set_meta( $image_id, $media_item['meta'] )`. <br> We can simply check if  `isset($media_item['meta'])` before calling `set_meta` function

jeffpaul:
@madmax3365 thanks for this issue and related PR!  We're working towards a 1.4.0 release in the near future and will look at pulling this into that or a quickly following release.

PR #331, opened 2019-02-18
Fix php warning.

madmax3365:
Closes: #330


jeffpaul:
@madmax3365 thanks again for this PR.

@dkotter would you please give this a review as part of your items slated for the 1.4.0 release?

helen:
Just making a note here that I want to be sure we account for this later when we tackle #284, since this now makes the lack of meta deletion also an issue outside of `set_meta()` itself.

PR #339, opened 2019-03-05
Improve Gutenberg detection pre GB 5.0

adamsilverstein:
Fix an issue where if you are using Gutenberg 4.9.0 and WordPress 4.9.9 you get a fatal error when going to the edit attachment screen.

* Previous to Gutenberg 5.0, `use_block_editor_for_post` was named `gutenberg_can_edit_post`

Originally reported by @ivanlopez.


ryanwelcher (approved):

helen:
Recognizing that I'm looking at this without the full context of how the logic flows you into this function in the first place, it seems to me you can still end up without either `use_block_editor_for_post()` or `gutenberg_can_edit_post()` defined which will then still fatal. Seems like there needs to be a full fallback here?

Issue #346, opened 2019-03-10
Posts save keeps running


Webagent05:
Hi there,

1. After enabling ... when I save a post the "save" keeps running. (no custom post type.)
![distributor-save-hangs](https://user-images.githubusercontent.com/48355318/54084364-fb4a8680-432f-11e9-9a58-f4513d17ed9d.jpg)

2. When I disable distributor, ... the issue is gone.

So something from distributor is interering saving posts. (It aren't distributed posts, but posts added in the concenring install.)

Hope to hear from you soon.

P.s. Where can I suggest features? I have some good ideas, ... I think. ;)

Kind regards,

Webagent05

jeffpaul:
@Webagent05 can you please confirm the WordPress and Distributor versions you're using as well as any other plugins/themes so we can check for compatibility issues?

As for suggesting features, you can add those as issues here in GitHub as well... thanks!

pafro:
@Webagent05 

I'm getting this "Save" always running too.  From a previous/current issue of yours I can see that we're both using the Newspaper theme.

I see this in the chrome console when it keeps showing 'Saving':
Request:
```
/EDIT/wp-admin/post.php?post=1&action=edit&meta-box-loader=1&_wpnonce=8d3e475b61&_locale=user
```
403 Response:
```
{"code":"wp_die","message":"The link you followed has expired.<\/p><p><a href=\"\/EDIT\/wp-admin\/post.php?post=1&#038;action=edit\">Please try again.<\/a>","data":{"status":403},"additional_errors":[]}
```

If I disable the Distributor plugin that exact request then works OK, returns 302

Distributor Plugin V1.4.0
WP version: 5.1.0
Latest Theme Version: 9.6.0

Webagent05:
No interactions anymore with the developers?

Would love to see this plugin work well.

jeffpaul:
@Webagent05 @Webagent05 I'd like to try your setup with the Newspaper theme as well, but in searching there are many available.  Can you confirm the slug for that theme or any further details so I can ensure I'm testing with the same theme?

Webagent05:
@jeffpaul it's the "tagdiv newspaper theme", just paste in google.

jeffpaul:
[...]

PR #351, opened 2019-03-15
Bypass check_admin_referrer() call in the use_block_editor_for_post()

drebbits:
Since WordPress 5.0 will do a [check_admin_referrer() inside the use_block_editor_for_posts()](https://github.com/WordPress/WordPress/commit/a119d4561e10ad6c5fdd27cb3ca8c7313705b6e4), this call [would fail and kill the request](https://github.com/WordPress/WordPress/blob/5.0/wp-includes/pluggable.php#L1104) if there's any custom meta boxes in the edit screen.

Unsetting the 'meta-box-loader' in the global request would bypass that check. Then, we will restore the value once we secured the date from the use_block_editor_for_posts() API.

This fixes the [infinite saving / previewing](https://github.com/10up/distributor/issues/346) on WordPress 5.0.

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -56,8 +56,8 @@ function is_using_gutenberg( $post ) {
 	}
 
 	/**
-	 * WordPress 5.1 will do a check_admin_referrer() inside the use_block_editor_for_posts(),
-	 * and this call would fail and returns a 404 if there's custom meta box.
+	 * WordPress 5.0 will do a check_admin_referrer() inside the use_block_editor_for_posts(),
+	 * and this call would fail, returns a 404 if there's custom meta box, and kills the request.
 	 *
 	 * Unsetting the 'meta-box-loader' in the global request would bypass that check.
 	 */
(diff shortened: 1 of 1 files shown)


adamsilverstein:
PR looks good!

It feels odd that we have to work around what feels like an upstream bug. I don't understand why the meta box check was added inside this function, might be worth investigating the origin further and perhaps opening an upstream PR to change.

adamsilverstein:
I verified this is an issue in the current version of WordPress & Distributor without any additional plugins.

I tested in WordPress trunk with Distributor on the `develop` branch.

### Steps to reproduce
* create a new post, add a title and some content
* enable custom fields in the Block Editor (Gutenberg) interface: this is under the ... more menu-> options: Advanced Panels: Custom Fields
image
* At the bottom of the page, under Custom fields: Add New Custom Field, click enter new:
image
* Create a test field and click add custom field:
image
* Click 'Save Draft'

### Expected Behavior
Saving works.

### Actual Behavior
Saving fails with `wp_die` error:
image

After this PR, saving works as expected.

Issue #326, opened 2019-02-08
On Multisite when logged in pages load really slow


ivanlopez:
When you have a multisite and you're logged in we've noticed that any page that shows the distributor button in the black bar that page is slow to load. Regardless if you're in the admin or front-end. The more sites you have on your network the slower it gets. When I looked at new relic I noticed that on page load we're sending an HTTP request to each site in order to build out the site list that displays in the distributor menu.

![screen shot 2018-11-26 at 4 09 54 pm](https://user-images.githubusercontent.com/1428016/52492491-50f80b80-2b97-11e9-88f1-4711b74790f4.png)

It looks like is a similar issue to #309

jeffpaul:
Thanks for this @ivanlopez, definitely looks like an opportunity for some performance tweaks.  We'll look into this along with #309 to see what we can improve.  If you have interest and availability to work on a PR, then we'd greatly appreciated... thanks!

dkotter:
@ivanlopez @jeffpaul Made a first pass at some fixes for this issue (see #355). What I was finding, testing this locally, is we currently make HTTP requests to figure out what internal sites a user has access to. This was causing significant slowness on large networks.

The PR mentioned switches that up to use `switch_to_blog` instead of making HTTP requests. Testing this locally with an example site of ~100 connections, I saw page load drop from ~300 seconds to ~6.5 seconds. I've also added caching around these authorized sites, which dropped page load down to about ~5 seconds. Obviously results will vary, based on how many internal connections you have, but should see some decent performance gains.

@ivanlopez If you have time and still have access to the same site to test on, would be great if you could test out the PR and see if that helps with performance issues you were seeing.

There is still more we can do here in regards to performance, though I think this is a good first step. In my opinion, next step would be to refactor the Push UI (and potentially Pull UI), so instead of loading all connection information right away, it only loads when you go to actually push something.

This means for users that are creating posts, editing posts or previewing posts, they won't have the overhead of generating this connection data when they don't actually need it. Thinking something along the lines of when you hover over the Distributor button in the admin bar, we then make a request to get the connection data (could also solve #275 at the same time). Would need some UX discussion to figure out the best way to make it obvious the connection data is loading in but would help overall performance for logged in users editing or viewing content.

[...]

Issue #309, opened 2019-01-17
get_available_authorized_sites() scaling issue with lots of sites


michaelradojevic:
The time it takes to run the wp_remote_post() calls in get_available_authorized_sites() really adds up when you have more than a few sites in a multisite network (we have 49). Logged in users are getting 10-20 second page load times on supported post types. Here's a screenshot from Query Monitor:

![image](https://user-images.githubusercontent.com/46765002/51355797-0d652280-1a6d-11e9-896f-00fd2c611484.png)

I'm not familiar enough with the WordPress API to know if there's a better way to do this, but having Distributor figure out the authorized sites in the background would at least allow the post to load immediately.

It looks like #270 might be a similar issue.

jeffpaul:
Thanks for this @michaelradojevic , definitely looks like an opportunity for some performance tweaks.  We'll look into this along with #270 to see what we can improve.  If you have interest and availability to work on a PR, we'd greatly appreciated it... thanks!

dkotter:
@michaelradojevic Thanks for this report. I just pushed some fixes in #355 that significantly sped things up for me. Would love to hear your thoughts on that and if you have time to test, to see if this helps in your use case. Thanks!

jeffpaul:
@michaelradojevic any updates on being able to test PR #355 to see how it improves/impacts your loading issue?

michaelradojevic:
@jeffpaul @dkotter Sorry about the lack of updates - I've been busy with lots of non-web stuff lately (do-all IT). I'll try to get some testing in by the end of this week and update with my results.

michaelradojevic:
@dkotter @jeffpaul That took care of it! With these commits I'm now seeing only ~70 ms added to page loads with Distributor enabled vs. the 10-20 seconds I'm seeing with the current release.

jeffpaul:
@michaelradojevic thanks for the feedback, stay tuned for the 1.5.0 release as that'll most likely be where these improvements get included!

PR #355, opened 2019-03-26
Refactor how we generate internal connection data

### LINKED ISSUE #270: A large network with hundreds sites leads to inaccessible Pull Content screen
This is the first time we try the plugin on our network with about 292 sites . We network-activated the Distributor plugin, went to a site's dashboard and selected Distributor > Pull content. But the page was never able to load after more than 5 minutes. Eventually the connection timed out. We need to know if there are any known issues for this plugin to work with a large number of sites. Thanks!

dkotter:
Currently when on a push or pull screen, we generate a list of connections available. For internal connections, we grab all sites on the network and then make HTTP requests to each site, determining what sites the current user has access to, as well as what post types on each site they have access to.

This will cause performance issues on large networks, so this PR is a first attempt at refactoring how we generate internal connections. First I've switched from using HTTP requests to using switch_to_blog. Then I store this information in a transient, with a key based on the current user. We attempt to clear out this transient anytime site data is changed, or a user is added/removed from a site. There might still be edge cases that the caching isn't cleared correctly yet, but currently I have this set at a 15 minute cache time, so it expires fairly quickly as a last resort.

Testing locally on a multisite with about 100 sites on the network, I went from around 300 second page loads to about 5 second page loads. Obviously results might vary but this should provide a decent performance boost for installs with lots of internal connections.

Fixes #270, #309 and #326

### DIFF
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -678,8 +678,8 @@ class NetworkSiteConnection extends Connection {
 	 * Build the available sites a specific user is authorized to use.
 	 *
 	 * @param int|string $user_id Current user ID
-	 * @param string $context The context of the authorization. Either push or pull
-	 * @param bool $force Force a cache clear. Default false
+	 * @param string     $context The context of the authorization. Either push or pull
+	 * @param bool       $force Force a cache clear. Default false
 	 *
 	 * @return array
 	 */
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -453,12 +453,14 @@ class NetworkSiteConnection extends Connection {
 	 */
 	public static function bootstrap() {
 		add_action( 'template_redirect', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'canonicalize_front_end' ) );
-		add_action( 'wp_ajax_dt_auth_check', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'auth_check' ) );
 		add_action( 'save_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'update_syndicated' ) );
 		add_action( 'before_delete_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'separate_syndicated_on_delete' ) );
 		add_action( 'before_delete_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'remove_distributor_post_from_original' ) );
 		add_action( 'wp_trash_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'separate_syndicated_on_delete' ) );
 		add_action( 'untrash_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'connect_syndicated_on_untrash' ) );
+		add_action( 'clean_site_cache', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'set_sites_last_changed_time' ) );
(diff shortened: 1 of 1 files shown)


includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -751,6 +674,102 @@ public static function get_available_authorized_sites( $context = null ) {
...
+	 * Build the available sites a specific user is authorized to use.
+	 *
+	 * @param int|string $user_id Current user ID
+	 * @param string     $context The context of the authorization. Either push or pull
+	 * @param bool       $force Force a cache clear. Default false
+	 *
+	 * @return array
+	 */
+	public static function build_available_authorized_sites( $user_id = '', $context = null, $force = false ) {
+		$user_id      = ! $user_id ? get_current_user_id() : $user_id;
```
adamsilverstein: should $user_id default to false above?

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -751,6 +674,102 @@ public static function get_available_authorized_sites( $context = null ) {
...
+						'site'       => $site,
+						'post_types' => $authorized_post_types,
+					);
+				}
+
+				restore_current_blog();
+			}
+		}
+
+		// Make sure we save and return an array
```
adamsilverstein:
```suggestion
		// Make sure we save and return an array.
```

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -751,6 +674,102 @@ public static function get_available_authorized_sites( $context = null ) {
 		return apply_filters( 'dt_authorized_sites', $authorized_sites, $context );
 	}
 
+	/**
+	 * Build the available sites a specific user is authorized to use.
+	 *
+	 * @param int|string $user_id Current user ID
```
adamsilverstein: align parameter descriptions

adamsilverstein:
Code looks good. I'll create a test network with 100 sites and give this a test.

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -751,6 +674,102 @@ public static function get_available_authorized_sites( $context = null ) {
...
+		if ( ! $last_changed ) {
+			$last_changed = microtime();
+			self::set_sites_last_changed_time();
+		}
+
+		$cache_key        = "authorized_sites:$user_id:$context:$last_changed";
+		$authorized_sites = get_transient( $cache_key );
+
+		if ( $force || false === $authorized_sites ) {
+			$sites           = get_sites();
```
adamsilverstein: Any reason we only get 100 sites? seems like we could handle far more. Maybe even make this filterable? 
```suggestion
			$sites           = $sites           = get_sites( array (
				'number' => 1000,
			));
```

adamsilverstein:
Excellent work @dkotter! Testing locally with 125 sites on my network, the difference is dramatic! Back end and front end page complete loads took a long amount of time or even failed before this.

I noticed `get_sites` returns 100 sites by default, I think we could bump this up: https://github.com/10up/distributor/pull/355/files#r273733065

includes/classes/InternalConnections/NetworkSiteConnection.php
dkotter: Set the default here to `false`

includes/classes/InternalConnections/NetworkSiteConnection.php
dkotter: Bumped this up to 1000

dkotter:
@adamsilverstein Thanks for the review! I've gone through and fixed everything, as well as bumped the amount of sites we grab to 1000. The code we had here previously didn't set anything other than the default, so I had left it at that, but I do think increasing that makes sense.

adamsilverstein:
@helen any concerns, or shall we go ahead and merge this?

PR #365, opened 2019-04-08
Lazy load available push connections

dkotter:
When on a screen that allows you to push content (currently when editing a supported post type or viewing that post on the front-end and being logged in) when the page loads, we grab all available connections to build out our Push UI.

If you're not going to actually push content, this ends up slowing down the page load unnecessarily. This PR attempts to fix this by not loading any of this connection information on page load but loading it only when a user goes to push (i.e. hovers over the Distributor button in the admin bar). This action triggers an ajax request that hits a new endpoint that builds this connection information. This builds out connection information the same way it does now, just happens on an ajax request and not on initial page load.

We then take the same HTML output we use currently and put that into a javascript template. Once our ajax request finishes, we take that template, combine it with the data we received and output that in our Distributor dropdown.

This is the second part of the performance fixes mentioned in #326. The changes here should also be combined with the changes in #355, as that PR provides the best performance improvement (and without those changes, for sites with lots of internal connections, the ajax request will probably fail due to timeouts).

I'm open to suggestions here as far as the interface we show when the data is being loaded. Currently I'm using the same spinner we show when items are being distributed, but it is kind of small and might need to at least be bigger, if not something different all together.

### DIFF
--- a/assets/js/push.js
+++ b/assets/js/push.js
@@ -53,6 +53,19 @@ jQuery( window ).on( 'load', () => {
 		actionWrapper           = distributorPushWrapper.querySelector( '.action-wrapper' );
 		postStatusInput         = document.getElementById( 'dt-post-status' );
 		asDraftInput            = document.getElementById( 'dt-as-draft' );
+
+		/**
+		 * Listen for connection filtering
+		 */
+		jQuery( connectionsSearchInput ).on( 'keyup change', _.debounce( ( event ) => {
+			if ( '' === event.currentTarget.value ) {
+				showConnections( dtConnections );
+			}
+
+			searchString = event.currentTarget.value.replace( /https?:\/\//i, '' ).replace( /www/i, '' ).replace( /[^0-9a-zA-Z ]+/, '' );
+
+			showConnections();
+		}, 300 ) );
 	}
 
 	/**
@@ -305,17 +318,4 @@ jQuery( window ).on( 'load', () => {
 
 		showConnections();
 	} );
-
-	/**
-	 * List for connection filtering
-	 */
-	jQuery( connectionsSearchInput ).on( 'keyup change', _.debounce( ( event ) => {
-		if ( '' === event.currentTarget.value ) {
-			showConnections( dtConnections );
-		}
-
-		searchString = event.currentTarget.value.replace( /https?:\/\//i, '' ).replace( /www/i, '' ).replace( /[^0-9a-zA-Z ]+/, '' );
-
-		showConnections();
-	}, 300 ) );
 } );
--- a/assets/js/push.js
+++ b/assets/js/push.js
@@ -153,6 +153,7 @@ jQuery( window ).on( 'load', () => {
 			return;
 		}
 
+		distributorPushWrapper.classList.remove( 'message-error' );
 		distributorPushWrapper.classList.add( 'loaded' );
 
 		const data = {
@@ -170,7 +171,8 @@ jQuery( window ).on( 'load', () => {
 			data: data
 		} ).done( ( response ) => {
 			if ( ! response.success || ! response.data ) {
-				doError();
+				distributorPushWrapper.classList.remove( 'loaded' );
+				distributorPushWrapper.classList.add( 'message-error' );
 				return;
 			}
 
@@ -183,7 +185,7 @@ jQuery( window ).on( 'load', () => {
 			setVariables();
 		} ).error( () => {
 			distributorPushWrapper.classList.remove( 'loaded' );
-			doError();
+			distributorPushWrapper.classList.add( 'message-error' );
 		} );
 	}
 
--- a/assets/js/push.js
+++ b/assets/js/push.js
@@ -1,6 +1,6 @@
 import jQuery from 'jquery';
 import _ from 'underscores';
-import { dt, dtConnections } from 'window';
+import { dt } from 'window';
 
 let selectedConnections = {},
 	searchString        = '';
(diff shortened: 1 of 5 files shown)


dkotter:
Cleaned up the failing tests. Had to make sure we run `npm install` before running our tests, so it recompiles our JS files.

assets/js/push.js
```diff
@@ -122,6 +148,46 @@ jQuery( window ).on( 'load', () => {
...
+
+				distributorPushWrapper.innerHTML = processTemplate( 'dt-show-connections' )( {
+					connections: dtConnections,
+				} );
+
+				setVariables();
+			}, 500 );
+		} ).error( () => {
+			setTimeout( () => {
+				doError();
```
adamsilverstein: maybe remove the loaded class from `distributorPushWrapper` on failure so load will try again

assets/js/push.js
```diff
@@ -122,6 +148,46 @@ jQuery( window ).on( 'load', () => {
...
+
+		const xhr = dt.usexhr ? { withCredentials: true } : false;
+
+		jQuery.ajax( {
+			url: dt.ajaxurl,
+			xhrFields: xhr,
+			method: 'post',
+			data: data
+		} ).done( ( response ) => {
+			setTimeout( () => {
```
adamsilverstein: why are these timeouts required?

includes/push-ui.php
```diff
@@ -67,6 +68,126 @@ function syndicatable() {
...
 
+/**
+ * Get available connections for use in the Push UI.
+ *
+ * @return void
+ */
+function get_connections() {
+	if ( ! check_ajax_referer( 'dt-load-connections', 'loadConnectionsNonce', false ) ) {
+		wp_send_json_error();
+		exit;
```
adamsilverstein: I think  there is no need for exit after wp_send_json_*, they die or wp_die.

includes/push-ui.php
```diff
@@ -67,6 +68,126 @@ function syndicatable() {
...
+				'url'        => $connection->base_url,
+				'name'       => $connection->name,
+				'syndicated' => ( ! empty( $connection_map['external'][ (int) $external_connection->ID ] ) ) ? true : false,
+			];
+		}
+	}
+
+	wp_send_json_success( $dom_connections );
+
+	exit;
```
adamsilverstein: exit not needed after wp_send_json

tests/wpacceptance/PushMenuTest.php
```diff
@@ -25,7 +25,9 @@ public function testMenuItemHover() {
 
 		$I->click( '#wp-admin-bar-distributor a' );
 
-		$I->seeElement( '#distributor-push-wrapper .new-connections-list' );
+		$I->waitUntilElementVisible( '#distributor-push-wrapper .new-connections-list' );
+
+		$I->seeElement( '#distributor-push-wrapper .add-connection' );
```
adamsilverstein: Can we add some additional tests or steps to verify that the sites menu loads correctly? I'm not sure this really verifies that.

adamsilverstein:
Testing results:
* with 100+ sites, I get an internal server error on the Ajax request, the menu never loads, I assume https://github.com/10up/distributor/pull/355 will fix this. Regardless could have better error handling, show the user a message when the menu fails to load.
* with a small number of sites, this worked very well, screencast: https://cl.ly/2e7695f4bc0a

The loader looks a little lonely spinning in the huge empty space while the menu loads. Couple of possible ideas: Add some text "Loading distributor site..."... instead of a spinner use pulsating gray boxes in the shape of the ui that is expected.

assets/js/push.js
dkotter: Don't think they are. Copied this `ajax` function from the other one we have in this file and it uses `setTimeout` (not sure why). But I've removed them from here.

includes/push-ui.php
dkotter: Yeah, good catch. Copied this from the other `ajax` function we have here but these aren't needed.

tests/wpacceptance/PushMenuTest.php
dkotter: This should verify the menu is loading correctly. The previous test verifies the element `.new-connections-list` exists. In our case, that element is loaded via `ajax`, so we have to `waitUntilElementVisible`, which will timeout eventually if that element doesn't become visible (if the ajax request fails or no connections are returned). Assuming that element becomes visible, we then check if there's an `.add-connection` element, which is the container that holds each individual connection.

dkotter:
@adamsilverstein Review items cleaned up.

And yeah, without the fixes in #355, sites with lots of internal connections will have issues with this ajax request. Open to suggestions on what we should show as the error here. If the request fails, I'm assuming it will probably always fail, so having it try again probably won't help. But some sort of error message would be nice, or at least better than nothing.

And yeah, the loading state needs to be addressed. I pulled the same loader we use when actually pushing to a connection, but it is pretty small and easy to miss. I like the idea of having a placeholder structure that pulses, might need some front-end help to make that happen though (will have to investigate how that is usually handled).

dkotter:
Okay, I've added a basic placeholder loader (following some information I found [here](https://github.com/zalog/placeholder-loading)). Not sure it's better then the loading icon but maybe at least gives us a starting point. Also made a few changes to make sure an error message is shown if an error occurs. Can adjust that error message though as needed.

out

tests/wpacceptance/PushMenuTest.php
adamsilverstein: 👍 thanks for clarifying

assets/js/push.js
```diff
@@ -122,6 +148,45 @@ jQuery( window ).on( 'load', () => {
...
+		} ).done( ( response ) => {
+			if ( ! response.success || ! response.data ) {
+				distributorPushWrapper.classList.remove( 'loaded' );
+				distributorPushWrapper.classList.add( 'message-error' );
+				return;
+			}
+
+			dtConnections = response.data;
+
+			distributorPushWrapper.innerHTML = processTemplate( 'dt-show-connections' )( {
```
adamsilverstein: @dkotter unfortunately `innerHTML` is insecure, if a malicious attacker were able to hijack the ajax request, they could return a malicious script that would get injected here. 

Ideal fix here would be to return a data structure and then construct the HTML using DOM construction or a template structure (template literals would work as well). Another option would be accepting the risk - it certainly is easier/feels safer leaving the HTML generation in place.

dkotter: Okay. Currently we have a javascript template we output within `push-ui.php` (line 437), which we then take that and use underscores JS to take that template, plus the data from our ajax request, to build out the markup (which we then add to the element using `innerHTML`). Note the data we use in this template is escaped using underscores escaping function (example `{{ connections[ key ]['name'] }}`.

So we do currently use a template structure here but are you saying that should be done differently? Note I copied what we were already doing another place in this same file (line 141), just to keep things consistent. Doesn't mean that's the right approach but if we want to change this, we'll probably want to change in both places.

jeffpaul:
@dkotter any chance you can work through the last update from Adam over the next week or so, that way they can land in the 1.5.0 release?

assets/js/push.js
helen: Let's just leave a comment for now and revisit the overall approach later. For instance, I think it might be nice to store the last 3-5 sites that were distributed to and front-load those.

```suggestion
			// Allowing innerHTML because processTemplate escapes values
			distributorPushWrapper.innerHTML = processTemplate( 'dt-show-connections' )( {
```

adamsilverstein:
Looks good! Thanks @helen for tracking down how that innerHTML I flagged was used.

bgasperosky:
I'm really happy to see this is being actively worked on. We have a multisite network of about three dozen sites and we're still growing. I've found with any site when Distributor is activated, a request to admin-ajax.php is sent to every site, whether the plugin is active or not. So I'm seeing a 35 second load time whenever edit.php or post.php is called on a site where Distributor is active, compared to about 1.5 seconds when it is not. I do have a test multisite network which is a copy of our production network up and running, so I'd be happy to help test this feature. I'm a newbie to GitHub, so how do I download a nightly or a beta to test with?
distributor-multisite-slow-due-to-access-checks

PR #353, opened 2019-03-18
Add application password auth flow

adamsilverstein:
Resolves #286

PR #368, opened 2019-04-16
Add Auth Setup Wizard

### LINKED ISSUE #286: Feature: Easier setup flow for Application Passwords & External connections
Since https://github.com/georgestephanis/application-passwords/pull/39#issuecomment-283677076 Application Password supports a much easier flow when setting up an application password. Distributor can link to the remote site with a special URL that lets the user create the new key on the fly, during the setup flow. 

Implementing this flow would make external connection setup much smoother.

From the PR:
> The idea is to give a page in wp-admin where you can send a user, they can authenticate via wp-login.php if needed, then manually approve (or reject) it and it will send the user back to the app with an application password in tow that the app can store and use.

> It saves the user from having to find their profile page and create and retype or copy/paste an application password. Should be more understandable.

adamsilverstein:
Resolves #286, #503

Replaces https://github.com/10up/distributor/pull/353

### Description

This PR adds a "Setup Wizard" for setting up external connections leveraging the Application Password plugin. Both sites need to be running the latest version of Distributor.

### Changes

* Use CSS class based visibility vs JS show/hide for auth type field visibility. This is less fragile/more scalable than the previous approach. A class is added to the wrapper with the slug of the selected auth type, enabling hiding/showing elements by wrapper class. Could be refactored, ideally further separating the auth types and using a more modern approach, likely a stateful React mini-app.
* CSS to hide some confusing bits of the Application Setup screen, ideally we can improve these in an upstream PR to https://github.com/georgestephanis/application-passwords/ with improvements to the Application Setup screen.
* JavaScript in `admin-external-connection.js` to handle clicking the Setup Connection Wizard "Authorize Connection" button
* Validate fields before redirecting to the Application Setup/Approval screen
* Validate remote site before connecting: parse dom to determine REST API root (provided as meta tag). Add and load dt_meta endpoint with plugin version. Require minimum version (1.5.0) during setup flow.
* Handle return from the Application Setup screen: localize the returned variables, then use them to populate the user/password fields and save the connection.
* Handle Application Setup rejection by showing an error message.
* Some whitespace/wrapper changes (diff best viewed with whitespace changes hidden)
* Hide the 'Choose Authentication' dropdown on existing connections, eg editing - you can't change the connection type once you set up a connection (is there a use case where you would want to?

### Potential Issues

* If the remote site already has an older version of Application Passwords installed, the connection will fail. Resolution: look for and warn about existing Application Password plugin installs.
* If the remote site is running an older version of Distributor and no Application Password plugin, the wizard will fail. resolution: detect the plugin version (or presence of correct remote Distributor version before continuing with wizard.
* non standard rest api roots. we may be able to add a return variable to the auth flow such as the api root.

### Connection Wizard Flow
* User enters the remote site URL and clicks "Authorize Connection"

* User is directed to remote site, where they (login and) **approve the connection**

* The new connection is created:


.gitignore
```diff
@@ -5,3 +5,4 @@ vendor/*
 !vendor/yahnis-elsts
 release/*
 dist/*
+screenshots/
```
adamsilverstein, line 9: ignore screenshots from debugging runs

assets/js/admin-external-connection.js
```diff
@@ -11,7 +17,117 @@ const titleField                  = document.getElementById( 'title' );
...
+		return false;
+	}
+
+	// @todo Check that the current version of Distributor is available on remote site here.
+	//
+
+	// First, locate the remote site REST API root.
+	jQuery.get( siteURL, function( remoteSite ) {
+		let endpoint = false;
+		// Look for the <link rel='https://api.w.org/' href='http://developwordpress.localhost/wp-json/' />
```
dkotter: Another potential risk with this approach is sites that have removed the REST API discovery links.

assets/js/admin-external-connection.js
```diff
@@ -11,7 +17,117 @@ const titleField                  = document.getElementById( 'title' );
...
+		const parsed = jQuery.parseHTML( remoteSite );
+		parsed.forEach( function( el ) {
+			const $el = jQuery( el );
+			const rel = $el.attr( 'rel' );
+			if ( 'https://api.w.org/' === rel ) {
+				endpoint = $el.attr( 'href' );
```
dkotter: Do we want to break out of this loop once we find an endpoint? Or are there scenarios where there might be multiple endpoints and we want to get the last one? Just thinking we could help with performance a bit by stopping processing once we find the endpoint (though not sure how much that matters)

composer.json
```diff
@@ -7,20 +7,36 @@
...
+              "reference":"master"
+            }
+        }
+    }
+  ],
   "require": {
     "php": ">=5.6",
-    "yahnis-elsts/plugin-update-checker": "^4.4"
+    "yahnis-elsts/plugin-update-checker": "^4.4",
+    "georgestephanis/application-passwords": "master"
```
dkotter: If we are bundling this with Distributor now, do we want to set a specific version, so we don't accidentally update this plugin and potentially break something?

dkotter:
@adamsilverstein Looks good! Left a few minor comments on the review

composer.json
adamsilverstein: Good idea! I'll change that.

assets/js/admin-external-connection.js
adamsilverstein: Yea, thats true. If this fails, we can fall back to 'guessing' `wp-json` and/or expose the manual UI instead. Plus inform the user why it failed.

assets/js/admin-external-connection.js
adamsilverstein: yes, sure - didn't even think about that.

adamsilverstein:
@dkotter Thanks for the review! I will work on addressing your feedback.

jeffpaul:
@adamsilverstein any chance you can work through these updates over the next week or so, that way they can land in the 1.5.0 release?

jeffpaul:
@adamsilverstein any chance you can work through these updates by the end of next week so they can land in the 1.5.0 release?

helen:
Punting to 1.6, still collecting review notes but tl;dr for the milestone change is that this still has some setup difficulties that need to be either documented or resolved before this is really a better solution. A lot of promise, though!

helen:
I get fatals with the Application Passwords plugin already activated, which is likely most existing users. They are for "Constant `APPLICATION_PASSWORDS_VERSION` already defined" and "Cannot declare class `Application_Passwords`, because the  name is already in use". Call stacks below.

<details>
<pre>
Notice: Constant APPLICATION_PASSWORDS_VERSION already defined in  /var/www/html/wp-content/plugins/application-passwords/application-passwords.php  on line 11

1 | 0.0063 | 527064 | {main}(  ) | .../plugins.php:0
2 | 0.9945 | 32818336 | plugin_sandbox_scrape( ??? ) | .../plugins.php:175
3 | 0.9950 | 32819872 | include( '/var/www/html/wp-content/plugins/application-passwords/application-passwords.php' ) | .../plugin.php:2223
4 | 0.9950 | 32819872 | define ( ???, ??? ) | .../application-passwords.php:11

Fatal error: Cannot declare class Application_Passwords, because the  name is already in use in  /var/www/html/wp-content/plugins/application-passwords/class.application-passwords.php  on line 9

1 | 0.0063 | 527064 | {main}(  ) | .../plugins.php:0
2 | 0.9945 | 32818336 | plugin_sandbox_scrape( ??? ) | .../plugins.php:175
3 | 0.9950 | 32819872 | include( '/var/www/html/wp-content/plugins/application-passwords/application-passwords.php' ) | .../plugin.php:2223
4 | 0.9961 | 32916824 | require_once( '/var/www/html/wp-content/plugins/application-passwords/class.application-passwords.php' )
</pre>
</details>

dkotter:
> I get fatals with the Application Passwords plugin already activated, which is likely most existing users. They are for "Constant `APPLICATION_PASSWORDS_VERSION` already defined" and "Cannot declare class `Application_Passwords`, because the name is already in use". Call stacks below.

@helen Interesting. I tried to reproduce this and haven't gotten the same error yet myself. Found I had the Application Passwords plugin installed separately all along and didn't realize it, but testing without that activated, with it activated on a single site or activated network-wide, didn't get this error.

You can see here: https://github.com/10up/distributor/pull/368/files#diff-919d7e0d1e427c335752286442d8eb99R111, we check if the `Application_Passwords` class exists before we load our version (may want an `is_plugin_active` check there as well). In theory though, this should prevent both copies being loaded at the same time.

If I remove that conditional check, I do get fatal errors then, though the interesting thing is, the error messages I get point to the version of the plugin we include with Distributor (`Fatal error:  Cannot declare class Application_Passwords, because the name is already in use in /var/www/html/wp-content/plugins/distributor/vendor/georgestephanis/application-passwords/class.application-passwords.php on line 9`), whereas your error message points to the version of the plugin in the normal plugins directory (`Fatal error: Cannot declare class Application_Passwords, because the  name is already in use in  /var/www/html/wp-content/plugins/application-passwords/class.application-passwords.php  on line 9`).

To me that implies that Distributor loaded first, included the copy of Application Passwords it is bundled with, and then the real Application Passwords loaded second, throwing the fatal errors at that point. From my understanding, plugins are supposed to load alphabetically, which should mean Application Passwords would always load before Distributor.

Just for more testing steps, what version of Application Passwords do you have installed (I tested 0.1.0 and 0.1.1)? This this a single site or multisite? If multisite, is it network activated or just site activated? And do you get these fatal errors immediately, or is there a certain page or action that triggers them?

dinhtungdu:
@dkotter I tested this PR today, I couldn't connect to external sites due to CORS policy, see: http://g.recordit.co/YPPeRpwMZ9.gif

Haven't tested the Application Password issue above though.

adamsilverstein:
@dinhtungdu this could be due to missing authorization headers: https://github.com/WordPress/application-passwords/wiki/Basic-Authorization-Header----Missing - for testing you may need to enable these manually.

Ideally the applications password plugin would detect this issue and automatically add the required lines to `.htaccess` if possible.

dinhtungdu:
@adamsilverstein thanks for the head up, I traced down the source code and found that it fails at this `get` request:

https://github.com/10up/distributor/blob/e1c031df110b78e5a326ab5356ab6d0445e16dd3/assets/js/admin-external-connection.js#L73

It worked if I connect to a site on the same network because two sites have the same origin (as the [gif](https://camo.githubusercontent.com/e0d55cf14c0b799a29fd496883a6ca810a015f81/68747470733a2f2f636c2e6c792f3164306462656631323636352f53637265656e253235323532305265636f7264696e6725323532353230323031392d30342d31362532353235323061742532353235323031312e333725323532353230414d2e676966) in the PR description, I use a multisite installation for developing distributor as well). I don't think at that point we need authorization headers; we have something to do with Access-Control-Allow-Origin header instead.

jeffpaul:
We can assume that the external site has Distributor installed, so we could consider a custom header that Distributor creates/uses for this.

dkotter:
@helen I was finally able to reproduce a couple situations that cause fatal errors.

The first is one we had talked about. If you have Distributor active and then try to activate Application Passwords, a fatal error is thrown, as we're already including that plugin ourselves. WordPress does catch this error, so it doesn't bring your entire site down (though I think that was added in 5.2?). 

There is the `activate_{$plugin}` hook that we may potentially be able to use here (I've never tried doing this) to either stop this situation from happening or at least showing a proper error message if this happens.

The other scenario I was able to find was on a multisite, if you have Application Passwords active on a single site in the network but you have Distributor network activated, you'll get fatal errors across all sites. The issue here is because Distributor is network activated, it loads first (including its' version of Application Passwords). And then Application Passwords is loaded second, causing the fatal error.

Also similar situation if you have Distributor and Application Passwords active on a single site within a network and then you try to network activate Distributor, it will throw fatals.

I just pushed a fix for these scenarios, where we move the loading of Application Passwords into the `plugins_loaded` hook. For me, this fixed these errors. Feel free to test things again with that fix in place and let me know if you still get errors.

dinhtungdu:
I did some research on CORS and it turns out that we need `Access-Control-Allow-Origin` to bet set correctly. The remote site doesn't know what site is going to make CORS request to it. We can set that header to `*` but I'm against doing it.

We can solve this problem by using AJAX. Instead of making requests directly using JS on the user's browser, we can use AJAX to send `GET` requests using the server. We'll need two AJAX actions, one for getting the endpoint and another for checking the version of Distributor installed on the remote site.

includes/external-connection-cpt.php
```diff
@@ -687,3 +724,73 @@ function filter_post_updated_messages( $messages ) {
...
+ *
+ * @param string $site_url Remote site URL.
+ *
+ * @return false|string
+ */
+function get_rest_url( $site_url ) {
+
+	$source = wp_remote_get( $site_url );
+
+	if ( ! is_wp_error( $source ) ) {
```
dkotter: @dinhtungdu Need to remove the negation (!) operator here, otherwise this will return false on success

dkotter (changes requested):

dkotter:
@dinhtungdu Nice work on this! I tested things again with a new external connection environment and I did see the CORS issue. With your latest changes though, I'm able to set up external connections using the wizard (I did have one issue that needed fixed for things to work, which I've noted in the code review).

includes/external-connection-cpt.php
dinhtungdu: Oh silly me 😅. Thank you, Darin!

helen:
This is really cool! I have a handful of things I've noticed, one of which is just a note that is probably expected behavior:

1. I had to edit the main plugin version to be `1.6.0-dev` to avoid the version check error. Just confirming that's the case for everybody else testing this as well (and noting for anybody else looking).
2. Restricted Site Access can cause the error to be a more cryptic "Distributor is not installed" (using redirect to a page) - this is not a blocker to anything, but it's a little odd to me that the header gets skipped in this case. I imagine Distributor is not expected to work when RSA is active, in any case, just a funny conflict in my local that I wanted to note in case it does come up again later.
3. UI bug - if I click "Manually Setup Connection" but then decide to switch to WordPress.com Application, the fields do not disappear. [See screenshot](https://share.getcloudapp.com/bLueYZNR) for what I mean. I think this should probably reset entirely, so hide the fields when you switch, but if you switch *back* to `Username/Password` it should give you the wizard thing again.
4. Some language things that I will be tweaking shortly :)

dkotter (approved):

Issue #370, opened 2019-04-24
`dt_push_post_args` filters are defined inconsistently.


gthayer:
The filter `dt_push_post_args` is used inconsistently, where [External Connections](https://github.com/10up/distributor/blob/develop/includes/classes/ExternalConnections/WordPressExternalConnection.php#L618) are using 3 variables with the 3rd being the class object, while [Internal Connections](https://github.com/10up/distributor/blob/develop/includes/classes/InternalConnections/NetworkSiteConnection.php#L115) have 4 variables with the 4th being the class object.

I believe we should add the missing `$args` value in the external connections call, and set the class object as the 4th value to match the internal connections call.

These should be consistant because filters looking for 4 values will get `Uncaught ArgumentCountError: Too few arguments` errors.

PR incoming...

PR #371, opened 2019-04-24
Unify `dt_push_post_args` usage

gthayer:
Unified the usage of `dt_push_post_args`, where External Connections are using 3 variables with the 3rd being the class object, while Internal Connections have 4 variables with the 4th being the class object. Added the `$args` value in the 3rd slot and pushed the class object to the 4th value.

Fixes: https://github.com/10up/distributor/issues/370


jeffpaul:
@gthayer it looks like the build failed on these changes, mind giving those a review to try and ensure this build passes?

gthayer:
@jeffpaul - I'm not familiar with the CI setup here, but it looks like there was a timeout error on the build:

`Nesk\Rialto\Exceptions\Node\FatalException: waiting for selector ".media-modal-content" failed: timeout 90000ms exceeded`

I'm not certain this was caused by the code changes. Do you have any additional insight into this? I might be overlooking a button, but I don't see a means of re-running the build.

jeffpaul:
@gthayer might be a permissions issue, I just re-ran the build.  🤞

gthayer:
@jeffpaul - It looks like the build failed again. I'm afraid I don't see what the cause might be. The commit is a fairly simple change to one line, and that file is not referenced in any of the notices in the build script.

I think the test "Push data sync" is failing, which I assume is this? dc99af51500440a672b0aeece0b49d1f/tests/wpacceptance/InternalPushTest.php#L54

jeffpaul:
@adamsilverstein @dkotter as either of you get a chance, could you check that failing test to see if it needs to be updated?

jeffpaul:
Resolves #123

PR #376, opened 2019-05-07
WP 5.2 version bump

jeffpaul:
@adamsilverstein tested and confirmed Distributor with 5.2 RC and everything worked as expected, so adding a badge to note the WP version we're tested up to.


jeffpaul:
Merging this to ensure we're showing Distributor's tested up to 5.2.

PR #378, opened 2019-05-10
Adds extra parameters to dt_create_missing_terms.

mmcachran:
### Description of the Change

Add taxonomy and term data to dt_create_missing_terms filter to allow whitelisting/blacklisting taxonomies/terms.

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -463,10 +463,11 @@ function set_taxonomy_terms( $post_id, $taxonomy_terms ) {
 			 * @since 1.0.0
 			 *
 			 * @param bool true Controls whether missing terms should be created. Default 'true'.
-			 * @param string    The taxonomy name.
-			 * @param array     Term data.
+			 * @param string              The taxonomy name.
+			 * @param array               Term data.
+			 * @param WP_Term|array|false WP_Term object or array if found, false if not. 
 			 */
-			$create_missing_terms = apply_filters( 'dt_create_missing_terms', true, $taxonomy, $term_array );
+			$create_missing_terms = apply_filters( 'dt_create_missing_terms', true, $taxonomy, $term_array, $term );
 
 			if ( empty( $term ) ) {
 
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -463,8 +463,10 @@ function set_taxonomy_terms( $post_id, $taxonomy_terms ) {
 			 * @since 1.0.0
 			 *
 			 * @param bool true Controls whether missing terms should be created. Default 'true'.
+			 * @param string    The taxonomy name.
+			 * @param array     Term data.
 			 */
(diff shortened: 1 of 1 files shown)


jeffpaul:
Resolves #379

jeffpaul:

helen:
Leaving a review note: I just want to make sure that this doesn't impact existing hooked functions - there's a change in PHP 7 that can throw notices/errors if the number of args doesn't match but I don't remember off the top of my head what exactly causes the issue. If you do know, that'd be great, but otherwise I will look into this in the next week or two.

includes/utils.php
```diff
@@ -463,8 +463,10 @@ function set_taxonomy_terms( $post_id, $taxonomy_terms ) {
 			 * @since 1.0.0
 			 *
 			 * @param bool true Controls whether missing terms should be created. Default 'true'.
+			 * @param string    The taxonomy name.
+			 * @param array     Term data.
 			 */
-			$create_missing_terms = apply_filters( 'dt_create_missing_terms', true );
+			$create_missing_terms = apply_filters( 'dt_create_missing_terms', true, $taxonomy, $term_array );
```
adamsilverstein: Since we have the `$term` already, we might as well pass it as well.

adamsilverstein:
> there's a change in PHP 7 that can throw notices/errors if the number of args doesn't match 

@helen i couldn't find that change although i vaguely recall it. In any case, I think its only an issue if your function expects more parameters than are passed (and provides no defaults)? In that case, this change would be safe as its adding arguments so existing handlers would work as expected.

includes/utils.php
helen:
```suggestion
			$create_missing_terms = apply_filters( 'dt_create_missing_terms', true, $taxonomy, $term_array, $term );
```

includes/utils.php
```diff
@@ -463,8 +463,10 @@ function set_taxonomy_terms( $post_id, $taxonomy_terms ) {
 			 * @since 1.0.0
 			 *
 			 * @param bool true Controls whether missing terms should be created. Default 'true'.
+			 * @param string    The taxonomy name.
+			 * @param array     Term data.
```
helen:
```suggestion
			 * @param array               Term data.
			 * @param WP_Term|array|false WP_Term object or array if found, false if not. 
```

includes/utils.php
```diff
@@ -463,8 +463,10 @@ function set_taxonomy_terms( $post_id, $taxonomy_terms ) {
 			 * @since 1.0.0
 			 *
 			 * @param bool true Controls whether missing terms should be created. Default 'true'.
+			 * @param string    The taxonomy name.
```
helen:
```suggestion
			 * @param string              The taxonomy name.
```

PR #384, opened 2019-06-05
Add ability to extend external connection details

avag-novembit:
Apply hook in the `meta_box_external_connection_details(..)` function so it would be possible to add 'external connection details' outside the plug-in (closes #383).

PR #386, opened 2019-06-05
Make menu content extendable

### LINKED ISSUE #385: Add hook to 'menu content'
Adding PR to apply hook in the `menu_content()` function.

avag-novembit:
Add `do_action( 'dt_add_to_menu_content', $post->ID, $connection_map )` hook to the `menu_content()` function (closes #385 ).

### DIFF
--- a/includes/push-ui.php
+++ b/includes/push-ui.php
@@ -510,6 +510,15 @@ syndicated<?php endif; ?>" data-connection-type="internal" data-connection-id="<
 							<?php esc_html_e( 'There was an issue distributing the post.', 'distributor' ); ?>
 						</div>
 					</div>
+					<?php
+					/**
+					 * Add neccessary markup/actions in menu content
+					 *
+					 * @param int $post->ID Post ID.
+					 * @param array Connection map.
+					 */
+					do_action( 'dt_add_to_menu_content', $post->ID, $connection_map );
+					?>
 
 				<?php else : ?>
 					<p class="no-connections-notice">


jeffpaul:

jeffpaul:
@avag-novembit it looks like this PR includes work from PR #384, mind stripping that duplicate stuff out so that this PR only includes code related to issue #385?

avag-novembit:
I see, I'll fix that.

adamsilverstein:
Hey @avag-novembit thanks for your contribution, can you add some context on how you foresee using this hook? I'm a bit leery about adding and supporting arbitrary HTML content inside our menu item or opening that as a supported option for developers.

avag-novembit:
> Hey @avag-novembit thanks for your contribution, can you add some context on how you foresee using this hook? I'm a bit leery about adding and supporting arbitrary HTML content inside our menu item or opening that as a supported option for developers.

I see. This hook is needed to make changes in menu item, in our case, to add a 'Redistribute' button to trigger the redistribute post functionality (https://prnt.sc/ocoii2).

adamsilverstein:
I will bring up this proposal for discussion with the team, for now if you need a button there I suggest trying to use JavaScript to insert the button you want to add. 

Can you explain the purpose of the 'Redistribute' button? Currently posts are redistributed whenever they are updated - that is, all distributed copies are updated, unless they have been "unlinked" at the remote end. Is the idea here a forced update even for unlinked posts?

arsendovlatyan:
@adamsilverstein we are trying prevent redistribution on post update and we want to do it with an add-on. 'Redistribute' button that we want to add will push updates right away, that's why we need it. 

However, it's not a must and your suggestion doing it with JavaScript can work, thanks a lot!

jeffpaul:
@arsendovlatyan so is your use case that the `Distribute` button only handles a one-time distribution with no additional updates to those remote sites and that the `Redistribute` button would keep those remote posts in sync with the origin post (akin to how the `Distribute` button usually happens)?

arsendovlatyan:
@jeffpaul in our case (it's pretty complex), we add posts (actually woo products) to website using REST API, so, when you hit "update" in admin, it's not the same post, as initially pushed by REST API.
We also use background tasks for initial push and to distribute updates later.
Sometimes, we notice issues (because of a bug in our code or something else) in distrusted post and need to push updates to fix it, we won't use "update" button, since it will change original post state, so, we are using "redistribute" button, which will just call "send_notification" function for that post.

jeffpaul:
@arsendovlatyan if you're able to utilize JS to insert the button you want to add as @adamsilverstein mentioned, then let's stick with that approach and close out this PR as no-longer-needed.  Can you confirm if you have the JS working for you?

arsendovlatyan:
@jeffpaul yes, we are good with the approach (using JS) at this moment, so, let's close the PR.

PR #411, opened 2019-06-14
get post from original id

### LINKED ISSUE #410: Add function to fetch post from 'original id'

Currently there is no function to fetch post using `dt_original_post_id` meta key stored in `{prefix}postmeta` table.

Going to send a PR added mentioned functionality in `utils.php` file.

**Designs**

Mentioned functionality is enwrapped in function `get_post_from_original_id(..)`:
    
    function get_post_from_original_id( $original_id ) {
	global $wpdb;
	return $wpdb->get_var( $wpdb->prepare( "SELECT post_id from $wpdb->postmeta WHERE meta_key = 'dt_original_post_id' AND meta_value = %s", $original_id ) );
    }

n/a

**Additional context**

n/a

Thanks.

Avag

avag-novembit:
Added function to get posts using original post id stored in `postmeta` table. closes #410.

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -786,3 +786,14 @@ function get_processed_content( $post_content ) {
 
 	return $post_content;
 }
+
+/**
+ * Get post in destination using original post id
+ *
+ * @param int $original_id Original post id.
+ * @return null|int
+ */
+function get_post_from_original_id( $original_id ) {
+	global $wpdb;
+	return $wpdb->get_var( "SELECT post_id from $wpdb->postmeta WHERE meta_key = 'dt_original_post_id' AND meta_value = '$original_id'" ); //phpcs:ignore
+}
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -795,5 +795,5 @@ function get_processed_content( $post_content ) {
  */
 function get_post_from_original_id( $original_id ) {
 	global $wpdb;
-	return $wpdb->get_var( "SELECT post_id from $wpdb->postmeta WHERE meta_key = 'dt_original_post_id' AND meta_value = '$original_id'" ); //phpcs:ignore
+	return $wpdb->get_var( $wpdb->prepare( "SELECT post_id from $wpdb->postmeta WHERE meta_key = 'dt_original_post_id' AND meta_value = %s", $original_id ) );
 }
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -793,7 +793,7 @@ function get_processed_content( $post_content ) {
  * @param int $original_id Original post id.
  * @return null|int
  */
-function get_post_from_original_id( $original_id ) {
+function get_post_id_from_original_id( $original_id ) {
 	global $wpdb;
 	return $wpdb->get_var( $wpdb->prepare( "SELECT post_id from $wpdb->postmeta WHERE meta_key = 'dt_original_post_id' AND meta_value = %s", $original_id ) );
 }


jeffpaul:

adamsilverstein:
@jeffpaul I'll take a look

adamsilverstein:
Hey @avag-novembit thanks for your contribution. 

I have a few questions about the Issue/PR:

* Can you describe your use case and why this should be in the plugin itself rather than handled in custom code?
* Can we address posts that are distributed using a network connection, meaning over a REST API connection not posts on the same network install?
* Can the functionality match the name `get_post_from_original_id` and return the post object not just an ID?
* Can you use `switch_to_blog` (using the `dt_original_blog_id`) and then use `get_post_meta` to get the id to avoid the direct database query?

adamsilverstein (changes requested):
left some questions and feedback

avag-novembit:
Hey @adamsilverstein, thanks for getting back. Let me try to answer the questions.

> * Can you describe your use case and why this should be in the plugin itself rather than handled in custom code?

The purpose of this pull request, and most of the my pull request, is to make the plug-in more extendable. It's used in add-ons (e.g. handle distribution of Advanced Custom Fields or WooCommerce products) which are in the development process currently. I think it'd be better to have a function inside the plug-in to provide API for add-ons, though open to other suggestions.

> * Can we address posts that are distributed using a network connection, meaning over a REST API connection not posts on the same network install?

Distributed to the so called external connections? If so, we can.

> * Can the functionality match the name `get_post_from_original_id` and return the full post object - (currently returns the post ID?)?

Agreed that it'd be better to call it `get_post_id_from_original_id(..)`, changed accordingly.

> * Can you use `switch_to_blog` (using the `dt_original_blog_id`) and then use `get_post_meta` to get the id to avoid the direct database query?

For the internal connections I think it would be possible, however for the external connections I believe it won't work out.

adamsilverstein:
> The purpose of this pull request, and most of the my pull request, is to make the plug-in more extendable.

If I understand this PR correctly, its purpose is _to find the post or posts that are distributed from a particular post_. EG you pass a post ID and the function would return all posts that (are on the same network install) and contain the `dt_original_post_id` reference back to the original (passed) post id.

To be genuinely useful as a plugin helper I would expect this function to work for any connection type: external or network (or others - connection types can be added). The current implementation only works with network connections.

Looking into the code I found that for the information you need is already tracked and available in the `connection map` stored in post meta on the original post under the key `dt_connection_map`. In this data point you will find a (serialized) array of the post IDs (along with which connection id and a timestamp). We use this data in part to ensure updates to the original post are propagated to distributed copies.

For example here is the data stored in a post I have distributed to both internal and external connections:

image

You can [search the codebase for usage](https://github.com/10up/distributor/search?q=dt_connection_map&unscoped_q=dt_connection_map) to see where we set up and use the connection map. 

I still feel like this PR could be useful for retrieving the list of distributed post ids from a post. I would consider adding a helper function that:

* Pulled data from the `dt_connection_map` meta and extracted the data in a usable format
* Returned the full distribution map: the list of remote connections / posts (post ids alone aren't that useful)
* Named to reflect its function, maybe `get_distributed_post_data`

avag-novembit:
Yes, the purpose of PR is to find the "child"/"dependent" post or posts that are distributed from a particular post. In essence the logic could be organized using `dt_connection_map`, written this way to be more convenient to operate with. Also, the functionality can be implemented in add-ons.

jeffpaul:

dkotter:
Going to close this out, though we can look at solving this differently if there's still desire for this functionality.

PR #417, opened 2019-06-24
Add hooks in push actions

### LINKED ISSUE #416: Add possibility to run 'push' actions in background

Currently the plug-in doesn't provide a way to run 'push' actions in background, which can be helpful in the cases when 'push' action has high resource consumption.

Added hooks in the `ajax_push(..)` and `send_notifications(..)` functions to make possible to delay function logic execution and run in background.

**Designs**

Created a PR.

n/a

**Additional context**

n/a

avag-novembit:
Makes possible to run post distribution and re-distribution in background, closes #416.

### DIFF
--- a/includes/push-ui.php
+++ b/includes/push-ui.php
@@ -542,7 +542,14 @@ syndicated<?php endif; ?>" data-connection-type="internal" data-connection-id="<
 
 					<div class="messages">
 						<div class="dt-success">
-							<?php echo esc_html( apply_filters( 'dt_successfully_distributed_message', esc_html__( 'Post successfully distributed.', 'distributor' ) ) ); ?>
+							<?php
+							/**
+							 * Filter distribution success message
+							 *
+							 * @param string Success message
+							 */
+							echo esc_html( apply_filters( 'dt_successfully_distributed_message', __( 'Post successfully distributed.', 'distributor' ) ) );
+							?>
 						</div>
 						<div class="dt-error">
 							<?php esc_html_e( 'There was an issue distributing the post.', 'distributor' ); ?>
--- a/includes/push-ui.php
+++ b/includes/push-ui.php
@@ -88,7 +88,52 @@ function ajax_push() {
 		exit;
 	}
 
-	$connection_map = get_post_meta( intval( $_POST['postId'] ), 'dt_connection_map', true );
+	// Maybe we need $_POST stored for background task
+	$params = $_POST;
+
+	if( ! wp_doing_cron() ) {
+		/**
+		 * Add possibility to send notification in background
+		 *
+		 * @param bool true     Whether send notification in background or not
+		 * @param int $post_id The post id
+		 */
+		$push_in_background = apply_filters( 'dt_push_allow_in_background', false, $params );
+
+		if ( true === $push_in_background ) {
+			wp_send_json_success(
+				array(
+					'results' => 'Success!!',
+				)
+			);
+
+			exit;
+		}
+	}
+
+	$result = push( $params );
+
+	wp_send_json_success(
+		array(
+			'results' => array(
+				'internal' => $result['internal_push_results'],
+				'external' => $result['external_push_results'],
+			),
+		)
+	);
+
+	exit;
+}
+
+/**
+ * Performs 'push' action
+ *
+ * @param array $params
+ *
+ * @return array
+ */
+function push( $params ) {
+	$connection_map = get_post_meta( intval( $params['postId'] ), 'dt_connection_map', true );
 	if ( empty( $connection_map ) ) {
 		$connection_map = array();
 	}
@@ -104,7 +149,7 @@ function ajax_push() {
 	$external_push_results = array();
 	$internal_push_results = array();
 
-	foreach ( $_POST['connections'] as $connection ) {
+	foreach ( $params['connections'] as $connection ) {
 		if ( 'external' === $connection['type'] ) {
 			$external_connection_type = get_post_meta( $connection['id'], 'dt_external_connection_type', true );
 			$external_connection_url  = get_post_meta( $connection['id'], 'dt_external_connection_url', true );
@@ -127,11 +172,11 @@ function ajax_push() {
 					$push_args['remote_post_id'] = (int) $connection_map['external'][ (int) $connection['id'] ]['post_id'];
 				}
 
-				if ( ! empty( $_POST['postStatus'] ) ) {
-					$push_args['post_status'] = $_POST['postStatus'];
+				if ( ! empty( $params['postStatus'] ) ) {
+					$push_args['post_status'] = $params['postStatus'];
 				}
 
-				$remote_id = $external_connection->push( intval( $_POST['postId'] ), $push_args );
+				$remote_id = $external_connection->push( intval( $params['postId'] ), $push_args );
 
 				/**
 				 * Record the external connection id's remote post id for this local post
@@ -149,7 +194,7 @@ function ajax_push() {
 						'status'  => 'success',
 					);
 
-					$external_connection->log_sync( array( $remote_id => $_POST['postId'] ) );
+					$external_connection->log_sync( array( $remote_id => $params['postId'] ) );
 				} else {
 					$external_push_results[ (int) $connection['id'] ] = array(
 						'post_id' => (int) $remote_id,
@@ -166,11 +211,11 @@ function ajax_push() {
 				$push_args['remote_post_id'] = (int) $connection_map['internal'][ (int) $connection['id'] ]['post_id'];
 			}
 
-			if ( ! empty( $_POST['postStatus'] ) ) {
-				$push_args['post_status'] = esc_attr( $_POST['postStatus'] );
+			if ( ! empty( $params['postStatus'] ) ) {
+				$push_args['post_status'] = esc_attr( $params['postStatus'] );
 			}
 
-			$remote_id = $internal_connection->push( intval( $_POST['postId'] ), $push_args );
+			$remote_id = $internal_connection->push( intval( $params['postId'] ), $push_args );
 
 			/**
 			 * Record the internal connection id's remote post id for this local post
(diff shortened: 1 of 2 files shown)


adamsilverstein:
Hi @avag-novembit, thank you for your contribution. 

Looking at the code, if the user returns true from `dt_push_allow_in_background`, normal 'push' behavior is bypassed. I don't see code for that here, are you working on that separately?

If we are going to enable async push requests, we probably want to include full support for the actual pushing as well, and proper UI to show the (background) progress.

includes/push-ui.php
```diff
@@ -88,7 +88,52 @@ function ajax_push() {
...
+			);
+
+			exit;
+		}
+	}
+
+	$result = push( $params );
+
+	wp_send_json_success(
+		array(
```
adamsilverstein: can we have `push` return the exact array we need and return it here directly?

includes/push-ui.php
```diff
@@ -498,13 +537,16 @@ function menu_content() {
...
+							<?php if ( $as_draft ) : ?>
+								<label class="as-draft" for="dt-as-draft"><input type="checkbox" id="dt-as-draft" checked> <?php esc_html_e( 'As draft', 'distributor' ); ?></label>
+							<?php endif; ?>
 						</div>
 					</div>
 
 					<div class="messages">
 						<div class="dt-success">
-							<?php esc_html_e( 'Post successfully distributed.', 'distributor' ); ?>
+							<?php echo esc_html( apply_filters( 'dt_successfully_distributed_message', esc_html__( 'Post successfully distributed.', 'distributor' ) ) ); ?>
```
adamsilverstein: Is this a new filter? if so, needs documenting.

The inner `esc_html__` can change to `__` - escaping should happen late, right before output.

includes/subscriptions.php
```diff
@@ -234,7 +234,25 @@ function delete_subscriptions( $post_id ) {
...
+	}
+
+	if ( ! wp_doing_cron() ) { // @codingStandardsIgnoreLine `wp_doing_cron(..)` is a WP function
+		/**
+		 * Add possibility to send notification in background
+		 *
+		 * @param bool  false   Whether run 'send notification' in background or not, default 'false'
+		 * @param array $params request data
+		 */
+		$send_notification_in_background = apply_filters( 'dt_send_notification_allow_in_background', false, $post_id );
```
adamsilverstein: this will affect all subscription updates for updated posts, these 'notifications' are required for updating pulled posts. This filter seems unrelated to the overall purpose of the PR, why is it included here?

includes/subscriptions.php
```diff
@@ -234,7 +234,25 @@ function delete_subscriptions( $post_id ) {
...
+
+		if ( true === $send_notification_in_background ) {
+			return;
+		}
+	}
+
+	if ( ! current_user_can( 'edit_post', $post_id ) ) {
```
adamsilverstein: can this stay as part of the logic check on L:237?

adamsilverstein (changes requested):
Left some feedback; if we enable this it should fully support background processing.

avag-novembit:
> Hi @avag-novembit, thank you for your contribution.
> 
> Looking at the code, if the user returns true from `dt_push_allow_in_background`, normal 'push' behavior is bypassed. I don't see code to handle actually pushing the content here, are you working on that separately?
> 
> If we are going to enable async push requests, we probably want to include full support for the actual pushing as well, and proper UI to show the (background) progress.

Hi @adamsilverstein, thanks for the comments.

Yes, 'Push in background' implementation is planned to be a separate add-on, and it will have its UI. Currently working on it.

includes/push-ui.php
avag-novembit: Changed accordingly.

includes/push-ui.php
avag-novembit: Changed accordingly.

includes/subscriptions.php
avag-novembit: We don't need this check when running in background as in cron there is no current in user.

includes/subscriptions.php
avag-novembit: I'll take a look at this.

arsendovlatyan: @adamsilverstein basically, we are trying to prevent send any REST API request during post update and move them into background. What if we will just add a filter, somewhere on the top and prevent send notifications, maybe near 
```
if ( ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) || wp_is_post_revision( $post_id ) ) {
		return;
	}
```

includes/subscriptions.php
```diff
@@ -238,24 +238,24 @@ function send_notifications( $post_id ) {
 		return;
 	}
 
-	if ( ! wp_doing_cron() ) { // @codingStandardsIgnoreLine `wp_doing_cron(..)` is a WP function
+	if ( ! wp_doing_cron() ) { //phpcs:ignore
+		if ( ! current_user_can( 'edit_post', $post_id ) ) {
```
adamsilverstein: Subscriptions are run in the cron context, will this user check work in cron? I suspect not.

includes/push-ui.php
```diff
@@ -546,7 +541,14 @@ function menu_content() {
 
 					<div class="messages">
 						<div class="dt-success">
-							<?php echo esc_html( apply_filters( 'dt_successfully_distributed_message', esc_html__( 'Post successfully distributed.', 'distributor' ) ) ); ?>
+							<?php
+							/**
+							 * Filter distribution success message
+							 *
```
adamsilverstein: add a sentence describing how this filter could be used

includes/push-ui.php
```diff
@@ -95,12 +95,12 @@ function ajax_push() {
 		/**
 		 * Add possibility to send notification in background
 		 *
-		 * @param bool  false   Whether run 'push' action in background or not, default 'false'
+		 * @param bool  true    Whether run 'push' action in background or not, default 'false'
 		 * @param array $params request data
 		 */
-		$push_in_background = apply_filters( 'dt_push_allow_in_background', false, $params );
+		$allow_push = apply_filters( 'dt_allow_push', true, $params );
```
adamsilverstein: add a sentence describing how this filter could be used

includes/push-ui.php
```diff
@@ -498,13 +532,23 @@ function menu_content() {
...
 					<div class="messages">
 						<div class="dt-success">
-							<?php esc_html_e( 'Post successfully distributed.', 'distributor' ); ?>
+							<?php
+							/**
+							 * Filter distribution success message
+							 *
+							 * @param string Success message
+							 */
+							echo esc_html( apply_filters( 'dt_successfully_distributed_message', __( 'Post successfully distributed.', 'distributor' ) ) );
```
adamsilverstein: this filter is related to the success message, can you add a sentence describing how it can be used/your use case

includes/subscriptions.php
```diff
@@ -234,10 +234,28 @@ function delete_subscriptions( $post_id ) {
...
+			return;
+		}
+
+		/**
+		 * Add possibility to send notification in background
+		 *
+		 * @param bool  true    Whether run 'send notification' in background or not, default 'false'
+		 * @param array $params request data
+		 */
+		$allow_send_notification = apply_filters( 'dt_allow_send_notifications', true, $post_id );
```
adamsilverstein: needs a description and better naming for the filter.

adamsilverstein:
@avag-novembit Thanks for your updates here. I left some additional feedback. I'm still not 100% clear on your use case and whether its something the code plugin should support directly, enable via filters or perhaps help you figure out how to implement using existing extensibility. 

When you have code ready that is close or complete, can you link to your project so we can better evaluate the value of these changes. Thanks!

includes/subscriptions.php
avag-novembit: Not sure what do you mean, there is no 'current user' in cron, so we need the `current_user_can(..)` check when 'is not doing cron'.

includes/push-ui.php
avag-novembit: In case the 'push' action is consuming a lot of resources, we can use the `dt_allow_push` filter to terminate function execution, and run the function in background. Here is simple use case example:

    add_filter( 'dt_allow_push', 'schedule_push_action', 10, 2 );
    
    function schedule_push_action( $allow_push, $params ) {
	if ( ! wp_next_scheduled( 'dt_push_posts_hook' ) ) {
		wp_schedule_single_event( time(), 'dt_push_posts_hook', [ $params ] );
	}

	return false;
    }

And then simply we can run it by hooking to scheduled event:

    add_action( 'dt_push_posts_hook', 'push_action', 10, 1 );

    function push_action( $params ) {
	\Distributor\PushUI\push( $params );
    }

Or we can call custom function depending on use case.

includes/push-ui.php
avag-novembit: In case when we are terminating 'push' function execution, it would be more descriptive to show another status message instead of "Post successfully distributed". For example: "Scheduled action to distribute post". Then show scheduled action status in a separate UI.

includes/subscriptions.php
avag-novembit: Well, we can change it to `dt_allow_subscription_update`, which is more descriptive perhaps.

By using this filter we can terminate function redistribution process and run it in background. It can be helpful in cases when resource consumption is high for that process. Here is a simple example how it can be implemented:

    add_filter( 'dt_allow_subscription_update', 'schedule_send_notifications', 10, 2 );
    
    function schedule_send_notifications( $allow_send_notifications, $post_id ) {
	if ( ! wp_next_scheduled( 'dt_redistribute_posts_hook' ) ) {
		wp_schedule_single_event( time(), 'dt_redistribute_posts_hook', [ $post_id ] );
	}

	return false;
    }

Here we are scheduling an event to run in background, and by returning `false` terminating post redistribution function execution. Then we can hook the post redistribution function to scheduled action or write a custom function to redistribute post.

    add_action( 'dt_redistribute_posts_hook', 'redistribute_post', 10, 1 );
    
    function redistribute_posts( $post_id ) {
	 \Distributor\Subscriptions\send_notifications( $post_id );
    }

avag-novembit:
@adamsilverstein thanks. Hope my last comments gives some idea how filters can be used.

jeffpaul:
@avag-novembit we're looking to solve a similar issue related to reports on ACF issues as part of our `2.1.0` release so I'm moving this PR and related issue to that release as well.

jeffpaul:

arsendovlatyan:
@jeffpaul we are maintaining this repo anymor, however, if you will decide to move forward, please  let me know if we will review it on our end.

jeffpaul:
Closing this PR because we're considering a broader implementation scope in #1306 (and also because we've sat on this PR long enough that it will be difficult to properly resolve merge conflicts).  Any one who has contributed to this PR and originally linked issue will be credited in any future implementation off 1306.

PR #419, opened 2019-06-28
Delete meta functionality

### LINKED ISSUE #284: Post meta is never actually deleted on the receiving end

avag-novembit:
In the `SubscriptionsController::receive_item(..)` method added logic to check for deleted meta data and, if found, delete it from website. closes #284

### DIFF
--- a/includes/classes/API/SubscriptionsController.php
+++ b/includes/classes/API/SubscriptionsController.php
@@ -197,7 +197,7 @@ class SubscriptionsController extends \WP_REST_Controller {
 	 */
 	public function receive_item( $request ) {
 		$post_id = (int) $request['post_id'];
-		$post = get_post( $post_id );
+		$post    = get_post( $post_id );
 		if ( empty( $post ) ) {
 			return new \WP_REST_Response( null, 404, [ 'X-Distributor-Post-Deleted' => 'yes' ] );
 		}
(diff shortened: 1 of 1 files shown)


jeffpaul:
I'm moving this to the `Future Release` milestone to be considered alongside a broad approach to how we handle meta removal/deletion from remote sites.  Apologies for the delay on response to this one @arsendovlatyan, but I want to be cautious with how we handle meta deletion.

jeffpaul:

dkotter:
With all the changes in Distributor since this PR was opened, going to close this out as it will need lots of refactoring to get it ready and probably easiest to open a new PR if this is something we still want to solve

PR #446, opened 2019-08-19
Enable filtering whether Distributor should update distributed post statuses when the origin post status changes

### LINKED ISSUE #112: Improve/verify handling of post status changes
### Updating post status 

What happens when a post that has been distributed is published, or changed from published to draft (or moved to the trash)

Opening this issue to discuss the expected behavior when a post that has been distributed has it's post status changed. 

Related: https://github.com/10up/distributor/issues/107 & https://github.com/10up/distributor/pull/110 & https://github.com/10up/distributor/pull/159 

### Expected behavior
After some internal discussion, we have decided that the expected behavior for post status changes is as follows:

**Distributing**:
- Pushing draft posts is disabled by default.
- Pushing drafts can be enabled by ~returning true from the `dt_drafts_can_be_distributed` filter~. [Update] This was replaced by the `distributable_post_statuses` [filter](https://github.com/10up/distributor/blob/master/includes/utils.php#L156).
- When a draft post is distributed, copies will always have a status of draft. The 'As Draft' option is not available when distributing drafts (showing it would be confusing).
- When a published post is distributed, copies will have a status of draft if the 'As Draft' checkbox is left checked, or 'publish' if it is unchecked.

**Updating**:
- When an origin published post is updated (or a draft published), no change is made to the status of distributed copies.This behavior can be overwritten by returning true from the (not yet implemented) `dt_published_posts_distribute_status` filter.  [Update 2023-11-29] The default behavior is that this is disabled/false.
- If the `dt_published_posts_distribute_status` is set to true and an origin post status is changed from publish to any other status (a "takedown"), or any non published post is updated ("embargo lifted"), the statuses of distributed copies are updated to match. Status could be trash, private, draft, autodraft, future, pending, etc.

**Deleting**
- When an origin post is deleted, remote posts are unlinked and the editor shows a warning message: `This post was distributed from [SITENAME]. However, the original has been deleted.` [Update 2023-11-29] This could be altered using the same `dt_published_posts_distribute_status` set to true to similarly delete the remote posts, but this should be disabled/false by default.

### Connections
- Post status updates should work the same for network and external connections

### Remaining Todo after testing
Add a `dt_published_posts_distribute_status` filter to determine if a posts's status should be distributed along with other post data.

adamsilverstein:
### Description of the Change

Filter whether Distributor should update distributed post statuses when the origin post status changes.

Closes https://github.com/10up/distributor/issues/112.

### Benefits

Enables distributing post status when desired, eg. takedowns.

### Verification Process

Code review without whitespace: https://github.com/10up/distributor/pull/446/files?w=1
 
**Repeat the following tests with a `Network Connection` and an `External Connection`**

- **Enable filter:**
`add_filter( 'dt_distribute_post_status', '__return_true' );`

 - Pull post subscription - pull published post then test changing origin post from published to draft also changes pulled copies status
 - Push post - push published post then test changing origin post from published to draft also changes pushed copies status

- **Disable filter:**
`add_filter( 'dt_distribute_post_status', '__return_false' );`

 - Pull post subscription - pull published post then test changing origin post from published to draft also changes pulled does not copy status
 - Push post - push published post then test changing origin post from published to draft also changes pushed does not copy status

### Unit tests
 - [x] Network Connection
   - [x] Pull
     - [x] Filter enabled
	 - [x] Filter disabled
   - [x] Push
     - [x] Filter enabled
     - [x] Filter disabled

 - [x] External Connection
   - [x] Pull
     - [x] Filter enabled
	 - [x] Filter disabled
   - [x] Push
     - [x] Filter enabled
     - [x] Filter disabled
 

### Checklist:


includes/classes/ExternalConnections/WordPressExternalConnection.php
```diff
@@ -556,7 +559,7 @@ public function push( $post_id, $args = array() ) {
 			'slug'                           => $post->post_name,
 			'content'                        => Utils\get_processed_content( $post->post_content ),
 			'type'                           => $post->post_type,
-			'status'                         => ( ! empty( $args['post_status'] ) ) ? $args['post_status'] : 'publish',
+			'status'                         => ( ! empty( $args['post_status'] ) ) ? $args['post_status'] : $distribute_status,
```
barryceelen: @adamsilverstein If I mark `$distribute_post_status` as true, the source post status will only be distributed if `$args['post_status']` is empty. `$args['post_status']` comes from the connection map for the origin post, how is that populated or changed? In any case, going off the filter name I would expect `$distribute_post_status` to always take precedence, but it's very possible I'm woefully missing context here.

adamsilverstein: In this case of push, the `post_status` comes from whether you have checked the 'distribute as draft' checkbox, if you have that setting overrides the filter. I do need to test subscriptions and pull, i'm not sure that works correctly. See https://github.com/10up/distributor/blob/b1dd60f9fdb6e22efd84a85807a6b7aec91e615e/includes/push-ui.php#L248-L250

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -100,10 +110,13 @@ public function push( $post_id, $args = array() ) {
 		if ( empty( $args['post_status'] ) ) {
 			if ( isset( $new_post_args['ID'] ) ) {
 
-				// Avoid updating the status of previously distributed posts.
-				$existing_status = get_post_status( (int) $new_post_args['ID'] );
-				if ( $existing_status ) {
-					$new_post_args['post_status'] = $existing_status;
+				if ( ! $distribute_post_status ) {
```
dinhtungdu: these nested `if` can be merged into one

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -68,14 +68,24 @@ public function push( $post_id, $args = array() ) {
...
 
+		/**
+		 * Filter whether Distributor should update post statuses when the origin post status changes.
+		 *
+		 * False by default, return true to have post statuses distributed.
+		 *
+		 * @since 2.0.0
+		 * @hook dt_distribute_post_status
+		 */
+		$distribute_post_status = apply_filters( 'dt_distribute_post_status', false );
```
dinhtungdu: we're using this filter in two places, I suggest replacing them with a function in `Distributor\Utils`

tests/wpacceptance/includes/TestCase.php
```diff
@@ -166,4 +166,82 @@ protected function dismissNUXTip( $actor ) {
...
+	 * Create an external connection.
+	 *
+	 * @param \WPAcceptance\PHPUnit\Actor $actor The Actor instance.
+	 */
+	protected function createExternalConnection( $actor ) {
+		$actor->moveTo( 'wp-admin/post-new.php?post_type=dt_ext_connection' );
+		$actor->typeInField( '#title', 'Test External Connection' );
+		$actor->typeInField( '#dt_username', 'wpsnapshots' );
+		$actor->typeInField( '#dt_password', 'password' );
+		$actor->typeInField( '#dt_external_connection_url', $this->getWPHomeUrl() . '/two/wp-json' );
```
dinhtungdu: we need to create a connection to `one`, not the reverse.

for the current test, we create a connection on site `one` to site `two` but pull a post from site `one` through internal network connection. The posts pulled by external connections always has `draft` status.

dinhtungdu (changes requested):
The filter works well. Distributed posts get the status updated when the origin changes. Worked with both internal and external connections with Classic editor enabled.

For Gutenberg, there is a small issue with pull: blocks content get converted to classic, then the page load `admin-syndicated-post.min.css` instead of gutenberg one => can't publish the post after pulling.

Added some comments about acceptance tests and code readability.

dinhtungdu:
@jeffpaul, I fixed the test, only one test is failing because of #516, waiting for #534 to pass WPA tests

jeffpaul:
In reviewing open PRs in the 1.6.0 milestone, this PR seemed still too risky and in need of testing/confirmation before merging so I'm going to punt this to our next release in hopes of having things in a more stable state by then.

jeffpaul:

Issue #472, opened 2019-11-01
Add a 'syndicatable' capability filter to the push UI


pragmatic-tf:
**Is your enhancement related to a problem? Please describe.**
The Distributor plugin provides a way to push/syndicate content out to any number of sites from either the edit post screen or when logged in and viewing a post on the front-end.

However, there are some use cases where it might not be desirable for the user who wants to push to another site (or sites) to also have the capability to edit the original post.

**Describe the solution you'd like**
Access to the push UI is currently based on the `edit_posts` capability but as there isn't a filter on this capability check I'd suggest simply adding one to the check found in the `syndicatable()` function. 

This will allow plugins to override the capability required to either a different core capability or to add their own custom capability.

**Designs**
N/A

**Describe alternatives you've considered**
There are capability filters which would allow users without post edit permissions to **pull** content, but the pull and push functionality may have different applications or use cases as their UI and user flow is slightly different. Adding a new filter to the push capability check gives more options for extending this plugin and is relatively low impact.

jeffpaul:
@pragmatic-tf welcome to Distributor and thanks for the feedback!  I'll review this with the team and get back to you with any questions, noting that we're aiming for a v2.0 release later this year so if you have time to submit a related PR for more detailed review then that'd be warmly welcomed as well.  Thanks!

pragmatic-tf:
Thanks for the welcome @jeffpaul! 

I've added the above related PR for further review. I wasn't too sure if the included comment docblock counted as documentation for the PR tasks section so I've left that blank for now. More than happy to update any documentation as required, or check it off if that's already covered.

jeffpaul:
Closed via #473.

PR #473, opened 2019-11-05
add `dt_syndicatable_capabilities` filter to push ui

pragmatic-tf:
### Description of the Change

Adds a new filter to the user capability check found in the `syndicatable()` function, which controls access to the Push UI in the WP admin bar.

### Alternate Designs

N/A

### Benefits

This will allow other plugins to dynamically update the capability required to enable the push UI in the WP admin bar.

### Possible Drawbacks

None that I can foresee - the code addition doesn't alter any existing functionality by itself.

### Verification Process
After applying the code change, I tested the filter on a local development install by changing the capability to allow different user roles access to the Push UI.

- Changing the capability to `read` allowed a test Subscriber to use the Push UI
- Changing the capability to `edit_published_posts` prevented a test Contributor from using the Push UI but still allowed a test Author to use it.
- Changing the capability to a custom capability - `dt_test_cap` - prevented any users from accessing the Push UI, except super-admins.
- Adding the `dt_test_cap` custom capability to the Author role, then allowed a test Author to access the Push UI.

### Checklist:

### Applicable Issues
See: #472


pragmatic-tf:
Just checking in on this PR - looks like the CI build is failing but I'm not too sure why... Is there anything I can do (or need to change) to help it pass the appropriate checks?

I also popped a question on issue #473 about documentation - is the inline code doc block enough or is there another place I should document the new filter?

jeffpaul:
@pragmatic-tf the doc block _should_ be sufficient, as once that gets merged it should appear on our technical documentation site: https://10up.github.io/distributor/.  As for the failing build, I'm uncertain what's going on there and will see if we can figure that out while reviewing this PR.  Thanks for checking in and for the PR, happy holidays!

dinhtungdu (commented):
`capabilities` is a plural word, so from the filter name, I would expect passing multiple capabilities to this filter. We have two options here:

- change the filter name to the singular
- or change the code to handle multiple capabilities

cc @jeffpaul

pragmatic-tf:
Hey @dinhtungdu 👋 totally get where you're coming from, and it's a good point.

 Use of the plural was in order to stay consistent with a few other capability filters in the rest of the plugin codebase:
`dt_capabilities`
`dt_external_capabilities`
`dt_pull_capabilities`
`dt_push_capabilities`

☝️ The above all handle single capability filters, but use the plural, so it all depends on whether the preference is to maintain consistency or go for semantic correctness? More than happy to update the PR for either.

It's worth noting that `current_user_can()` doesn't accept multiple capabilities so that's not a possible solution unfortunately.

dinhtungdu (approved):
@pragmatic-tf thanks for the head up, I haven't checked the naming convention, so I'm changing my review now.

About the multiple capabilities, we can do it by looping through the array and apply `current_user_can` for each but it will break the current convention, so let's ignore it for now.

PR #474, opened 2019-11-05
Update build-docs.yml

adamsilverstein:
Attempting to get the docs build working correctly, fixes some formatting errors in the config file.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   build and deploy:
     runs-on: ubuntu-latest
-    node-version: [12.x]
+    node-version: '12.x'
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/build-docs.yml
```diff
@@ -25,4 +25,4 @@ jobs:
       uses: maxheld83/ghpages@v0.2.1
       env:
         BUILD_DIR = "docs/"
-        secrets = ["GH_PAT"]
+      secrets: ${{ secrets.GH_PAT }}
```
adamsilverstein: This may need to be indented (part of env:) - this is based on https://github.com/maxheld83/ghpages/blob/master/README.md where it isn't in env, but i think that may be wrong

adamsilverstein: changed

helen (approved):
Looks reasonable, let's give it a shot!

Issue #488, opened 2019-11-19
Not handled response Bad Request from spoke site.


grogou:
**Is your enhancement related to a problem? Please describe.**
When I'm distributing post to a website, which actually is down, I still get a "post successfully distributed" message.

**Describe the solution you'd like**
Currently, we are handling only the case, when response  `status code` is `400`, `Bad request`, In other cases, for example, 403, 504, etc we will show "post successfully distributed", I suggest to handle them as an error like `Bad request and show error message, this will also allow retrying distribution later.

**Designs**
n/a

**Describe alternatives you've considered**
I think it's even better to use post_id, to identify if the post was successfully distributed or not, so, if we have received post_id on target, we can say distribution was successful, otherwise, show error message.

**Additional context**
Tested for external connections only.

jeffpaul:
@eriktad welcome to Distributor and thanks for the feedback!  I don't know that we can rely on post_id matching on the destination site(s), but otherwise would be open to a PR on this in a future release if you're able to craft a solution for this?

grogou:
@jeffpaul Thank you too!
@tlovett1 
In the case, where the response body is not empty and I have only  
```
["data"]=>
  {
    ["status"] => 400
  }
```
this is a case if the received response is `Bad request`, This try/catch is not working if PHP notice and warnings switched off.

https://github.com/10up/distributor/blob/025cc7ea5e8e693ad35f676809cd95e2e093cb78/includes/classes/ExternalConnections/WordPressExternalConnection.php#L650-L654

 PHP works without an undefined index notice, and we don't receive an exception, we receive `remote_id = NULL`.

dinhtungdu:
@eriktad Can you add instruction to reproduce this issue. I tried to distribute post to a down site but I couldn't reach the try-catch block at line 650, the `push` function got returned at line 537.

jeffpaul:
@eriktad can you confirm your steps to reproduce this issue as we've been unable to reproduce it?  I'm moving this to `Future Release` until we're able to reproduce the issue and if so then determine a resolution.  I agree that if we are able to reproduce this that its something we should address, its just not something we've been able to reproduce as of yet.

arsendovlatyan:
@dinhtungdu you are right, taking entire site down, won't bring up this issue, however, here is how I have reproduced it just now.
- I went to destination website and changed "distributor" user (which I'm using for external connection) permission to "subscriber", so, it won't be able to create posts
- opened the source website and pushed a post, it shows "success"message 
[...]

PR #492, opened 2019-11-26
Handle hub bad request issue

grogou:
### Description of the Change
Related to #488 
Removed try/catch case and added a new condition to check we receive `remote_id` or not

### Alternate Designs
n/a

### Benefits
We can handle the case, were not distributed the post ( really not distributed but shown like distributed in the hub ) with `post_id = 0` can't distribute because spoke is down or we don't receive some remote post id or we have a bad request.

### Possible Drawbacks
We don't have drawbacks, because we fix a bug.

### Verification Process

I have created an environment for test, downed spoke site and hub for distribution, My code tested up and work clean. This case I see in my live sites where my spoke site go down some time.

### Checklist:

### DIFF
--- a/includes/classes/ExternalConnections/WordPressExternalConnection.php
+++ b/includes/classes/ExternalConnections/WordPressExternalConnection.php
@@ -647,11 +647,15 @@ class WordPressExternalConnection extends ExternalConnection {
 
 		$body_array = json_decode( $body, true );
 
-		try {
-			$remote_id = $body_array['id'];
-		} catch ( \Exception $e ) {
+		if( isset( $body_array['data']['status'] ) && $body_array['data']['status'] == 400 ){
+			return new \WP_Error( 'bad-request', esc_html__( 'Status code is 400', 'distributor' ) );
+		}
+
+		if( !isset( $body_array['id'] ) ){
 			return new \WP_Error( 'no-push-post-remote-id', esc_html__( 'Could not determine remote post id.', 'distributor' ) );
 		}
+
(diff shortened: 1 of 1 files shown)


jeffpaul:
@eriktad thanks for the PR (and related issue), we'll get this through review and back to you with any questions or concerns.

includes/classes/ExternalConnections/WordPressExternalConnection.php
```diff
@@ -647,12 +647,12 @@ public function push( $post_id, $args = array() ) {
 
 		$body_array = json_decode( $body, true );
 
-		try {
-			$remote_id = $body_array['id'];
-		} catch ( \Exception $e ) {
+		if( !isset( $body_array['id'] ) ){
```
dinhtungdu, line 650: we have some cs issues here (spacing)

jeffpaul:
@eriktad can you confirm your steps to reproduce this issue as we've been unable to reproduce it?  I'm moving this to `Future Release` until we're able to reproduce the issue and if so then determine a resolution.  I agree that if we are able to reproduce this that its something we should address, its just not something we've been able to reproduce as of yet.

arsendovlatyan:
@eriktad  Can you help @jeffpaul and his team to reproduce this issue?

I guess if we will just add "deny from all" in .htaccess of destination website, which will trigger 403 (permission denied) issue and try to push a post already existing external connection, we won't get "Could not determine remote post id." error message, which is the expected behavior.  Can you confirm this ?

Issue #497, opened 2019-12-06
Shortcodes within plugins are not rendered when pulling content within multisite


petenelson:
When on a network/multisite install and you pull content from one site to another, shortcodes within plugins from the selected site are not being rendered before inserting the post into the current site. The raw shortcode gets added to the post content.  Pushing content from one site to another is not affected as the push functionality runs `the_content` filter before pushing the post.

This is due to some limitations around WP core. For a network pull, the code switches to the selected site via `switch_to_blog()` to get the post. Because switching blogs does not activate any plugins, and if the shortcode is within a plugin, running `the_content` filter would still not have any effect.

A possible solution would be to make an HTTP call to the selected site's REST API endpoint right after copying the post. This exposes the rendered content of the post, and we could update the current site's post with the rendered content.

Because I need to get this working for a client's site, I'll likely take that code and refactor it for a patch for Distributor.

jeffpaul:
Thanks again for the heads up on this @petenelson even more so in offering to craft a PR to resolve this! 🙏

jeffpaul:
Closed via #498.

PR #498, opened 2019-12-06
Fix/pull apply the content

petenelson:
### Description

Make a REST request for rendered post content after pulling a post from a network connection.

### Benefits

Allows the pulled post to get rendered shortcodes and anything else updated by `the_content` filter.

### Possible Drawbacks

The additional HTTP request has a small performance impact.

### Checklist:

### Applicable Issues

#497


jeffpaul:
@dkotter as you have availability, it would be great to get this through review so we can make any necessary changes and get this to a merge (and potential Distributor minor release to resolve this).

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -267,6 +267,12 @@ public function pull( $items ) {
 
 			restore_current_blog();
 
+			// Allow the sync'ed post to be updated via a REST request to
```
dkotter: @petenelson If you could change these code comments into an actual DocBlock, similar to how most of the other actions are documented (and that should follow https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters), that would be great

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -934,4 +940,53 @@ public static function wpseo_og_url( $og_url ) {
...
+		$rest_url = false;
+		if ( ! empty( $original_blog_id ) && ! empty( $original_post_id ) ) {
+			$rest_url = Utils\get_rest_url( $original_blog_id, $original_post_id );
+		}
+
+		if ( empty( $rest_url ) ) {
+			return;
+		}
+
+		$request = apply_filters( 'dt_update_content_via_request_args', [], $new_post_id, $this );
```
dkotter, line 1009: Same here, if we can add DocBlock here, that would be great

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -934,4 +940,53 @@ public static function wpseo_og_url( $og_url ) {
...
+
+		$response = wp_remote_get( $rest_url, $request );
```
dkotter: To support VIP, I think we'll need to enhance this a bit. Basically we need to check if we are on WordPress.com VIP and if so, use the `vip_safe_wp_remote_get` function. Can see examples of this here: https://github.com/10up/distributor/blob/develop/includes/classes/ExternalConnections/WordPressExternalConnection.php#L184

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -934,4 +940,53 @@ public static function wpseo_og_url( $og_url ) {
...
+
+		$body = wp_remote_retrieve_body( $response );
```
dkotter: We'll want to add a few more checks here, just to be safe. Basically check the `$response` value isn't a WP Error and check that the response code (`wp_remote_retrieve_response_code`) matches what we want. Can see examples of that here: https://github.com/10up/distributor/blob/develop/includes/classes/ExternalConnections/WordPressExternalConnection.php#L200

includes/utils.php
```diff
@@ -783,3 +783,31 @@ function get_processed_content( $post_content ) {
...
+
+	$obj       = get_post_type_object( $post->post_type );
+	$rest_base = ! empty( $obj->rest_base ) ? $obj->rest_base : $obj->name;
+	$base      = sprintf( '%s/%s', 'wp/v2', $rest_base );
+
+	$rest_url = rest_url( trailingslashit( $base ) . $post->ID );
+
+	restore_current_blog();
+
+	return apply_filters( 'dt_get_rest_url', $rest_url, $blog_id, $post_id );
```
dkotter, line 876: Should have a DocBlock on this filter as well

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -267,6 +267,12 @@ public function pull( $items ) {
 
 			restore_current_blog();
 
+			// Allow the sync'ed post to be updated via a REST request to
+			// get the rendered content.
+			if ( apply_filters( 'dt_pull_post_apply_rendered_content', true, $new_post_id, $this, $post_array ) ) {
```
dkotter: Not a CR comment but wondering if we want to default this to `false`? So sites that need the extra step of pulling in rendered content can opt-in but this won't impact other setups that don't need this? cc/ @jeffpaul

petenelson: I was thinking the same thing, with an eventual option in the settings to turn that off without a code change.

jeffpaul: My only concern is we're currently doing this for pushed posts so I imagine the perception is that we'd do this by default for pulled posts as well.  I _think_ I'd rather support the broader case and allow sites that know they don't need to render content first before pulling could adjust the default to `false` and gain back whatever performance the HTTP call is impacting.  Unless I'm misunderstanding things or not fully thinking through other scenarios, if so please call those out.

includes/classes/InternalConnections/NetworkSiteConnection.php
petenelson: FYI, `wp_remote_retrieve_body()` already checks for WP_Error.

dkotter: Ah, didn't realize that. I still like the explicit check, so we can output a different error message in each instance, to better help with debugging issues.

dkotter (dismissed):

jeffpaul:
@petenelson looks like there's a merge conflict on this, could you help resolve that so we can look to merge this in?

petenelson:
Merge conflict has been resolved

dkotter (approved):

Issue #512, opened 2020-02-03
The `is_using_gutenberg` function doesn't always work


dkotter:
**Describe the bug**
In Distributor, we have a utility function that will let you know if a particular post is using Gutenberg, `is_using_gutenberg`. This checks a few things:

1. Is either the Gutenberg plugin being used or are we on a version of WordPress that comes with Gutenberg?
1. If so, then checks if the post has Gutenberg content (does this by looking for the `<!-- wp:` string)

I've run across a scenario where that string doesn't exist, so this function returns false even though the post is using Gutenberg. I've seen this when the post was originally written in the Classic Editor and then switched to use Gutenberg. The post content is dumped in the Classic block but it doesn't contain the string we are looking for.

This causes a few issues, the main one being the post can't be published, as the Publish button is disabled via CSS (and because we think the post isn't using Gutenberg, we don't load the proper CSS file that overrides this).

There may be other scenarios where post content won't have the proper Gutenberg tags but this is the only one I've run across so far.

**Steps to Reproduce**
1. Activate the Classic Editor plugin
2. Write a new post using the Classic Editor
3. Disable this plugin
4. Distribute this post to another site
5. Notice the Publish button is disabled

**Expected behavior**
If possible, have a better check to see if a post is using Gutenberg, so we load everything correctly.

PR #514, opened 2020-02-05
Replace has_blocks with use_block_editor_for_post in is_using_gutenberg

johnwatkins0:
### Description of the Change

This would fix the issue described in #512 by providing a more reliable check for whether the Gutenberg editor is enabled for a post. 

The issue @dkotter found comes from a specific scenario where a post may be distributed to a Gutenberg-enabled site without having `<!-- wp:` in the post content. In this case, the core function `has_blocks` returns false because checking for `<!-- wp:` [is exactly what `has_blocks` does](https://github.com/WordPress/WordPress/blob/317465e2feb965ea7d86529e54908b3fbea539a8/wp-includes/blocks.php#L56-L65).

This replaces `has_blocks` with `use_block_editor_for_post`, which checks whether the current post's post type has the editor enabled and shows in REST. This works with the Classic Editor plugin because that plugin [uses the `use_block_editor_for_post_type` filter to have this function always return false](https://github.com/WordPress/classic-editor/blob/2099a42394710fdd717589da789a3404aa6e6eb6/classic-editor.php#L104)

### Alternate Designs

I can't think of any good alternatives to this small change.

### Benefits

After being distributed in draft status, posts with no blocks can be published.

#### Before

Adapting Darin's steps to reproduce from #512:

1. Activate the Classic Editor plugin
2. Write a new post using the Classic Editor
3. Disable the Classic Editor plugin
4. Distribute the post to another site, keeping the post in draft status on the destination site
5. Notice on the destination site the Publish button is disabled

#### After

Following the above steps, the Publish button will now be enabled on the destination site

### Checklist:

**Note**: There were multiple issues with tests. I was unable to run `composer install` because of missing PHP extensions on my system (can share more details if helpful). I see tests are failing on all the PRs, so I'm going to assume that's not just me.

### Applicable Issues

#512 

### Changelog Entry

Replace `has_blocks` with `use_block_editor_for_post` in is_using_gutenberg check

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -46,17 +46,13 @@ function is_using_gutenberg( $post ) {
 		return false;
 	}
 
-	// WordPress 5.0 introduces the has_blocks function.
-	// We pass `post_content` directly because `has_blocks()`
-	// tries to retrieve a local post object, but this may be a remote post.
-	if ( function_exists( 'has_blocks' ) ) {
-		return has_blocks( $post->post_content );
+	if ( function_exists( 'use_block_editor_for_post' ) ) {
+		return use_block_editor_for_post( $post );
 	} else {
-		// This duplicates the check from `has_blocks()` as of WP 5.2.
+		// This duplicates the check from `use_block_editor_for_post()` as of WP 5.2.
 		return false !== strpos( (string) $post->post_content, '<!-- wp:' );
 	}
 }
-
 
 /**
  * Get Distributor settings with defaults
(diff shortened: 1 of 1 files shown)


includes/utils.php
```diff
@@ -351,6 +347,7 @@ function prepare_meta( $post_id ) {
 				$meta_value = maybe_unserialize( $meta_value );
 				/**
 				 * Filter whether to sync meta.
+				 *
```
johnwatkins0, line 350: This is PHPCS - "There must be exactly one blank line before the tags in a doc comment"

dkotter (approved):

dkotter:
@johnwatkins0 Thanks for this! Tested this and works great, fixes the issue I had run into.

jeffpaul:
May also resolve #516

PR #515, opened 2020-02-06
Ensure connection check does not create empty posts

### LINKED ISSUE #255: Empty posts created when pushing
When post/s are pushed from site A to site B many empty posts are created, as well as the intended post/s, empty posts are listed as:

(no title) — Draft

WordPress version : 4.9.8
Plugins installed : 'Distributor' and 'Application Passwords'

johnwatkins0:
### Description of the Change

This would resolve the issues recently noted on #255, particularly: 

> On wp-admin/admin.php?page=distributor, clicking the (Verify) link for an External Connection will generate a single (no title) — Draft Page on that external site.
On wp-admin/post.php?post=###&action=edit (where ### is a number for a valid External Connection), clicking the Update Connection button will generate two (no title) — Draft Pages on that external site.

(In the first case, the empty page was actually being created each time the single External Connection post was loaded, not just when accessed via the (Verify) link.)

I'm not sure whether this fix would resolve the older comments in #255, though. We don't have enough details for those.

At [this point](https://github.com/10up/distributor/blob/2e8e0779a4dd83739c78317d7faa5f331f7c36d7/includes/classes/ExternalConnections/WordPressExternalConnection.php#L846-L855) in the `check_connections` method that runs on `save_post` and in an AJAX request in the External Connection post type editor, the plugin iterates through post types and makes a REST POST request for each one. If the POST request authenticates, that post type is marked as one that can receive post requests.

The problem with this is that the `page` and `wp_block` post types have no required fields, so their REST POST endpoints create an empty post even when no data is set. 

To fix this, I added `id => -1` to the request body sent to the POST endpoint, because this will cause the request to be invalid. It still authenticates, though, which is all we need to know.

**Note:** Before debugging this, I tested on the `feature/auth-flow` branch (#368) and confirmed that that update did not fix the issue.

### Alternate Designs

Other ways of invalidating the POST request body while still authenticating could be explored, although in my experience there's no circumstance in which passing `id => -1` will result in a valid request.

### Benefits

Empty pages and wp_blocks (and maybe custom post types) will no longer be created when the External Connection post type editor is visited or updated.

#### Before

Replicate the test scenarios quoted above **using separate WordPress instances**.  When visiting the External Connection editor page, or when updating it, if you have a valid external connection, you will see empty pages created on the remote site. Empty `wp_block` posts will also appear in the database. 

#### After

Visiting the External Connection editor page for an external connection with a valid connection will not result in empty pages anymore.

### Checklist:

### Applicable Issues

Maybe resolves #255 

### Changelog Entry

Resolve issue where empty pages and `wp_block` posts were created on remote sites while verifying external connections

### DIFF
--- a/includes/classes/ExternalConnections/WordPressExternalConnection.php
+++ b/includes/classes/ExternalConnections/WordPressExternalConnection.php
@@ -855,7 +855,7 @@ class WordPressExternalConnection extends ExternalConnection {
 										// this runs.
 										'body'    => array(
 											'test' => 1,
-											'id' => -1
+											'id'   => -1
 										),
 									)
 								)
--- a/includes/classes/ExternalConnections/WordPressExternalConnection.php
+++ b/includes/classes/ExternalConnections/WordPressExternalConnection.php
@@ -855,7 +855,7 @@ class WordPressExternalConnection extends ExternalConnection {
 										// this runs.
 										'body'    => array(
 											'test' => 1,
-											'id'   => -1
+											'id'   => -1,
 										),
 									)
 								)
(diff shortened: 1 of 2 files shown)


dinhtungdu (commented):
This PR works, but there is an opening PR (#245) that refactor the `check_connections` method that scales better than the current implementation, let's take a look at that PR before this

jeffpaul:
Closing in favor of #245.

PR #517, opened 2020-02-07
Add: Debug info to Site Health Info screen

### LINKED ISSUE #407: Add debug info to Site Health Info screen
WordPress 5.2 introduced the Site Health Info screen.  Adding Distributor-specific debug info there can help solve issues that come to us from support forums, GitHub, email, etc. as users can copy this info and send to us for faster assistance with issues.

- [x] display Distributor version and settings info
- [ ] display list of (internal & external) connections and their status/health/Distributor version (for external connections)
- [ ] display Application Passwords version (if its installed though we may not need this and are probably better served pushing a PR to their repo to add that to Site Health Info)

**Designs**
n/a

n/a

**Additional context**
n/a

dinhtungdu:
### Description of the Change
Add debug information of Distributor to Site Health Info screen.

### Alternate Designs
n/a

### Possible Drawbacks
n/a

### Verification Process
Go to Tools > Site Health > Info and check the Distributor section.

### Checklist:

### Applicable Issues
Relates to #407

### Changelog Entry

Added: Site Health debug information.


includes/debug-info.php
```diff
@@ -0,0 +1,160 @@
...
+
+	return $info;
+}
+
+/**
+ * Get and format internal connections.
+ *
+ * @return array
+ */
+function get_formatted_internal_connnections() {
```
dinhtungdu, line 127: We have duplicated code here (and the next function) and in the `pull-ui.php`. I'm considering splitting it and move the logic to `utils.php`.

jeffpaul:
@dinhtungdu some feedback/question from my testing:

1. Change `Valid license` to `Valid registration`.
2. Change `Email` to `Registration email`.
3. Ensure `Settings` notes details for `Override Author Byline` and `Media Handling`.
4. Is there any better formatting we can apply to the `Internal Connections` and `External Connections` sections besides the current blob of data?

dinhtungdu:
@jeffpaul I fixed 1 and 2, 4.
For 3, do you mean the field title and description?

jeffpaul:
@dinhtungdu yes, we'll want to log how users have those settings configured.

dinhtungdu:
@jeffpaul thanks for the head up, this PR is ready for review again

jeffpaul:
@dinhtungdu 1-3 look good.  For 4, can we call out the `Title`, `URL`, `Status`, `Auth method`, `Username`, `Roles Allowed to Push` labels and their associated value for each External Connection plus the additional data you're capturing?  I still need to test on a multisite to see how things compare for Internal Connections.

dinhtungdu:
@jeffpaul, Thanks for the feedback, I updated both internal and external connections data, see https://prnt.sc/rkhnbq.

Please note that you need to run `npm run build` to build the new stylesheet for Site info screen.

includes/debug-info.php
```diff
@@ -0,0 +1,245 @@
...
+		return;
+	}
+	wp_enqueue_style( 'dt-site-health', plugins_url( '/dist/css/admin-site-health.min.css', __DIR__ ), [], DT_VERSION );
+}
+
+/**
+ * Setup actions and filters
+ *
+ * @since 2.0.0
+ */
```
johnwatkins0: Stray docblock here.

includes/debug-info.php
```diff
@@ -0,0 +1,245 @@
...
+/**
+ * Add distributor debug information to Site Health screen.
+ *
+ * @see WP_Debug_Data::debug_data
+ * @filter debug_information
+ *
+ * @param array $info The full array of site debug information.
+ * @return array Filtered debug information.
+ *
+ * @since 2.4.0
```
johnwatkins0: 2.0.0?

includes/debug-info.php
```diff
@@ -0,0 +1,245 @@
...
+	return $info;
+}
+
+/**
+ * Get and format internal connections.
+ *
+ * @return array
+ */
+function get_formatted_internal_connnections() {
+	if ( empty( \Distributor\Connections::factory()->get_registered()['networkblog'] ) ) {
```
johnwatkins0: Not a blocker, but I see PHP `use` being used elsewhere in the plugin. Why not use it in this file?

johnwatkins0 (changes requested):
A few very minor comments, but otherwise code looks good and it's working well locally.

includes/debug-info.php
dinhtungdu: I don't think I get your idea :D, do you mind explaining that again 😄?

johnwatkins0: For example, you could put `use Distributor\Connections;` at the top of this file, and make lines like this a little less cluttered:
```PHP
<?php

use Distributor\Connections;

... 

if ( empty( Connections::factory()->get_registered()['networkblog'] ) ) {
```

dinhtungdu: Oh, I see now, what a silly question 😳! Updated. Thank you!

johnwatkins0 (dismissed):
Looks great!

jeffpaul:
@dinhtungdu are we able to detect the Distributor version for the External Connections and report that for those in the respective section?

dinhtungdu:
@jeffpaul With current codebase, no, but we can with a custom public REST API endpoint.

jeffpaul:
While it would be nice to be able to confirm all Distributor versions are current/equal, I'm not sure whether the value there is high enough to warrant a public API... @dkotter  what's your take?

dinhtungdu:
@jeffpaul That's my concern, too.

dkotter:
@jeffpaul @dinhtungdu This is something that was added as part of the Auth Setup Wizard PR (see https://github.com/10up/distributor/pull/368/files#diff-923616c19cf0f150ddc8b464ca53da19R257). So if that stays in (and we get that merged at some point) we will be able to utilize that here as well.

jeffpaul:
@dinhtungdu let's wait on this PR until #368 gets merged and you can build on top of the API then.

dinhtungdu:
Added version to debug info. But the current solution may not scale if the site has many connections.

I thought about caching or saving version info to `dt_external_connections` meta as well, but if we do so, we can't ensure that the versions in debug info are up-to-date.

What do you think @dkotter ?

jeffpaul:
Unless we have alternatives that clearly scale better, I'd say we get this merged in to benefit anyone triaging Distributor setups and then work to also include the Distributor version as a column in the External Connections list table (as previously suggested by @dinhtungdu).

dkotter:
@dinhtungdu @jeffpaul Yeah, not sure there's a great answer here. Ideally, this health info screen isn't accessed very often, so not sure how much of a concern this will be in reality. But for sites that have hundreds of external connections, this will add a significant overhead to that page.

We could take the same approach we did for the Push UI, where we don't load anything on page load, only when the menu is hovered into. So instead of querying for the version information when the health info screen is loaded, we do it after page load, so it doesn't hurt performance.

Not sure if that's worth the effort here or not but caching won't help a ton, as the data won't be cached until the page is visited (and again, hopefully this page isn't visited often) and as mentioned, saving to meta can become out-of-date.

dinhtungdu:
@jeffpaul After considering some options, I incline to merge this PR. I added a filter to disable the debug info as well.

dkotter (approved):

PR #518, opened 2020-02-08
Run NetworkSiteConnection::update_syndicated on correct hook when using Gutenberg

### LINKED ISSUE #464: NetworkSiteConnection::update_syndicated running too early in a Gutenberg/REST context
**Describe the bug**
When updating a post that has been distributed to another site in a multisite (internal networkconnection) using Gutenberg, updates to meta, terms, etc are not pushed. The issue here seems to be that [the `update_syndicated` method in the `NetworkSiteConnection` class is hooked to the `save_post` action.](https://github.com/10up/distributor/blob/b1dd60f9fdb6e22efd84a85807a6b7aec91e615e/includes/classes/InternalConnections/NetworkSiteConnection.php#L477) This works fine when using the classic editor, but when using Gutenberg and posts are saved through the REST-api, the `save_post` hook seems to fire earlier. Looking at the [`update_item` method of the `WP_REST_Posts_Controller`](https://github.com/WordPress/WordPress/blob/10772e8499d0695e3d9c57e4a8a24efe492afe41/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L684) class one can see that `wp_update_post`that triggers the `save_post` action [is called on line 697](https://github.com/WordPress/WordPress/blob/10772e8499d0695e3d9c57e4a8a24efe492afe41/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L697), and only after that are meta, terms etc handled by the REST-api. This means that when Distributor is running the `update_item` method, meta, terms etc haven't been updated/saved yet.

As a work around I've added the following code to an mu-plugin:
```
add_action(
	'rest_after_insert_post',
	function( WP_Post $post ): void {
		\Distributor\InternalConnections\NetworkSiteConnection::update_syndicated( $post->ID );
	}
);
```
This is not optimal since it means that the post will be distributed twice on save, it's no biggie either, but it makes saving a bit slower since `update_syndicated` is run twice.

I'm not sure what the best solution to this problem is to make it work with both Gutenberg and the classic editor. One could of course hook the `update_syndicated` method (with some minor modifications to account for the different values sent through the `save_post` and `rest_after_insert_{$post_type}` actions) to the `rest_after_insert_{$post_type}` action for all post_types where show_in_rest = true. This would mean that the `update_syndicated` method would be run twice, but that could maybe be handled by aborting that method if something like `( ( defined( 'REST_REQUEST' ) && REST_REQUEST ) && doing_action( 'save_post' ) )` was true. But I feel that there are probably smarter ways than that. ;)

Anyways, thanks for a great plugin and I'd be happy to help out with this issue if y'all are interested.

**Steps to Reproduce**

1. On a WordPress multisite create a post using Gutenberg with at least a title, category and a featured image. (Important, do not use any legacy metaboxes in Gutenberg because that triggers another save in WP to handle back-compat, which makes this issue go away)
2. Distribute this post to another site in the network and check that it looks good.
3. Change the title, category and featured image and update the post.
4. Check the distributed post on the other site and see that the title has changed but not the category or the featured image.
5. Update only the title of the post.
6. Check the distributed post on the other site and see that the title has changed and the category and featured image saved in step 3 has been distributed (since they where already saved when step 5 was run).

**Expected behavior**

Meta, terms etc should be distributed on update.

**Environment information**
 - Distributor version: Latest from develop branch.
 - Theme and version: Twenty Nineteen 1.4
 - Other installed plugin(s) and version(s): none.
 - WordPress 5.2.3

lakrisgubben:
### Description of the Change

Add a check to the `NetworkSiteConnection::update_syndicated` method to see if the current post is saved through Gutenberg. If so add a new action that runs the same method but on the `rest_after_insert_{$post_type}` hook to fix an issue where updates to terms/meta etc where not beeing syndicated.

For a deeper discussion about what problems this is trying to solve see issue #464

### Alternate Designs

Had some back-and-forth on the #464 issue, but this solution seemed like the most "elegant" to a problem without any ideal solutions.

### Benefits

Terms/meta etc. will be syndicated when updating an already syndicated post, no matter if it's written using the classic editor, Gutenberg without legacy metaboxes or Gutenberg with legacy metaboxes.

### Possible Drawbacks

Maybe if WP changed how the legacy metaboxes are saved this would need to be reworked, but I'm having a hard time seeing them change that functionality.

### Verification Process

Tested this with the classic editor and Gutenberg with and without legacy metaboxes and terms/meta etc are all syndicated correctly when updating a post. See #464 for more details on the bug and how to trigger it.

### Checklist:

- [x ] I have read the [**CONTRIBUTING**](https://github.com/10up/distributor/blob/develop/CONTRIBUTING.md) document.
- [ x] My code follows the code style of this project.

### Applicable Issues

Fixes #464

### DIFF
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -573,14 +573,23 @@ class NetworkSiteConnection extends Connection {
 	/**
 	 * Update syndicated post when original changes
 	 *
-	 * @param  int $post_id Post ID.
+	 * @param  int|WP_Post $post Post ID or WP_Post
+	 * depending on which action the method is hooked to.
 	 */
-	public static function update_syndicated( $post_id ) {
+	public static function update_syndicated( $post ) {
+		$post    = get_post( $post );
+		$post_id = $post->ID;
 		if ( ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) || wp_is_post_revision( $post_id ) || ! current_user_can( 'edit_post', $post_id ) ) {
 			return;
 		}
 
 		if ( 'trash' === get_post_status( $post_id ) ) {
+			return;
+		}
+
+		// If using Gutenberg, short circuit early and run this method later to make sure terms and meta are saved before syndicating.
+		if ( \Distributor\Utils\is_using_gutenberg( $post ) && doing_action( 'save_post' ) && ! isset( $_GET['meta-box-loader'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+			add_action( "rest_after_insert_{$post->post_type}", array( '\Distributor\InternalConnections\NetworkSiteConnection', 'update_syndicated' ) );
 			return;
 		}
 


lakrisgubben:
Regarding the PHPCS warning on line 591 in includes/classes/InternalConnections/NetworkSiteConnection.php: "Processing form data without nonce verification." Since we're not actually processing any form data – just checking if a $_GET variable isset – I'm inclined to add a phpcs:ignore statement to that line. But happy to do something else if y'all think that's better. :)

dkotter (dismissed):

dkotter:
@lakrisgubben Thanks for this, looks great and in my testing, things update correctly now when using Gutenberg. If you want to add that ignore statement so the tests pass, I think we'll be good here.

Thanks again!

lakrisgubben:
thanks @dkotter! I've added the phpcs:ignore statement now.

dkotter (approved):

PR #528, opened 2020-02-19
Docs: Unable to PUSH new CPTs

### LINKED ISSUE #366: Unable to PUSH new CPTs
In order for CPTs to be supported, we state in our [Known Caveats/Issues section](https://github.com/10up/distributor#known-caveatsissues) that:

> In addition, in order for distribution to work with external connections that have custom post type content, that post type needs to be registered with the argument `show_in_rest => true` on the external site.

I've been triaging a couple user setups who are leveraging the [CPT UI plugin](https://wordpress.org/plugins/custom-post-type-ui/) for creating CPTs.  In my testing I'm seeing that with proper External Connections created and the same CPT setup on each side of the connection, that you can PULL new CPTs without issue.  However, when you try to PUSH a new CPT the Distributor menu only shows `No connections available for distribution.` instead of the existing external connection site as expected.

The CPT UI plugin does default to `show_in_menu: true` when it registers its CPTs and I've confirmed those settings via its `Post Types and Taxonomies registered by Custom Post Type UI` page (e.g., http://sampledomain.com/wp-admin/admin.php?page=cptui_listings).  I don't think we require anything added to the custom "supports" value, but that's something we might investigate as possibly needed (at least in relation to the CPT UI plugin, [see more on that here](https://docs.pluginize.com/article/28-third-party-support-upon-registration)).

I was eventually able to get PUSH working for these external connections, but only after going into each sites' `Edit External Connection` pages and clicking `Update Connection`.  So I'm not sure if there's a race condition happening, if Distributor needs to more regularly check for newly registered CPTs, have something added to the CPT UI custom "supports" value, or something I've yet to imagine.

dinhtungdu:
### Description of the Change
Add new Know issues section about unable to push new CPTs.

### Alternate Designs
See https://github.com/10up/distributor/issues/366#issuecomment-584409063, https://github.com/10up/distributor/issues/366#issuecomment-586355383 and https://github.com/10up/distributor/issues/366#issuecomment-587457294

### Benefits
Cover one more issue.

### Possible Drawbacks
n/a

### Verification Process
Manual verification.

### Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/10up/distributor/blob/develop/CONTRIBUTING.md) document.

### Applicable Issues
Closes #366

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,8 @@ __Parent Posts__ - Distributor does not "bring along" parent (or child posts). I
 
 __Custom Post Type Support__ - Internal Connections (multisite) support multiple post types. In order for distribution to work with External Connections that have custom post type content, that post type needs to be registered with the argument `show_in_rest => true` on the external site.
 
+__Unable to push to new custom post types__ - If new post types are created after establishing external connections, only `Pull` work for those newly created post types. To make `Push` work, update the affected external connections by editing the connections then clicking the `Update connection` button.
+
 __Backwards Compatibility__ - While we strive to be mindful of backwards compatibility much the same way WordPress itself is, we do not currently guarantee continued interoperability between different versions of Distributor. We assume the current userbase for this plugin has a high degree of control over any site that has been set up as an external connection and urge you to keep Distributor up to date.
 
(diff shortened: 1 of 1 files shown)


jeffpaul (approved):
Made a couple tweaks to how we describe what's needed, but this is great... thanks for getting this covered @dinhtungdu!

PR #529, opened 2020-02-21
Add filter for request timeout

### LINKED ISSUE #523: Add filter for request timeout
In case the posts have large meta, updating the post meta may take more than 5 seconds. We have **\Distributor\Subscriptions\send_notifications** function working in background, and sending big amount of posts to be distributed (4000 in our case) when others aren't yet completed, causes to have multiple processes working in parallel with the endpoint. Some of the processing scripts get timeout, and we have incomplete post metas distributed as result. We need to wait for the endpoint response, and only after that send the next post to be distributed.

Suggestion is to add a filter for the timeout, so we can hook in and modify it as desired.
https://github.com/10up/distributor/blob/develop/includes/subscriptions.php#L283

I can see a similar filter in **\Distributor\ExternalConnections\push()**
https://github.com/10up/distributor/blob/develop/includes/classes/ExternalConnections/WordPressExternalConnection.php#L616

aahovhannisyan:
Add filter for request timeout to modify it according to needs.
Closes #523 .

### DIFF
--- a/includes/subscriptions.php
+++ b/includes/subscriptions.php
@@ -285,9 +285,10 @@ function send_notifications( $post_id ) {
 				 *
 				 * @hook dt_subscription_post_timeout
 				 *
-				 * @param {int} $timeout The timeout to use for the remote post. Default `5`.
+				 * @param int $timeout The timeout to use for the remote post. Default `5`.
+				 * @param \WP_Post $post The post object
 				 *
-				 * @return {int} The timeout to use for the remote post.
+				 * @return int The timeout to use for the remote post.
 				 */
 				'timeout' => apply_filters( 'dt_subscription_post_timeout', 5, $post ),
 				/**
--- a/includes/subscriptions.php
+++ b/includes/subscriptions.php
@@ -280,7 +280,16 @@ function send_notifications( $post_id ) {
 		$request = wp_remote_post(
 			untrailingslashit( $target_url ) . '/wp/v2/dt_subscription/receive',
 			[
-				'timeout' => 5,
+				/**
+				 * Filter the timeout used when calling `\Distributor\Subscriptions\send_notifications`
+				 *
+				 * @hook dt_subscription_post_timeout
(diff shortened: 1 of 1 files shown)


includes/subscriptions.php
```diff
@@ -280,7 +280,16 @@ function send_notifications( $post_id ) {
...
+				/**
+				 * Filter the timeout used when calling `\Distributor\Subscriptions\send_notifications`
+				 *
+				 * @hook dt_subscription_post_timeout
+				 *
+				 * @param {int} $timeout The timeout to use for the remote post. Default `5`.
+				 *
+				 * @return {int} The timeout to use for the remote post.
+				 */
+				'timeout' => apply_filters( 'dt_subscription_post_timeout', 5 ),
```
dinhtungdu: What about adding `$post` as an argument of this filter, this may help some edge cases (I haven't think about) and make this filter align with `dt_push_post_timeout` filter.

aahovhannisyan: @dinhtungdu thanks for review. Added.

includes/subscriptions.php
```diff
@@ -280,7 +280,16 @@ function send_notifications( $post_id ) {
...
 			[
-				'timeout' => 5,
+				/**
+				 * Filter the timeout used when calling `\Distributor\Subscriptions\send_notifications`
+				 *
+				 * @hook dt_subscription_post_timeout
+				 *
+				 * @param {int} $timeout The timeout to use for the remote post. Default `5`.
+				 *
+				 * @return {int} The timeout to use for the remote post.
```
dinhtungdu: IMO, we should follow phpDoc standard here by removing `{}` from the types

aahovhannisyan: @dinhtungdu removed braces and added tag for $post parameter.

dinhtungdu (approved):
@armdev17 thanks for your work! LGTM 👍

PR #538, opened 2020-03-09
Fix: Push UI should not disappear when push is in progress

### LINKED ISSUE #275: Push UI should not disappear when push is in progress
Because the admin bar submenus appear on hover, the dropdown for the push UI disappears if you move your mouse, making it very easy to miss out on whether your push has succeeded and allowing you to navigate away when you shouldn't. I think the right solution to this is to freeze the submenu open while push is happening and blur out and disable the rest of what's visible in the window around it.

dinhtungdu:
### Description of the Change

- Add a background hack to prevent accident click when push is in progress.
- Keep UI visible when push is in progress.
- Fix watch task to make it work with gulp 4.

### Alternate Designs
n/a

### Benefits
- Better UX for pushing posts.

### Possible Drawbacks
n/a

### Verification Process

1. Set up and configure distributor, add at least 1 connection.
2. On the edit page, hover on the Distributor link of the admin bar.
3. See: the other parts of the page covered with a black overlay.
4. Choose a connection, click the `Distribute` button and quickly move the cursor outside the push UI.
5. See: the UI is still visible.
6. See: other parts of the edit page can't be clicked.

### Checklist:

### Applicable Issues

Closes #275

### DIFF
--- a/assets/js/push.js
+++ b/assets/js/push.js
@@ -191,7 +191,7 @@ jQuery( window ).on( 'load', () => {
 	 */
 	function distributorMenuEntered() {
 		distributorMenuItem.focus();
-		document.body.classList.toggle( 'distributor-show' );
+		document.body.classList.add( 'distributor-show' );
 
 		if ( distributorPushWrapper.classList.contains( 'loaded' ) ) {
 			return;
@@ -238,7 +238,7 @@ jQuery( window ).on( 'load', () => {
 	 */
 	function distributorMenuExited() {
 
-		if ( actionWrapper.classList.contains( 'loading' ) ) {
+		if ( actionWrapper && actionWrapper.classList.contains( 'loading' ) ) {
 			return;
 		}
 
--- a/assets/js/push.js
+++ b/assets/js/push.js
@@ -237,11 +237,16 @@ jQuery( window ).on( 'load', () => {
 	 * Handle exiting the distributor menu.
 	 */
 	function distributorMenuExited() {
+
+		if ( actionWrapper.classList.contains( 'loading' ) ) {
+			return;
+		}
+
 		distributorMenuItem.blur();
-		document.body.classList.toggle( 'distributor-show' );
+		document.body.classList.remove( 'distributor-show' );
 	}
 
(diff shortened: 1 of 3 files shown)


assets/js/push.js
```diff
@@ -237,11 +237,16 @@ jQuery( window ).on( 'load', () => {
 	 * Handle exiting the distributor menu.
 	 */
 	function distributorMenuExited() {
+
+		if ( actionWrapper.classList.contains( 'loading' ) ) {
```
dkotter: `actionWrapper` may not be set up yet here, which leads to an error:
```
Uncaught TypeError: Cannot read property 'contains' of undefined
```

dkotter (changes requested):

dkotter:
While testing this, found one minor bug I left a comment about.

Overall I think this is better then what we have but I ran into on issue.

If I click to distribute something and then immediately move my cursor out of the dropdown, it stays open during and after the distribution, which is nice. But if I then move my cursor back to the overlay, it immediately closes, which is a little jarring.

dinhtungdu:
Thank you @dkotter, I addressed all issues you mentioned, please take another look to this : )

assets/js/push.js
```diff
@@ -237,11 +237,16 @@ jQuery( window ).on( 'load', () => {
 	 * Handle exiting the distributor menu.
 	 */
 	function distributorMenuExited() {
+
+		if ( actionWrapper && actionWrapper.classList.contains( 'loading' ) ) {
```
dkotter: @dinhtungdu This does get rid of the JS error, which is great. But in my testing, if hover over the Distributor menu and then quickly hover out, while connections are still be loaded in, the menu closes (because the `actionWrapper` doesn't exist yet, so this check is bypassed).

I think we may want to change this to:
```
if ( ! actionWrapper || actionWrapper.classList.contains( 'loading' ) )
```

dinhtungdu: I made it that way because if I hover over the menu then quickly hover out, I consider it as accident behavior. But after testing again, closing the menu after quickly hovering out looks buggy. Thanks for the head up!
I added a check to cover the case when we have connections loading issue as well.

dkotter (approved):

PR #540, opened 2020-03-10
Fix: Uninstall distributor

### LINKED ISSUE #349: Uninstall distributor (completely)
Hi there,

How to uninstall/delete distributor completely? It leaves db values behind.

After deleteing and reinstalling, ... settings are still the same.

dinhtungdu:
### Description of the Change
- Add an `uninstall.php` script to delete distributor data from the database.
- Add `DT_REMOVE_ALL_DATA ` constant to prevent accidental data deletion.
- Deletes entries from the options table.
- Do not delete posts and meta other than Subscriptions.

A modal will display when deactivating the plugin. This modal does not display in sub-sites.

image

### Benefits
Delete distributor data on uninstalling.

### Possible Drawbacks
n/a

### Verification Process
Check the database manually.

### Checklist:

### Applicable Issues
Closes #349 

### Changelog Entry

> Added - Provide way to delete all Distributor data on plugin deletion, utilizing the `DT_REMOVE_ALL_DATA` constant.

### Credits

Props @dinhtungdu, @faisal-alvi, @peterwilsoncc, @ravinderk, @qasumitbagthariya, @jeffpaul, @dkotter


jeffpaul:
Moving this to `2.1.0` as we still need to discuss what items we can/should safely remove during uninstall.

ravinderk:
@peterwilsoncc, Can you provide a list of data we should remove when someone installs a plugin?

cc: @dkotter @jeffpaul

jeffpaul:
@peterwilsoncc could be a nice addition to v2, but will let you decide whether to include or punt to v2.0.1/v2.1.0

uninstall.php
```diff
@@ -0,0 +1,30 @@
...
+
+// Delete options.
+$wpdb->query( "DELETE FROM $wpdb->options WHERE option_name LIKE 'dt\_%';" );
+
+// Remove transients.
+$wpdb->query( "DELETE FROM $wpdb->options WHERE option_name LIKE '_transients\_dt\_%';" );
+
+// Delete our data from the post and post meta tables.
+$wpdb->query( "DELETE FROM $wpdb->posts WHERE post_type IN ( 'dt_subscription', 'dt_ext_connection' );" );
+$wpdb->query( "DELETE meta FROM $wpdb->postmeta meta LEFT JOIN $wpdb->posts posts ON posts.ID = meta.post_id WHERE posts.ID IS NULL;" );
```
peterwilsoncc: This will leave term relationships in place (an extender may add these) and skip a lot of hooks other developers may be using. 

To speed up `wp_delete_post`, we can use `wp_suspend_cache_invalidation()`  and `wp_cache_delete_multiple()` for the cache invalidation (this will require a WP version check).

That will ensure the term relationships are removed and term counts counts are correct

uninstall.php
```diff
@@ -0,0 +1,30 @@
...
+
+// Clear cache.
+wp_cache_flush();
```
peterwilsoncc: I know that some hosts block this outside of tests and WPCLI.

In combination with the suggestion above, we can call `wp_cache_set_posts_last_changed()`

uninstall.php
```diff
@@ -0,0 +1,30 @@
...
+	die;
+}
+
+global $wpdb;
+
+// Delete options.
+$wpdb->query( "DELETE FROM $wpdb->options WHERE option_name LIKE 'dt\_%';" );
+
+// Remove transients.
+$wpdb->query( "DELETE FROM $wpdb->options WHERE option_name LIKE '_transients\_dt\_%';" );
```
peterwilsoncc: * this needs to account for the transient without a key https://github.com/10up/distributor/blob/4ceb996dc23e6cdec514bb53544998ea922d8e8d/includes/classes/InternalConnections/NetworkSiteConnection.php#L864-L865
* we might need to flush the option's and all_options caches for this. Transients are only stored in the options table for non-persistent caches, so it the page immediately refreshes after this then we might be able to get away with it

peterwilsoncc (commented):
I've added some notes inline.

I'm in two minds about deleteing all the data in case the plugin was deactivated accidently but I don't think there is much we can do with that. Some of our options get quite large

uninstall.php
ravinderk: @peterwilsoncc I prefixed it. Let me know if I need to include any other transient key.

uninstall.php
ravinderk: Direct SQL queries are faster than WordPress core function. I am using them because we are sure about the size of the database.

ravinderk:
@peterwilsoncc At this moment we are deleting distributor subscriptions ( `'dt_subscription', 'dt_ext_connection'`). We should delete pulled posts. Can you share your thought?

If we decide not to delete, we must leave comments, terms, and term taxonomy relationships data.

We should only delete connections and options data and leave posts data but what about posts metadata about the distributor?

cc: @jeffpaul @dkotter

jeffpaul:
I think any pushed/pulled posts should stay as-is as its feasible that someone uses Distributor to migrate content and then deletes Distributor once that's done, so we should NOT be changing any post content or meta accordingly (besides the subscription meta).

ravinderk:
@peterwilsoncc  Distributor supports multisite. In this case, the Super Admin can only delete the plugin. I think we should clear distributor plugin data from subsites, what does this think?

cc: @jeffpaul

jeffpaul:
@ravinderk so you're asking about the case where distributor is network installed and network activated (different from the case where it's installed or activated on individual sites for a use case where someone only wants distributor to work on parts of a multi site)?

ravinderk:
> @ravinderk so you're asking about the case where distributor is network installed and network activated (different from the case where it's installed or activated on individual sites for a use case where someone only wants distributor to work on parts of a multi site)?

@jeffpaul You are correct.

jeffpaul:
I'm not quite wildly in favor of deleting Distributor syndication mapping post meta when someone deactivates the plugin, though I suppose if someone actually uninstalls (again, specifically considering this differently from deactivating) it's probably fine to clear Distributor data.  The downside is someone who then wants to re-install and activate Distributor that all previous syndication mappings would be gone and make things super difficult to get synchronized again, but that would fall into a `_doing_it_wrong()` scenario.

@ravinderk want to work on iterating on the PR here to properly remove Distributor data when someone uninstalls the plugin (again, NOT when its merely deactivated but ONLY when it's uninstalled)?

ravinderk:
> I'm not quite wildly in favor of deleting Distributor syndication mapping post meta when someone deactivates the plugin, though I suppose if someone actually uninstalls (again, specifically considering this differently from deactivating), it's probably fine to clear Distributor data. The downside is that for someone who then wants to re-install and activate Distributor, all previous syndication mappings would be gone and make things super difficult to get synchronized again, but that would fall into a `_doing_it_wrong()` scenario.
> 
> @ravinderk wants to work on iterating on the PR here to properly remove Distributor data when someone uninstalls the plugin (again, NOT when it's merely deactivated but ONLY when it's uninstalled)?

@jeffpaul I implemented logic to remove distributor data from the database only if:
- Admin uninstalling plugin
- `DT_REMOVE_ALL_DATA` is set to `true` (to prevent accidental deletion of data) 

I also recommend showing the modal when uninstalling the distributor plugin to inform the admin.

> @ravinderk, so you're asking about the case where the distributor is network installed and network activated (different from the case where it's installed or activated on individual sites for a use case where someone only wants the distributor to work on parts of a multi-site)?

Removing distributor data from each subsite can be very resource-intensive. I think we should not do it and inform the admin about this when uninstalling the plugin

jeffpaul:
>I also recommend showing the modal when uninstalling the distributor plugin to inform the admin.

> Removing distributor data from each subsite can be very resource-intensive. I think we should not do it and inform the admin about this when uninstalling the plugin.

I concur on both points.

ravinderk:
@jeffpaul I implemented the deactivation modal.

cc: @peterwilsoncc

uninstall.php
```diff
@@ -0,0 +1,39 @@
...
+if ( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
+	die;
+}
+
+/*
+ * Only remove ALL product and page data if DT_REMOVE_ALL_DATA constant is set to true in user's
+ * wp-config.php. This is to prevent data loss when deleting the plugin from the backend
+ * and to ensure only the site owner can perform this action.
+ */
+if ( defined( 'DT_REMOVE_ALL_DATA' ) && true === DT_REMOVE_ALL_DATA ) {
```
peterwilsoncc, line 20: This will only remove the data from the current site (the network base site on the network tab) but the data will remain on sub-sites.

It might be possible to loop through the sites on small networks.

uninstall.php
```diff
@@ -0,0 +1,39 @@
...
+	$wpdb->query( "DELETE FROM $wpdb->options WHERE option_name LIKE '_transients\_dt\_%';" );
+
+	// Delete our data from the post and post meta tables.
+	$wpdb->query( "DELETE FROM $wpdb->posts WHERE post_type IN ( 'dt_subscription', 'dt_ext_connection' );" );
+	$wpdb->query( "DELETE meta FROM $wpdb->postmeta meta LEFT JOIN $wpdb->posts posts ON posts.ID = meta.post_id WHERE posts.ID IS NULL OR meta.meta_key LIKE 'dt\_%';" );
+
+	// Clear cache.
+	if ( wp_cache_flush() ) {
+		wp_cache_set_posts_last_changed();
+	}
```
peterwilsoncc:
```suggestion
	wp_cache_set_posts_last_changed();
	wp_cache_delete( 'alloptions', 'options' );
	
	// The cache for individual posts will need to be removed (it doesn't use last changed)
	// The cache for individual options will need to be removed
	// On sites will a persistent cache, the transients will need to be removed too
```

includes/bootstrap.php
```diff
@@ -216,6 +217,77 @@ function() {
...
+		$modal_title                   = esc_html__( 'Distributor Deactivation', 'distributor' );
+		$modal_button_title_deactivate = esc_html__( 'Deactivate', 'distributor' );
+		$modal_button_title_cancel     = esc_html__( 'Cancel', 'distributor' );
+		$script                        = <<<EOD
+			jQuery(document).ready(function($) {
+				const deactivateButton = jQuery('#deactivate-distributor');
+				deactivateButton.on( 'click', function() {
+					$("#my-modal").dialog({
+						modal: true,
+						title: "$modal_title",
```
peterwilsoncc: This and the other dynamic options will need to be run through `wp_json_encode()` to account for items note encoded by esc_html.

includes/bootstrap.php
```diff
@@ -216,6 +217,77 @@ function() {
 	1
 );
 
+
+/**
+ * Add a deactivation modal when deactivating the plugin.
+ *
+ * @since x.x.x
+ */
+add_action(
```
peterwilsoncc: This will need to use a named function rather than a closure. That will allow developers to unhook it from the site if they wish.

peterwilsoncc (commented):
I've added a few notes inline

uninstall.php
ravinderk: @peterwilsoncc, after discussing with @jeffpaul, we decided to take care of the main website. It will be helpful for users to be aware of this. 

@jeffpaul Do you have thoughts?

includes/bootstrap.php
ravinderk: @peterwilsoncc I used `esc_js`. Is it fine?

includes/bootstrap.php
ravinderk: Fixed: https://github.com/10up/distributor/pull/540/commits/38c8d1565f44c3e909aafcba712d1172eabb8c3c

includes/bootstrap.php
ravinderk: fixed: https://github.com/10up/distributor/pull/540/commits/d10a2f733efbfaa928b4289e59b51dcafcbd36cc

uninstall.php
jeffpaul: I think that so long as we've presented the admin uninstalling Distributor with an "are you sure" sort of message on the impacts of fully removing Distributor data, then we should look for a reliable and performant, if slow and long-running, approach to clearing that data across all sites on a network.

jeffpaul:
Let's follow what WP.org recommends/requires: https://developer.wordpress.org/plugins/plugin-basics/uninstall-methods/

jeffpaul:
@10up/open-source-practice another PR that would be good to look into getting wrapped up and ready for release

faisal-alvi:
Update:

- Looping through all subsites in the network and deleting data.
- All plugin options will be deleted.
- Only the posts and post meta of the Subscription will be deleted; other internal/external posts will be kept.
- Improved DB query; only deletes the subs, first deletes the meta, and then posts.
- Improved popup message; mentions what's being deleted, recommends taking backup as the action would be undone

image

_Note: Ignoring E2E failures because they are being fixed in https://github.com/10up/distributor/pull/1248_

Sidsector9:
@peterwilsoncc some changes have been pushed after your review. Can you please verify your comments resolved by Ravinder? The feature has been tested by Faisal, just requires a confirmation to merge this PR. Thanks!

peterwilsoncc (commented):
The data deletion function will leave a bunch of caches in place for posts and post meta that doesn't exist, so we'll need to get the post IDs to be deleted, delete them and flush the caches. This is because the individual posts & post meta don't use the last changed time.

As distributor makes use of site options and site transients, we'll also need to delete the options differently depending on the type of install. Since WordPress 6.6 the autoloading of options depends on the options size, so we'll need to check for additional values to determine if the option is cached in all options or as an individual option.

As my suggested changes are quite a lot, I've put together a gist with the items split out in to multiple functions to allow for the cache to be deleted.

https://gist.github.com/peterwilsoncc/286025a3ca49b90c4d4f2e06741396a2

> [!Warning]
> Please review my code carefully and check the queries generated. 
>
> Much of the code was written by Copilot so could do with some tidy-up, for one. Secondly, it uses some WPDB methods that are quite rare so it would be good to have a logic check

If you think deleting the options via the option ID will be faster (it's the primary key, then feel free to update the code to do so.

includes/bootstrap.php
```diff
@@ -216,6 +217,75 @@ function() {
...
+				esc_html__( 'By default, the database entries are not deleted when you deactivate Distributor. If you are deleting Distributor completely from your website and want those items removed as well, add the code below to wp-config.php:', 'distributor' ),
+				'define( \'DT_REMOVE_ALL_DATA\', true );',
+				esc_html__( 'After adding this code, the Distributor plugin data will be removed from the website database when deleting the plugin. This will not delete the posts with their metadata other than the subscription. You can review uninstall.php (in the plugin root directory) to learn more about the deleted data. After deleting the Distributor plugin, you can remove the code from the wp-config.php file. Please make sure that this action cannot be undone; take a backup before proceeding.', 'distributor' )
+			);
+		}
+	);
+
+	$modal_title                   = esc_js( esc_html__( 'Distributor Deactivation', 'distributor' ) );
+	$modal_button_title_deactivate = esc_js( esc_html__( 'Deactivate', 'distributor' ) );
+	$modal_button_title_cancel     = esc_js( esc_html__( 'Cancel', 'distributor' ) );
```
peterwilsoncc: `esc_js` is for inline attributes such as `onclick` and can cause errors for use within script tags.

```suggestion
	$modal_title                   = wp_json_encode( __( 'Distributor Deactivation', 'distributor' ) );
	$modal_button_title_deactivate = wp_json_encode( __( 'Deactivate', 'distributor' ) );
	$modal_button_title_cancel     = wp_json_encode( __( 'Cancel', 'distributor' ) );
```

includes/bootstrap.php
```diff
@@ -216,6 +217,75 @@ function() {
...
+	$script                        = <<<EOD
+			jQuery(document).ready(function($) {
+				const deactivateButton = jQuery('#deactivate-distributor');
+				deactivateButton.on( 'click', function() {
+					$("#my-modal").dialog({
+						modal: true,
+						title: "$modal_title",
```
peterwilsoncc:
```suggestion
						title: $modal_title,
```

`wp_json_encode()` will add the quotes.

includes/bootstrap.php
```diff
@@ -216,6 +217,75 @@ function() {
...
+						width: 550,
+						buttons: [
+							{
+								text: "$modal_button_title_cancel",
```
peterwilsoncc:
```suggestion
								text: $modal_button_title_cancel,
```

As above

includes/bootstrap.php
```diff
@@ -216,6 +217,75 @@ function() {
...
+								class: "button-secondary",
+								click: function() {
+									$(this).dialog("close");
+								}
+							},
+							{
+								text:"$modal_button_title_deactivate",
```
peterwilsoncc:
```suggestion
								text:$modal_button_title_deactivate,
```

As above.

uninstall.php
```diff
@@ -0,0 +1,173 @@
...
+		);
+
+		if ( ! empty( $subscription_post_ids ) ) {
+			$ids_string = implode( ',', array_map( 'intval', $subscription_post_ids ) );
+
+			// Delete subscription meta.
+			$wpdb->query(
+				$wpdb->prepare(
+					"DELETE FROM $wpdb->postmeta WHERE post_id IN ($ids_string)"
+				)
```
peterwilsoncc:
```suggestion
				"DELETE FROM $wpdb->postmeta WHERE post_id IN ($ids_string)"
```

No need to use `prepare` without placeholders. :1234: Applies elsewhere in this PR.

That said, I think I'd prefer to use the `implode...count` method to generate the placeholders and use a prepare statement just to be safe. See the code in `wp_prime_options_cache()` https://github.com/WordPress/WordPress/blob/4d0a623d7a01ca03aca529bfb35d7bca7c35ba48/wp-includes/option.php#L261

uninstall.php
```diff
@@ -0,0 +1,173 @@
...
+	 */
+	function delete_site_options() {
+		global $wpdb;
+
+		$option_prefixes = array(
+			'dt_',
+			'_transient_dt_',
+			'_transient_timeout_dt_',
+		);
+
```
peterwilsoncc, line 74: This will need to include the site_transient prefixes for `! multisite()` -- they're stored in the options table but with still use the site prefix.

uninstall.php
```diff
@@ -0,0 +1,173 @@
...
+			// Delete the options using the retrieved IDs.
+			$wpdb->query(
+				$wpdb->prepare(
+					"DELETE FROM $wpdb->options WHERE option_id IN ($ids_string)"
+				)
+			);
+
+			// Flush the options cache.
+			wp_cache_delete_multiple( $options_to_delete, 'options' );
```
peterwilsoncc: This is being called with the option ID but the cache uses the option_name

uninstall.php
```diff
@@ -0,0 +1,173 @@
...
+
+			// Delete the sitemeta using the retrieved IDs.
+			$wpdb->query(
+				$wpdb->prepare(
+					"DELETE FROM $wpdb->sitemeta WHERE meta_id IN ($ids_string)"
+				)
+			);
+
+			// Flush the site options cache.
+			wp_cache_delete_multiple( $sitemeta_to_delete, 'site-options' );
```
peterwilsoncc: This is using the option ID for the cache names, but it's cached using the pattern `$network_id:$option_name`

uninstall.php
```diff
@@ -0,0 +1,173 @@
+<?php
+/**
+ * Distributor uninstall script.
+ *
+ * @package distributor
+ * @since   x.x.x
+ */
+
+// phpcs:disable
```
peterwilsoncc:
```suggestion
```

Rather than disabling all styles, just disable them on a line by line basis with a comment explaining why. It will help prevent future devs (probably ourselves) from introducing errors.

uninstall.php
```diff
@@ -0,0 +1,173 @@
...
+
+		// Delete network-wide sitemeta options.
+		delete_sitemeta_options();
+	} else {
+		// Single site.
+		dt_delete_data();
+	}
+}
+
+// phpcs:enable
```
peterwilsoncc:
```suggestion
```

As above.

peterwilsoncc:
@jeffpaul This still needs a little more from the previous review: fixing the options cache IDs, site options and adding the line phpcs notes for each line.

faisal-alvi:
Correct @peterwilsoncc This was dropped to pick some other work. I will drop a comment once this is ready for review. Thanks!

uninstall.php
faisal-alvi: I was under the impression that `_site_transient_dt_` and `_site_transient_timeout_dt_` are used exclusively in multisite networks. That’s why I’m not quite sure why we should include them in this context, and that also with `! multisite()` condition!? Shouldn't we use a `multisite()` condition? Side note: we are already removing them from the `sitemeta` table.

uninstall.php
faisal-alvi: This is using `meta_id` from the `sitemeta` table, do you mean we should pass `$network_id:{sitemeta.meta_key}`? Where we can use `sitemeta.site_id` as a `$network_id`. So something like this?:

`wp_cache_delete_multiple( [Array of {sitemeta.site_id:sitemeta.meta_key}], 'site-options' );`

faisal-alvi:
@peterwilsoncc I've added a few commits and included a couple of questions for confirmation. Could you please take a look at https://github.com/10up/distributor/pull/540#discussion_r1758861046 and https://github.com/10up/distributor/pull/540#discussion_r1758976777?

uninstall.php
peterwilsoncc: Site transients can be used on single sites too but are stored in the options table rather than the site meta table with the `_site_transient` prefix. They use the `*_site_option` functions which eventually flow through to the `*_option` functions on single site installs.

faisal-alvi:
@peterwilsoncc I’m now retrieving both the option IDs and names. IDs to delete the records, and names to clear the caches.

For multisite, I’m using `$network_id: sitemeta.meta_key` for cache deletion. Let me know if any adjustments are needed.

uninstall.php
peterwilsoncc, line 1: File needs to be added to this block of code

https://github.com/10up/distributor/blob/c27d99a923785f32e9d5c4560363144f83da39d6/webpack.config.release.js#L11-L29

peterwilsoncc (commented):
This is looking good but some odd behaviour when I try to delete the plugin

* the dist/css file can't be deleted.
* The database queries on a single site were failing. Not sure why.

To test I've been running the release script and unzipping and activating a duplicate copy to make sure I don't delete the git repo.

faisal-alvi:
Hey @peterwilsoncc,

I tested again on both a multisite network and a single site, but I didn’t encounter any odd behavior when deleting the plugin.

- The `dist/css` files were successfully deleted on my end (see the below GIF).
- The database queries were functioning properly, and all entries in the options table were being deleted.

all-files-deleted-distributor

I also ran the release script and used the zip to test.

Can you please check again? Or @qasumitbagthariya if you have some time to test this, that would be great.

peterwilsoncc (dismissed):
@faisal-alvi Looks like it was file permission error between my local and virtual environments. Once I ran chmod it worked as expected.

jeffpaul:
@qasumitbagthariya over to you for QA

qasumitbagthariya (dismissed):

qasumitbagthariya:
@faisal-alvi 

### QA Update  ✅

-------------

I have verified this PR in the `fix/349` branch, which has been fixed and is functioning as intended.

I tested the following on this branch:
 - Distribution de-activate popup ✅
 - Deletes entries from the options table. ✅
- Do not delete posts and meta other than Subscriptions. ✅


https://github.com/user-attachments/assets/6cb2ef4a-e561-40b1-b5e4-c087e68d1104



**Testing Environment** 

<Details>

- WordPress: 6.8.2
- Theme: Twenty Twenty-Five 1.3
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS 15.2 
- Branch: fix/349

</Details>


Steps to Test- As mentioned in the PR description.
Test Results - It is working as expected.
Functional Demo / Screencast - 
Special Notes -  Ready for UAT
Testing Document status:
Cases related to this Issue/PR are added to the Critical Flow Wiki pages: 

- [x] Yes
- [ ] Not Required/Applicable for this PR
------------------------------------------------

dkotter (approved):

PR #547, opened 2020-03-13
Fix: Delay `send_notifications` until after all meta has saved

### LINKED ISSUE #399: Delay `send_notifications` until after all meta has saved (Yoast SEO problem).
When updating an existing post, the Yoast SEO meta information sent to a linked post is not up-to-date. It appears Yoast does not save its metadata until after _send_notifications_ has already run, so Distributor's _distributor_meta_ array is incorrect.

If possible, have send_notifications run after all new meta has finished saving. I've tried simply adding a higher priority in the plugin code to the _save_post_ action, but wasn't able to get that working.

For now, I've created [a plugin](https://github.com/timstl/distributor-yoast-sync) that filters _dt_subscription_post_args_. This plugin goes through the _$_POST_ data and updates _$post_body['post_data']['distributor_meta']_ with the correct Yoast meta values. While this method works, it's specific to Yoast SEO and can easily break with changes to Yoast SEO.

**Additional context**
I have only produced this problem with external connections; I have not tried on network sites.

I marked this as an enhancement because it's related to another plugin. Other custom fields seem to pass correctly during an update, including the (non-media) ACF fields I have tested. I suspect there are other plugins where this is a problem, though.

The above plugin also fixes an issue with Yoast social images pointing to the incorrect media IDs and URLs, but that is partially a separate issue related to media IDs and custom fields that I believe is already known.

Thanks!

Tim

dinhtungdu:
### Description of the Change
Use `wp_insert_post` instead of `save_post` to update syndicated posts. `wp_insert_post` is the last action when updating a post, use it will ensure distributor update remote posts after all meta is saved.

### Verification Process
Check with Yoast SEO as describe in #399 

### Checklist:

### Applicable Issues
Closes #399

### DIFF
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -484,7 +484,7 @@ class NetworkSiteConnection extends Connection {
 	 */
 	public static function bootstrap() {
 		add_action( 'template_redirect', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'canonicalize_front_end' ) );
-		add_action( 'wp_insert_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'update_syndicated' ) );
+		add_action( 'wp_insert_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'update_syndicated' ), 100 );
 		add_action( 'before_delete_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'separate_syndicated_on_delete' ) );
 		add_action( 'before_delete_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'remove_distributor_post_from_original' ) );
 		add_action( 'wp_trash_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'separate_syndicated_on_delete' ) );
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -484,7 +484,7 @@ class NetworkSiteConnection extends Connection {
 	 */
 	public static function bootstrap() {
 		add_action( 'template_redirect', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'canonicalize_front_end' ) );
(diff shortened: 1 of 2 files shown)


includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -602,12 +602,6 @@ public static function update_syndicated( $post ) {
 			return;
 		}
 
-		// If using Gutenberg, short circuit early and run this method later to make sure terms and meta are saved before syndicating.
-		if ( \Distributor\Utils\is_using_gutenberg( $post ) && doing_action( 'save_post' ) && ! isset( $_GET['meta-box-loader'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
```
dkotter, line 606: @dinhtungdu Wondering the reasoning behind removing this code? I haven't actually tested these new changes, so maybe this is still working as expected, but the reason we have this here, is Gutenberg updates things via REST requests and both the `save_post` and `wp_insert_post` hooks fire before things like featured image, terms and meta are updated in that context.

This led to a bug where if you changed the featured image on a Gutenberg post, that change would not be syndicated, because the change isn't actually saved until after `save_post` runs (which I think will be the same for `wp_insert_post`). So we added this check here (which is not ideal but couldn't find a better approach), that checks if we're using Gutenberg, we're in the `save_post` action and we don't have any custom fields being saved, and then adds a hook on the `rest_after_insert_`, which fires only after the post is saved and all data is updated.

Again, didn't test this so ignore if I'm wrong, but I think we need to keep this and probably just update the `doing_action` check to look for `wp_insert_post`

dinhtungdu: I tested on WordPress 5.3.2 with block editor and the meta, term and featured image are synced successfully. But you're right, we may have race condition here so it's better to keep this code. Thanks so much for the head up!

jeffpaul:
Given that we're aiming to cover larger structural changes for Distributor in an upcoming release, it may make more sense to tandem this PR and related issue then.

peterwilsoncc:
I'm going to close this off as all of the files it touches are conflicted and with the new minimum requirement bumps it's possible to do this consistently for both the Gutenberg (REST) and classic updates.

Issue #563, opened 2020-03-18
Distributor menu in admin bar - better keyboard support


samikeijonen:
## Is your enhancement related to a problem? Please describe.
- Can’t select individual connection with keyboard. Or un-select.
- Focus styles missing inside the “Distributor menu”.
- I’d show connections as checkboxes in the first place. Now they are div or span.

## Describe the solution you'd like
Re-think the interface with accessible markup and keyboard support.

Related issue is #557

PR #564, opened 2020-03-18
Use WP native way of opening the Distributor menu

### LINKED ISSUE #557: Open distributor link in admin bar - can't use with keyboard
This probably needs to be chopped in smaller issues but at the moment pushing article feature using `distributor` link in admin bar is unusable with keyboard.

- Can’t open the Distributor menu with keyboard.
- Can’t select individual connection with keyboard. Or un-select.
- Focus styles missing inside the “Distributor menu”.
- I’d show connections as checkboxes in the first place. Now they are `div` or `span`.

Re-think the interface with accessible markup and keyboard support.

samikeijonen:
Fixes #557 

## Description of the Change

Remove `hoverIntent ` and add empty placeholder child item in Distributor menu. This way WP automatically handles also keyboard support.

## Benefits

- Also keyboard users can open the distributor menu from admin bar.
- Rely on WP method what comes to code. No need for own custom solution.

## Possible Drawbacks

Data use to be loaded on `hoverIntent `. Now that's removed, and data is loaded `jQuery( window ).on( 'load', ()`. I'm uncertain does that make any difference.

## Verification Process

- Use mouse to open Distributor menu in admin bar.
- Use keyboard to open Distributor menu in admin bar.

### Checklist:

### Applicable Issues

#557 and #563


samikeijonen:
**Note**: this actually broke distributing on my local in my test post. I'm not sure how I should test it though? Should I have same branch in both local installs?

dinhtungdu:
This doesn't break distributing in my test, however, I would expect to control the push panel by keyboard also, let's come back to this when #557 is resolved.

Also, this PR conflicts with #538

dkotter:
@samikeijonen Thanks for the fix here. Love to see these accessibility improvements.

This does reintroduce a performance issue that had previously been fixed. Basically, we only load connection information when the Distributor menu item is hovered over (instead of loading those on page load, which this PR adds that back).

For sites with lots of connections, this gives a dramatic bump in performance and offloads the loading of that data only when it's needed (when a user hovers over the menu and is looking to distribute content).

I would like to see this accessibility improvement get added but is there a way to do this that still allows us to only load connection information when the menu is opened?

dinhtungdu:
@dkotter Nice catch! I think we can solve that issue by adding event listeners for `keydown`, `touchstart`, and `mouseenter`. Currently, the admin bar doesn't have custom events that we can add listeners to.

samikeijonen:
Thanks @dkotter !

> For sites with lots of connections, this gives a dramatic bump in performance and offloads the loading of that data only when it's needed (when a user hovers over the menu and is looking to distribute content).

Yep, that's what I was thinking myself and noted in **Possible Drawbacks**.

I'm running out of time a little bit, can you or @dinhtungdu jump in this PR for that change?

samikeijonen:
@dinhtungdu Noting that this fix is only for the issue #563.

#557 is related for sure but what happens inside the distributor "menu" is a little bit more complicated. I'd like to do more markup changes in there.

samikeijonen:
@dinhtungdu I actually might have some time to look into this today. I let you know.

samikeijonen:
@dkotter & @dinhtungdu If you can check a6a061e3ca359e7f291e2cad3a4ff793416c6a01 if that works for you.

dinhtungdu (dismissed):
LGTM 👍

dkotter:
@samikeijonen @dinhtungdu We recently merged in #538, which ensures the Distributor menu stays open during distribution. That caused some merge conflicts here and I tried to clean those up myself but after doing and testing things again, I was seeing some issues here (basically breaks the changes we made in 538).

If either of you have time to fix things up here, that would be great.

jeffpaul:
This has been tested locally and confirmed passing tests, current check is failing as PR author doesn't have write access to the repo.

Issue #380, opened 2019-05-13
Featured Image not transfered


dmaslogh:
I've started using this plugin, run several tests on it and it's working fine except for the fact that's not transferring any Featured Image. Not from posts, pages or CPTs.

jeffpaul:
@dmaslogh welcome to Distributor and thanks for the feedback, it's greatly appreciated!  Can you confirm the following to help us triage your issue?

1. Are you using internal connections (via multisite) or external connections (e.g. separate WP installs)?
2. What filetype is/are the featured image(s) that aren't transferring?  Note that #180 already calls out an issue with SVGs, but I'm curious if you're seeing it with other filetypes.
3. What version of WordPress, Distributor, and other active plugins/theme are on your site?

dmaslogh:
@jeffpaul thanks!

1. Using internal multisite connection.
2. Most of them are JPG, but there could be some PNG as well. No SVG though.
3. Versions of the main packages are:
  - WP 5.1.1
  - Divi (latest 3.22.x)
  - Distributor 1.4.1
  - ACF 5.7.13
  - Events Manager 5.9.5
  - Gravity Form 2.4.8.9
  - and some other with less impact like wp-pagenavi, akismet, etc.

Hope this helps.

jeffpaul:
@dmaslogh there are some reports of weirdness with ACF, any chance the images are in a field there that aren't getting distributed or is it literally the default WP featured image field?

dmaslogh:
@jeffpaul it's the standard Featured Image from WP enabled in a CPT. We are not dealing with ACF image fields.

jeffpaul:
@dmaslogh is the CPT setup on both the origin and distributed sites?  If so, are these custom CPTs or are they created using a plugin (e.g. [CPT UI](https://wordpress.org/plugins/custom-post-type-ui/))?

dmaslogh:
@jeffpaul I've created a plugin that registers the same CPT on both sites. The CPTs are created by code, the internal fields are created using ACF.

jeffpaul:
@dmaslogh ok, that's helpful detail.  I'm labeling this as a `bug`, but if there's any further details on errors or code samples you can provide to help with further triage that would be greatly appreciated, thanks!

dmaslogh:
@jeffpaul I'll need to check with my client, but I think I can share with you some code if you need it. Just let me know if there's something specific you find useful for this research and I'll try to provide it asap.
Thanks!

[...]

PR #567, opened 2020-03-19
Use filesystem for copying media when doing a network pull/push

petenelson:
### Description of the Change
Use filesystem for copying media when doing a network pull/push instead of `download_url()`

### Benefits
For network/multisite on the same server, there should be no need to make an HTTP request to download media when doing a pull or push. In some instances, due to DNS resolution, firewalls, HTTP authentication, etc, downloading media via a URL may fail. This update uses the filesystem to copy media rather than downloading it.

### Possible Drawbacks
None that I am aware of.

### Verification Process
Added some debugging code to save `dt_original_file_path` to post meta that is only updated when a file is copied successfully. Performed both multisite pushes and pulls to confirm the media is copied.

### Checklist:
- [X] All new and existing tests passed.

### Applicable Issues
#380 

### Changelog Entry
* Use filesystem for copying media when doing a network pull/push instead of `download_url()`

### DIFF
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -144,7 +144,7 @@ class NetworkSiteConnection extends Connection {
 			 * @return {bool} If Distributor should push the post media.
 			 */
 			if ( apply_filters( 'dt_push_post_media', true, $new_post_id, $media, $post_id, $args, $this ) ) {
-				\Distributor\Utils\set_media( $new_post_id, $media );
+				\Distributor\Utils\set_media( $new_post_id, $media, [ 'use_filesystem' => true ] );
 			};
 		}
 
@@ -250,7 +250,7 @@ class NetworkSiteConnection extends Connection {
 				 * @return {bool} If Distributor should set the post media.
 				 */
 				if ( apply_filters( 'dt_pull_post_media', true, $new_post_id, $post->media, $item_array['remote_post_id'], $post_array, $this ) ) {
-					\Distributor\Utils\set_media( $new_post_id, $post->media );
+					\Distributor\Utils\set_media( $new_post_id, $post->media, [ 'use_filesystem' => true ] );
 				};
 			}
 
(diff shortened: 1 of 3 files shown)


includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -144,7 +144,7 @@ public function push( $post_id, $args = array() ) {
 			 * @return {bool} If Distributor should push the post media.
 			 */
 			if ( apply_filters( 'dt_push_post_media', true, $new_post_id, $media, $post_id, $args, $this ) ) {
-				\Distributor\Utils\set_media( $new_post_id, $media );
+				\Distributor\Utils\set_media( $new_post_id, $media, [ 'use_filesystem' => true ] );
```
dinhtungdu, line 147: The PR works in my test. But I think we only use filesystem for network connection, so we can use `DT_IS_NETWORK` to detect multisite installations without introducing a new argument for `set_media` function.

dinhtungdu: @petenelson What do you think about my comment above? (Forgot to tag you 😅)

jeffpaul:
As there's still a requested change on this PR, I'm going to move this to our next release so we can clear the 1.6 milestone and get to a release.

includes/classes/InternalConnections/NetworkSiteConnection.php
dkotter: @dinhtungdu So I think the only issue using that constant here is that is only set to true if we are on multisite and if Distributor is network activated. So there's a chance someone could be using multisite but not have the plugin network activated, just activated on the individual sites they want to use. In that case, using that constant wouldn't work

dkotter (approved):

dkotter:
Made some minor formatting changes and fixed the unit tests. This works well in my tests 👍

iaingooey:
Hi There, Is there any chance of 2.1 coming out? just regarding the Featured images not pulling through to duplicate sites?

Cheers

PR #569, opened 2020-03-26
Better keyboard support for distributor admin menu

### LINKED ISSUE #563: Distributor menu in admin bar - better keyboard support

samikeijonen:
Fixes #563

## Description of the Change

- Use `<button>` instead of `<div>`.
- Use Core button styles where possible for consistency and better focus styles.
- Use `disabled` attribute when post have been syndicated.
- This is not full refactoring but first step in better accessibility.

## Alternate Designs

None.

## Benefits

Better keyboard support when using correct semantics out of the box.

## Possible Drawbacks

Complex codebase, needs testing that nothing is broken.

## Verification Process

Distribute and push article using only keyboard. 

## Checklist:

## Applicable Issues

#563

### DIFF
--- a/assets/css/push.css
+++ b/assets/css/push.css
@@ -55,11 +55,13 @@
 #wpadminbar #distributor-push-wrapper .new-connections-list {
 	border: 1px solid var(--borderColor);
 	max-height: 125px;
+	overflow: auto;
 }
 
 #wpadminbar #distributor-push-wrapper .selected-connections-list {
 	margin-bottom: 1em;
-    max-height: 145px;
+	max-height: 145px;
+	overflow: auto;
 }
 
 #wpadminbar #distributor-push-wrapper .connections-selected {
--- a/assets/css/push.css
+++ b/assets/css/push.css
@@ -211,7 +211,7 @@
 }
 
 #wpadminbar #distributor-push-wrapper .add-connection.syndicated {
-	cursor: default;
+	cursor: not-allowed;
 }
 
 #wpadminbar #distributor-push-wrapper .add-connection.syndicated span {
--- a/assets/css/push.css
+++ b/assets/css/push.css
@@ -200,7 +200,7 @@
 	text-align: left;
 
 	&:focus {
-		box-shadow: 0 0 0 2px #fff;
+		box-shadow: inset 0 0 0 2px #fff;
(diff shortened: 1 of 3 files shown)


dinhtungdu (changes requested):
The code looks good @samikeijonen. There is a small issue that internal connections can't be selected using keyboard because they're still using `div`. Can you please update?

assets/css/push.css
```diff
@@ -177,55 +175,43 @@
...
-	border-width: 1px;
-    border-style: solid;
-	box-shadow: 0 1px 0 #006799;
-	color: #fff;
-	text-decoration: none;
-	text-shadow: 0 -1px 1px #006799, 1px 0 1px #006799, 0 1px 1px #006799, -1px 0 1px #006799;
-	height: 30px;
-	line-height: 28px;
-	padding: 0 12px 2px;
-	border-radius: 3px;
```
dinhtungdu, line 191: we should keep the `border-radius` and `text-shadow` because they're overridden by admin bar style.

samikeijonen:
@dinhtungdu Unfortunately I have zero hours for updates at the moment.

dinhtungdu:
@samikeijonen no problem, I will take over this.

dinhtungdu (dismissed):

dkotter:
@dinhtungdu Testing this locally, I'm seeing the connection list break if you have more than 4 connections:

Screen Shot 2020-05-15 at 11 28 51 AM

I believe once there are 5 or more connections, we constrain those to a box that you can scroll through, to keep the list down. Looks like something changed here that is preventing that from working properly, so the entire connection list is shown and there's no way to scroll through them all.

dinhtungdu:
Thank @dkotter, you're right, the style to handle overflow was removed. I added it again, the list is now working correctly with more than 4 connections.

dkotter (approved):

dkotter:
@dinhtungdu Looks good 👍

PR #570, opened 2020-03-27
Fix: wrong permission route

dinhtungdu:
### Description of the Change
The permission route in #245 which was merged to `develop` is wrong, that causes creating external connections failed. This PR fixes the URL and adds more checks to prevent potential warnings and errors.

### Verification Process
See the tests pass.

### Checklist:

### DIFF
--- a/includes/classes/ExternalConnections/WordPressExternalConnection.php
+++ b/includes/classes/ExternalConnections/WordPressExternalConnection.php
@@ -775,8 +775,6 @@ class WordPressExternalConnection extends ExternalConnection {
 			$output['errors']['no_distributor'] = 'no_distributor';
 		}
 
-		$routes = $data['routes'];
-
 		$types_path = untrailingslashit( $this->base_url ) . '/' . self::$namespace . '/types';
 
 		if ( function_exists( 'vip_safe_wp_remote_get' ) && \Distributor\Utils\is_vip_com() ) {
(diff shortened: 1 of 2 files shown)


jeffpaul:
I wonder if this may be why @dkotter and @helen are having different results in testing #368, assuming Helen pulled develop down after #245 was merged and Darin before?

dinhtungdu:
@jeffpaul I don't think so, that issue relates to Application Password load order.

dkotter:
@dinhtungdu Testing this out, I had an issue where it would connect to an external connection but wouldn't give me the proper message (gave me the yellow light instead of the green). Tracked down the issue and the problem was with the array we return to our ajax function. The array keys needed reset, otherwise it ended up as a javascript object and not an array, which was breaking our javascript functionality.

I've made that fix, as well as some minor code formatting fixes (that were mostly from the original PR). This looks good to me now.

dkotter:
@dinhtungdu Testing this out, I had an issue where it would connect to an external connection but wouldn't give me the proper message (gave me the yellow light instead of the green). Tracked down the issue and the problem was with the array we return to our ajax function. The array keys needed reset, otherwise it ended up as a javascript object and not an array, which was breaking our javascript functionality.

I've made that fix, as well as some minor code formatting fixes (that were mostly from the original PR). This looks good to me now.

dkotter (approved):

dkotter:
And just for performance numbers, on my local install, I saw these external connection test requests go from 8-9 seconds down to 1.5-2 seconds, so a pretty nice improvement there.

PR #574, opened 2020-03-31
Fix: Check for value of remote post id to verify push result

### LINKED ISSUE #488: Not handled response Bad Request from spoke site.

dinhtungdu:
### Description of the Change
We're checking for the existence of `post_id` key, not the value of it. There is a case where post_id equals `0`, and we still assume that the push is successful.
See https://github.com/10up/distributor/issues/488#issuecomment-605895066

For some reasons, I can't update #492, this PR is made to replace it.

### Verification Process
See https://github.com/10up/distributor/issues/488#issuecomment-60589506.

### Checklist:

### Applicable Issues
Fixes #488 and closes #492.

### DIFF
--- a/includes/classes/ExternalConnections/WordPressExternalConnection.php
+++ b/includes/classes/ExternalConnections/WordPressExternalConnection.php
@@ -647,9 +647,7 @@ class WordPressExternalConnection extends ExternalConnection {
 
 		$body_array = json_decode( $body, true );
 
-		try {
-			$remote_id = $body_array['id'];
-		} catch ( \Exception $e ) {
+		if ( empty( $body_array['id'] ) ) {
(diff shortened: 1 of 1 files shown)


dinhtungdu:
Note that we still need #570 to be merged for passing the tests.

dkotter (approved):

Issue #504, opened 2020-01-07
Confirm pull for users in Pull Content screen (redirect to Pulled list?)


hearvox:
After a post is pulled using the Dashboard> Distributor> Pull Content screen, user may not see the "Posts have been pulled" admin notice (I missed it), so not now the pull succeeded.

One solution is, after a pull, to redirect the Pulled content list (?page=pull&status=pulled).

jeffpaul:
@hearvox thanks for the feedback, we'll look at this in an upcoming release but would gladly pull this into an earlier release if someone's able to work on a PR for this.

jeffpaul:
Closed via #575.

PR #575, opened 2020-04-01
Redirect to pulled content tab after content is pulled via the settings

joshuaabenazer:
Takes care of #504 

### Description of the Change
Redirect to the pulled content tab after content is pulled via the Pull Content Dashboard page

### Benefits
The user will be taken to the Pulled Content page where in they will be able to view or edit the pulled content right after pulling it.

### Possible Drawbacks
If the user wants to pull more than a single post but does not use the bulk pull feature, then they may get annoyed that they would need to head back to the "New" tab. Although for multiple pulls, the bulk pull feature should be used instead.

### Verification Process
1. Head over to Dashboard -> Distributor -> Pull Content
1. Identify any post to be pulled and hit the "Pull" action or bulk "Pull" more than a single post
1. Verify that you are now on the "Pull" tab which should have a list of the most recently pulled items

### Checklist:

### Applicable Issues
#504 

### Changelog Entry

- Changed - Once content is pulled successfully via the "Pull Content" settings page, it redirects over to the "Pull" tab of the page instead of remaining in the default "New" tab

### DIFF
--- a/includes/pull-ui.php
+++ b/includes/pull-ui.php
@@ -253,7 +253,8 @@ function process_actions() {
 				setcookie( 'dt-duplicated', 1, time() + DAY_IN_SECONDS, ADMIN_COOKIE_PATH, COOKIE_DOMAIN, is_ssl() );
 			}
 
-			wp_safe_redirect( wp_get_referer() );
+			// Redirect to the pulled content tab
+			wp_safe_redirect( add_query_arg( 'status', 'pulled', wp_get_referer() ) );
 			exit;
 		case 'bulk-skip':
 		case 'skip':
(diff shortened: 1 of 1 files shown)


dinhtungdu (dismissed):
LGTM

dkotter:
@joshuaabenazer Thanks for this PR! This looks good to me but what are your thoughts on also doing this for items that are marked as skipped? Should those redirect to the Skipped tab instead of staying on the current tab? I think it makes sense to have both Pulled and Skipped items function similarly but open to your thoughts on that.

jeffpaul:
@joshuaabenazer any thoughts on @dkotter's feedback and potential addition for this PR?

dkotter (approved):

PR #578, opened 2020-04-20
Fix an an undefined variable issue

dkotter:
### Description of the Change

In #574, we removed a try/catch block in order to fix an issue where distribution didn't fully work and the post ID value is 0, instead of undefined. But within that try/catch block, we were setting a `$remote_id` variable that is then used twice after that.

Because that code was removed, that variable then becomes undefined when it is referenced later, causing issues. Items will still distribute correctly but the connection will not be fully set up, so updates will not be pushed through.

This PR fixes this issue.

### Alternate Designs

None

### Benefits

Fixes undefined variable PHP errors and ensures connections are set up properly.

### Possible Drawbacks

None

### Verification Process

1. Prior to this change, distribute a post to an external connection
2. Notice that the post was distributed successfully.
3. Try updating the post that was distributed
4. Notice updates were not pushed through
5. Check out this PR and run the same test, ensuring that updates happen now

### Checklist:

### DIFF
--- a/includes/classes/ExternalConnections/WordPressExternalConnection.php
+++ b/includes/classes/ExternalConnections/WordPressExternalConnection.php
@@ -655,10 +655,10 @@ class WordPressExternalConnection extends ExternalConnection {
 
 		if ( ! empty( $response_headers['X-Distributor'] ) ) {
 			// We have Distributor on the other side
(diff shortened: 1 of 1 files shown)


dinhtungdu (approved):
Thanks for this @dkotter!

Issue #561, opened 2020-03-17
Distributed post content get converted to classic (from block) after updating the original


dinhtungdu:
**Describe the bug**

After updating original post (blocks based content), the content of distributed posts is converted to single classic editor. In `is_using_gutenberg` function, we use `use_block_editor_for_post` which is return `false` in my tests; it even isn't loaded sometimes.
**Steps to Reproduce**

1. Setup distributor and external connection.
2. Publish a blocks based post.
3. Push that post to an external connection.
4. See the distributed post is block based.
5. Update the original post.
6. See the distributed post content gets converted to a single classic block.

**Expected behavior**

The content should be block based after updating the original post.

dkotter:
@dinhtungdu I was able to reproduce this as well, both with internal and external connections.

After some debugging, the issue for me had to do with Gutenberg making two requests to save a post (see https://github.com/WordPress/gutenberg/issues/12903). The first request correctly identifies we are using Gutenberg and updates the syndicated item with the raw content and then the second request has the failing `is_using_gutenberg` check and then updates the syndicated item with the rendered content, not the raw content.

We had partially solved for this for internal connections (see #518) but didn't bring those changes over to external connections. So I think that should be brought over and then we need to determine the best way to ensure updates only happen once and that they happen with the correct content (I have a few ideas but need to think them through a bit more)

dinhtungdu:
@dkotter Thanks for testing, I used `xDebug` and discovered that for the first request, the `use_block_editor_for_post` hasn't been loaded yet, so `is_using_gutenberg` return `true` because we check for the content of the post. For the second request, `use_block_editor_for_post` is loaded and it returns `false` because of the following block of code.

```
	if ( isset( $_GET['meta-box-loader'] ) ) {
		check_admin_referer( 'meta-box-loader', 'meta-box-loader-nonce' );
		return false;
	}
```

In my test, the content at the time it's passed to `is_using_gutenberg` is always block-based content. But due to `use_block_editor_for_post` returning `false`, the rendered content is used to update syndicated posts.

Also note that once the syndicated post content is converted to classic, it will be always classic even the raw content (blocks) is synced correctly because we check for Gutenberg status on both sides before populating content.

jeffpaul:
@dkotter is this issue fully resolved by #579?

dinhtungdu:
FYI, in my latest test on the current `develop` branch, I don't have this issue anymore.

jeffpaul:
[...]

PR #579, opened 2020-04-20
Fix some Gutenberg related issues

dkotter:
### Description of the Change

In #518, we added a check for Internal/Network connections, that when sending updates for Gutenberg posts, we run those updates on a later hook, to ensure all data is sent properly (as the REST API saves some things after the `save_post` hook we were using).

We never brought those same changes over to External connections, so this PR brings that same code to there. If using Gutenberg and we are in the `save_post` action, we utilize the `rest_after_insert_{$post->post_type}` hook instead, to ensure things like featured images and term information have been saved before we send updates.

In addition, as noted in #561, we have an issue where Gutenberg content will distribute correctly but when updates are pushed, all content will get put into a Classic block. The issue happens when you have custom meta boxes, Gutenberg will fire two save events: one for the actual post, one to save the custom meta. The first save request works fine but the second request will not correctly see our post as using Gutenberg, so the rendered content is distributed, not the raw content.

The issue lies within the core `use_block_editor_for_post` function we use. Within that, it checks for the existence of the `$_GET['meta-box-loader']` variable. If that exists (which it always will when custom meta boxes are being saved), it will always return `false`, which we assume that means the post isn't using Gutenberg. For now, I've added a check within our own `is_using_gutenberg` function, to bypass using `use_block_editor_for_post` if `$_GET['meta-box-loader']` is set.

### Alternate Designs

There are other ways to check if an item is using Gutenberg. I went with the most simple approach here but one of the other approaches might be better. One I had thought of but didn't take was duplicating the core `use_block_editor_for_post` function and removing the part that was failing, instead of bypassing that function all together.

### Benefits

Updates made to items that have already been distributed to external connections should now properly distribute those updates, including featured image and term information.

Also, updates made to Gutenberg items, for both internal and external connections, should now properly distribute the block information, instead of putting everything into a Classic block. 

### Possible Drawbacks

For the first change, none. For the change made to our `is_using_gutenberg` function, I don't think there are any but we are slightly changing how we detect Gutenberg posts, so there could be some unintentional impact there.

### Verification Process

1. Push a post to an external connection
2. Update the featured image of that post
3. Ensure that change was updated on the external site

1. Create a Gutenberg post and push to a connection
2. Update that post
3. Ensure that the syndicated item still retains block information

### Checklist:

### Applicable Issues

#561

### DIFF
--- a/includes/classes/ExternalConnections/WordPressExternalConnection.php
+++ b/includes/classes/ExternalConnections/WordPressExternalConnection.php
@@ -655,10 +655,10 @@ class WordPressExternalConnection extends ExternalConnection {
 
 		if ( ! empty( $response_headers['X-Distributor'] ) ) {
 			// We have Distributor on the other side
-			\Distributor\Subscriptions\create_subscription( $post_id, $remote_id, untrailingslashit( $this->base_url ), $signature );
+			\Distributor\Subscriptions\create_subscription( $post_id, $body_array['id'], untrailingslashit( $this->base_url ), $signature );
 		}
 
-		return $remote_id;
+		return $body_array['id'];
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


helen (approved):
This looks good to me, it's frustrating we keep running into issues in this block detection method because we were trying to stick with core as much as possible, let's just be sure to keep an eye on this in the future.

PR #586, opened 2020-06-02
Ensure there isn't a trailing slash on the REST URL before we break t…

### LINKED ISSUE #585: View links for pushed content sometime use REST URL
**Describe the bug**
While testing #368, I pushed a post to an external connection and the View link that shows after distributing pointed to a `wp-json` URL, which is not correct.

View links for external connections were added in #571. In particular, there's a new `get_site_url_from_rest_url` function that tries to get the site URL from the REST URL. This function takes in the REST URL, breaks that down into it's various pieces, removes the last piece (in theory, the `wp-json` portion, or whatever REST prefix the site is using, if customized) and puts the URL back together.

What I'm finding is if you have an external connection that has a trailing slash, when that URL is broken into it's various pieces, the last piece will be an empty string. That last piece is the only piece we remove and  we then put the URL back together, which will keep the `wp-json` part. We then use that URL for the syndicated URL, so it ends up looking like `example.com/wp-json/?p=10`.

There may be a better way to parse the URL here but at the very least, we should probably ensure there isn't a trailing slash on the REST URL, before we parse it.

**Steps to Reproduce**
1. Create a new external connection with a trailing slash
2. Push a post to this connection
3. Check the View link that is added once distribution is done
4. See link pointing to the REST URL, not the site URL

**Expected behavior**
After distributing a post to an external connection, the View link should point to the FE URL, not the REST URL.

dkotter:
### Description of the Change

Ensure there isn't a trailing slash on the REST URL before we break that URL into it's various pieces, in our `get_site_url_from_rest_url` function.

Fixes #585

### Benefits

View links for external connections will be correct if that external connection is set up with a trailing slash

### Possible Drawbacks

None

### Verification Process

1. Created a new external connection with a trailing slash
2. Pushed a post to this connection
3. Verified the View link pointed to the REST URL (i.e. example.com/wp-json/?p=10)
4. Added the fixes here and tested again
5. Verified the View link pointed to the FE URL (i.e. example.com/?p=10)

### Checklist:

### Applicable Issues

#585

### DIFF
--- a/includes/push-ui.php
+++ b/includes/push-ui.php
@@ -440,7 +440,7 @@ function menu_button( $wp_admin_bar ) {
  * @return string Site URL.
  */
 function get_site_url_from_rest_url( $rest_url ) {
-	$_url = explode( '/', $rest_url );
+	$_url = explode( '/', untrailingslashit( $rest_url ) );
 
 	if ( count( $_url ) < 2 ) {
 		return $rest_url;


dinhtungdu (approved):

jeffpaul:
I'll hold off on merging this until we get the auth wizard PR merged in so that we don't create further conflicts on that merge.

jeffpaul:
The auth wizard PR is merged and it doesn't appear that we've got any merge conflicts here.  So @dinhtungdu @dkotter you two 👍 to merge this one in?

dkotter:
Good on my end

PR #589, opened 2020-06-09
Style and UX fixes for push UI

### LINKED ISSUE #577: Select All button placement and style
In #495, a Select All button and a None button were added to the Distributor drop-down menu. This makes it easy to select all connections to distribute to at once, rather then having to select these individually.

Right now these buttons show above the connection list but seems like maybe we would want to move these buttons below that list, maybe even right align instead of left align?

Also, we use the same style for both the Select All and None buttons. Thinking maybe the None button should take the secondary button style that WordPress has, or maybe even just a text link?

Here are some screenshots on how it currently looks:

Default with no connections currently selected:
Screen Shot 2020-04-13 at 8 48 53 AM

All connections selected:
Screen Shot 2020-04-13 at 8 49 00 AM

Some connections selected but not all:
Screen Shot 2020-04-13 at 8 49 11 AM

helen:
### Description of the Change

Polishing style and UX bits for the push UI after adding `Select All` button.

I currently have 3 things left on my list of nice-to-have items for this PR:
- [x] Better no search results state (currently shows empty box with collapsed border, should probably show a "No results" item in the box)
- [ ] Show the number of selected connections - this needs i18n consideration in whatever string, easier to just put the count in parens rather than trying to pluralize
- [x] ~Add some kind of subtle indicator that a list can be scrolled, this tripped me up a few times. [Example](https://css-scroll-shadows.now.sh/?bgColor=8680ce&shadowColor=222222&pxSize=15)~ Can't use this method due to background colors, haven't found anything else, skipping for now

#### Before
Screen Shot 2020-06-08 at 8 15 15 PM

#### After
Screen Shot 2020-06-08 at 8 08 54 PM

### Alternate Designs

n/a

### Benefits

A UI that's easier to understand.

### Possible Drawbacks

Not sure.

### Verification Process

Local testing with various states (search, selection, distributed, clearing). I also tested keyboard nav which looks fine to me.

### Checklist:

### Applicable Issues

Fixes #577 

### Changelog Entry

Needs.

### DIFF
--- a/assets/css/push.css
+++ b/assets/css/push.css
@@ -65,12 +65,17 @@
 
 #wpadminbar #distributor-push-wrapper .new-connections-list {
 	border: 1px solid var(--borderColor);
-	max-height: 125px;
+	max-height: 155px;
 	overflow: auto;
+
+	& .no-results {
+		margin: 0 !important;
+		padding: 0.5em;
+	}
 }
 
 #wpadminbar #distributor-push-wrapper .selected-connections-list {
-	margin-bottom: 1em;
+	margin-bottom: 0.5em;
 	max-height: 145px;
 	overflow: auto;
 }
@@ -223,6 +228,7 @@
 
 	&.selectall-connections {
 		margin-top: .5em;
+		margin-bottom: -10px; /* compensation for margin above messages div to reduce jumping */
 	}
 }
 
--- a/assets/css/push.css
+++ b/assets/css/push.css
@@ -95,13 +95,15 @@
 		display: none;
 	}
 }
-#wpadminbar #distributor-push-wrapper .selectall-connections.unavailable,
-#wpadminbar #distributor-push-wrapper .selectno-connections.unavailable {
-	opacity: .5;
+
+#wpadminbar #distributor-push-wrapper .connections-selected button.selectno-connections {
+	margin-left: 10px;
 }
+
 #wpadminbar #distributor-push-wrapper .connections-selected label {
 	color: inherit;
 }
+
 #wpadminbar #distributor-push-wrapper .connections-selected input[type="checkbox"] {
 	vertical-align: middle;
 	margin-right: 2px;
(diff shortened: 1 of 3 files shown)


helen:
New no results state:

Screen Shot 2020-06-09 at 7 38 20 PM

helen:
I'm going to leave the selected connections count for another issue/PR another day, this is ready for secondary review and merge.

dkotter:
@helen Tested this out and this looks really good to me, definitely better then what we currently have. Did have a few comments though:

For me, the Clear link doesn't line up with the Selected Connections text:
Screen Shot 2020-06-11 at 8 42 20 AM

This is more just my opinion but if you click the Select All button, wondering if we should disable that button, as clicking it again won't do anything at that point. At the moment, the button looks the same in both states:
Screen Shot 2020-06-11 at 8 42 41 AM

helen:
@dkotter Which browser are you using, and is this in the admin or on the front end? I've been checking in both Firefox and Chrome on Mac, not sure if maybe there are other factors that could be impacting this as well, since I noticed that themes on the front end can impact the CSS here. I notice your buttons are tall as well.

As for disabling the Select All button, I am not opposed to it, I just don't know if it will end up being distracting turning on and off as you filter/search for connections? It doesn't seem harmful to leave it just not doing anything, but I can certainly try it and see how it feels locally.

dkotter:
I tested in the latest version of Chrome and Firefox (in the admin) and I see the same in both.

If I turn off the min-height and vertical-align styles, things look better to me:
Screen Shot 2020-06-11 at 3 25 48 PM

Seems like those are core styles but maybe I have some other conflict here. I didn't see one at a quick glance but it's possible.

helen:
Changing the vertical alignment led to some other issues for me but removing the min-height seems to have worked out for me in the admin, so I pushed a change for that. One more look and then let's merge?

dkotter (approved):

dkotter:
👍 That fixed it, looks good for me now

PR #590, opened 2020-06-09
Update subscriptions.php

pascalknecht:
### Description of the Change

Changing priority of the hook to 99 allows for plugins that are hooked to save_post with a normal priority to do their work before syncing happens

### Benefits

Meta values from plugins like yoast will be shared properly.

### Possible Drawbacks

Maybe take 98 to allow plugins to hook in after this plugin on purpose.

### Verification Process

Synced a post with the primary category from yoast defined. The id of the category was now correct.

### Checklist:

### DIFF
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -472,7 +472,7 @@ class NetworkSiteConnection extends Connection {
 	 */
 	public static function bootstrap() {
 		add_action( 'template_redirect', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'canonicalize_front_end' ) );
-		add_action( 'save_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'update_syndicated' ) );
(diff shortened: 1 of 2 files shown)


jeffpaul:
@pascalknecht welcome to Distributor and thanks for the PR!  In general we like to start with an Issue first to ensure agreement on the problem and ideal approach on solving it.  In lieu of that here, do you have any examples of how this change is useful or tests that help demonstrate this?

pascalknecht:
Hi @jeffpaul 
I can give you an example:
When you save the primary category in yoast the id will be saved in the save_post hook with the same priority as distributor uses to distribute that post. Now its unclear which function will be executed first and distributor may distribute incorrect information.

This is just one example. Any plugin modifying the post in the save_post hook with the default priority could run into this issue.

dkotter:
@pascalknecht Thanks for the PR. I think using 99 here is fine as a priority. The one suggestion I would make though is we should also add this priority to our `save_post` hook that fires for internal connections. What you have in this PR only fires for external connections and probably makes sense to standardize those.

Can see that hook here: https://github.com/10up/distributor/blob/develop/includes/classes/InternalConnections/NetworkSiteConnection.php#L493

dkotter (approved):

PR #630, opened 2020-08-24
Update: Error reporting

dinhtungdu:
### Description of the Change
This PR brings a better error reporting system to distributor:
- Display detailed error while creating a new external connection through the Wizard. Covered cases:
    - Invalid domains.
	- The remote site is not a WordPress installation.
- Update connection detail page:
	- Clearly state that if the account is not authenticated or doesn't have permission to post.
	- Add a new permission table for the current account.
- Display pull errors after pulling from remote, [screenshot](https://user-images.githubusercontent.com/5423135/90996554-4cfcd480-e5e9-11ea-8629-8cb7318293ac.png).

- Update and display detailed errors for push UI, [screenshot](https://user-images.githubusercontent.com/5423135/90996527-35bde700-e5e9-11ea-89f7-b23c3b4c5393.png).

## Possible Drawbacks

The return value of `push` methods is changed from `int` to `array`. So it may cause compatibility issues for sites that uses those methods directly.

### Verification Process
Manual verification.

### Checklist:


includes/pull-ui.php
```diff
@@ -458,3 +473,56 @@ function dashboard() {
...
+
+	<div id="pull-errors" class="notice notice-warning is-dismissible">
+		<p><?php esc_html_e( 'Some errors occurred while pulling selected posts. Please review the list of the errors bellow for more details.', 'distributor' ); ?></p>
+		<ul>
+			<?php foreach ( $pull_errors as $id => $errors ) : ?>
+			<li>
+				<strong><?php esc_html_e( $posts[ $id ] ); ?>:</strong>
+				<ul>
+				<?php foreach ( $errors as $error ) : ?>
+					<li><?php esc_html_e( $error ); ?></li>
```
dkotter: Should just use `echo esc_html` here and line 516, as we aren't translating these strings here.

assets/js/push.js
```diff
@@ -87,33 +89,73 @@ jQuery( window ).on( 'load', () => {
...
-		setTimeout( () => {
-			distributorPushWrapper.classList.remove( 'message-error' );
-		}, 6000 );
+		_.each( prepareMessages( messages ), function( message ) {
+			const errorItem = document.createElement( 'li' );
+			errorItem.innerText = message;
+			errorDetails.appendChild( errorItem );
+		} );
+
+		// setTimeout( () => {
```
dkotter: If this isn't needed anymore, should probably just remove it

includes/external-connection-cpt.php
```diff
@@ -516,6 +526,34 @@ function meta_box_external_connection_details( $post ) {
 		<ul class="endpoint-errors"></ul>
 	</div>
 
+	<?php if ( ! empty( $external_connection_status ) ) : ?>
+	<div class="post-types-permissions hide-until-authed">
```
dkotter: Minor thing but this block should probably be indented within this `if` statement

includes/external-connection-cpt.php
```diff
@@ -516,6 +526,34 @@ function meta_box_external_connection_details( $post ) {
...
+		<h4><?php esc_html_e( 'Post types permissions', 'distributor' ); ?></h4>
+
+		<table class="wp-list-table widefat">
+			<thead>
+				<th><?php esc_html_e( 'Post types', 'disthibutor' ); ?></th>
+				<th><?php esc_html_e( 'Can pull?', 'disthibutor' ); ?></th>
+				<th><?php esc_html_e( 'Can push?', 'disthibutor' ); ?></th>
```
dkotter: Typo on the text domain here and the lines above

includes/external-connection-cpt.php
```diff
@@ -516,6 +526,34 @@ function meta_box_external_connection_details( $post ) {
...
+			</thead>
+			<tbody>
+				<?php foreach ( $post_types as $post_type ) : ?>
+					<?php
+					if ( 'dt_subscription' === $post_type->name ) {
+						continue;
+					}
+					?>
+					<tr>
+						<td><?php echo $post_type->label; ?></td>
```
dkotter: Should have `esc_html` around this `echo`

includes/classes/ExternalConnections/WordPressExternalConnection.php
```diff
@@ -658,7 +658,15 @@ public function push( $post_id, $args = array() ) {
...
-		return $body_array['id'];
+		$remote_post = array(
+			'id' => $body_array['id'],
+		);
+
+		if ( ! empty( $body_array['push-errors'] ) ) {
+			$remote_post['push-errors'] = $body_array['push-errors'];
+		}
+
+		return $remote_post;
```
dkotter, line 669: Couple things here.

1. Since we are changing from returning just an ID to an array, we should change the dockblock for this function
2. This isn't a code review comment but because we are changing what this function returns, we'll need to be very clear in our release notes, so any 3rd party integrations that might utilize this `push` function can update accordingly, if needed

      2a. This also means we return different things in the `push` method for Internal and External connections now. Might be a factor once we start work on cleaning up and standardizing how we push and pull between connection types

dkotter (changes requested):

dkotter:
@dinhtungdu This looks really good, nice work here! I left some minor code review comments, let me know if you have any questions on those.

Testing this out locally, everything seems to work great, didn't run into any issues.

I left this in one of my review comments but the main thing I think we need to be clear on here is we're changing the return of the external connections `pull` method, from returning an ID to an array. If anyone has built add-ons or customizations that utilize that method, it could break things for them.

helen (approved):
Opened #635 to track the return type change of `push()` (may need to update it to reflect that it actually applies to network connections as well), otherwise this seems fine to me.

helen (approved):

dkotter (approved):

Issue #267, opened 2018-11-29
Do internationalization properly


helen:
Coming out of #266, let's do internationalization properly! These are probably worth breaking out into sub-issues:

- [x]  Fix up the way strings are internationalized generally (I see a bunch of English-centric assumptions that don't actually work such as appending an `s` to pluralize)
- [x] Do everything correctly to get things registered in PHP and JS
- [ ] Add JS PO file generation to the build process
- [ ] Lay out a plan for how we source and include translations

jeffpaul:
Noting that we may want to consider including https://github.com/varunsridharan/action-wp-pot-generator as part of our release process.

peterwilsoncc:
> Add JS PO file generation to the build process

See #947 

> Do everything correctly to get things registered in PHP and JS

Covered by #916 (use wp-scripts) and #935 (fixing untranslable strings)

> Lay out a plan for how we source and include translations

Some discussion of using [wearerequired/traduttore](https://github.com/wearerequired/traduttore) in #660 but under discussion.

jeffpaul:
Noting that I would be fine holding off on `Lay out a plan for how we source and include translations` as long as the POT file generated is as accurate and complete as we can make it (as then Distributor users can generate their own translation files for locales of particular need and include those themselves).  Current research seems to point to sourcing and including translations may be more work than eventual benefit (for a plugin not able to make use of the translate.wordpress.org infrastructure).

PR #662, opened 2020-11-02
Fix: i18n po/mo not loaded

### LINKED ISSUE #660: i18n po/mo not loaded
**Describe the bug**
I translated Distributor in french but only strings in the plugin page but the translations are not loaded in the plugin setings it self:
distrubor-fr
distrubor-fr-settings

[distributor-fr_FR.zip](https://github.com/10up/distributor/files/5432900/distributor-fr_FR.zip)

**Environment information**
 - WordPress version: 5.5.1

dinhtungdu:
### Description of the Change
This PR loads textdomain properly, allowing distributor to ship translations with future release. Props @sebastienserre for reporting this.

### Verification Process
See #660 

### Checklist:

### Applicable Issues
Fixes #660

### DIFF
--- a/distributor.php
+++ b/distributor.php
@@ -183,6 +183,8 @@ add_action(
 		) {
 			\Distributor\Connections::factory()->register( '\Distributor\InternalConnections\NetworkSiteConnection' );
 		}
+
+		load_plugin_textdomain( 'distributor', false, dirname( plugin_basename( __FILE__ ) ) . '/lang' );
 	},
 	1
 );


jeffpaul:
@dinhtungdu will want to make sure we confirm the checklist items on #267 for this PR.

distributor.php
```diff
@@ -183,6 +183,8 @@ function() {
 		) {
 			\Distributor\Connections::factory()->register( '\Distributor\InternalConnections\NetworkSiteConnection' );
 		}
+
+		load_plugin_textdomain( 'distributor', false, dirname( plugin_basename( __FILE__ ) ) . '/lang' );
```
dkotter, line 187: So this loads from the `lang` directory but that isn't a directory that we currently have in Distributor. I get that this is setting us up for the future but may be worth adding that directory as part of this PR?

dkotter:
@jeffpaul @dinhtungdu Left a comment on the PR but leaving it here as well. This PR loads a textdomain but doesn't actually add the `lang` directory (or add any translations). I understand this is just a single step in the right direction and actual translations can come later but I think we might want to add the `lang` directory as part of this PR, so it's there and ready to drop translations into.

jeffpaul:
I believe the `lang` directory and POT file get built into the `stable` branch as part of the release action, but maybe that's worth having in `develop` and `trunk` as well?  Will defer to @helen, though may just be worth punting this if there's any discussion needed so as to not hold up the 1.6.2 release.

dkotter:
> I believe the `lang` directory and POT file get built into the `stable` branch as part of the release action, but maybe that's worth having in `develop` and `trunk` as well? Will defer to @helen, though may just be worth punting this if there's any discussion needed so as to not hold up the 1.6.2 release.

Ah, that makes sense. But yeah, does bring up the point of how do people submit a PR with language files if there isn't a `lang` directory to put those files in (which was brought up on #660)

I'm fine with this PR but there's definitely more work around translations that needs done, so could argue for waiting on this until we have the other parts in place.

jeffpaul:
I'm going to punt to 1.7.0 to give us time to chat about if/how we want to handle MO/PO files and other topics related to translations.

distributor.php
dinhtungdu: @dkotter We have that directory in the `stable` branch. We build the pot file and put it into `lang` folder for the release.

peterwilsoncc:
This was resolved in e6f1dd37194b7276d53b37e5a21d3a68fdb01fb2 for another PR, so closing this off.

PR #665, opened 2020-11-10
Make sure the Distributor admin bar works correctly when loading via AMP

### LINKED ISSUE #658: Support for AMP - amp-dev-mode
We recently launched a fully AMP site. AMP Standard mode for logged-in users strips out any non-compliant styles or scripts so our jQuery and Underscores based JS is getting stripped from admin bar making it less convenient for users to Distribute a post. 

https://github.com/ampproject/amp-wp/issues/5533

Weston pointed toward this blog post. 
https://weston.ruter.net/2019/09/24/integrating-with-amp-dev-mode-in-wordpress/
 
Appears we need to add admin-bar as dependency to any enqueues and 
data-ampdevmode to any inline script tags.

dkotter:
### Description of the Change

When using the official [AMP WP plugin](https://github.com/ampproject/amp-wp), if you have the front-end of your site set to always use AMP and you're logged in, the Distributor admin bar item shows but does not work. The issue lies with AMP removing the assets we rely on, as well as mangling the templates we output.

This PR ensures all of our assets are loaded via amp-dev-mode, so they don't get removed. It also changes our template system too utilizing Mustache, which AMP supports.

### Alternate Designs

None

### Benefits

Front-end distribution will now work properly when using AMP

### Possible Drawbacks

We now have two different template systems: one used only for AMP and the previous one we use elsewhere. Ideally we would use the same templates for both but for now, I think it's safer to keep those separate.

### Verification Process

1. Install the AMP WP plugin and get that set up
2. Ensure you are logged in and visit the frontend view of a single post
3. Hover over the Distributor admin bar item and make sure it loads
4. Choose a connection to distribute to and make sure that works

### Checklist:

### Applicable Issues

Closes #658

### DIFF
--- a/assets/js/push.js
+++ b/assets/js/push.js
@@ -192,9 +192,14 @@ jQuery( window ).on( 'load', () => {
 			}
 
 			if ( 'mustache' === template ) {
+				if ( selectedConnections[connection.type + connection.id] ) {
+					connection.added = true;
+				} else {
+					connection.added = false;
+				}
+
 				showConnection = Mustache.render( document.getElementById( 'dt-add-connection' ).innerHTML, {
 					connection: connection,
-					selectedConnections: selectedConnections
 				} );
 			} else {
 				showConnection = template( {
@@ -276,12 +281,12 @@ jQuery( window ).on( 'load', () => {
 			if ( 'mustache' === template ) {
 				const mustacheData = { 'connections' : [] };
 				for ( const prop in dtConnections ) {
-					mustacheData['connections'].push( dtConnections[prop] );
+					mustacheData.connections.push( dtConnections[prop] );
 				}
 
 				distributorPushWrapper.innerHTML = Mustache.render( document.getElementById( 'dt-show-connections' ).innerHTML, {
-					connections: mustacheData['connections'],
-					foundConnections: mustacheData['connections'].length
+					connections: mustacheData.connections,
+					foundConnections: mustacheData.connections.length
 				} );
 			} else {
 				distributorPushWrapper.innerHTML = template( {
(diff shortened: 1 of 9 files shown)


jeffpaul:
Something looks off from Travis, haven't dug in yet to see what's erroring there but may need to if this fails testing with @rickalee.

rickalee:
@dkotter Great work so far. Looks like we still need to address jQuery compatibility. Screenshot is AMP Validation errors.

https://rickal.ee/DOu9pBBp

After I 'Kept' jQuery in AMP Validation, I was able to successfully push to External Location from frontend.

dkotter:
> @dkotter Great work so far. Looks like we still need to address jQuery compatibility. Screenshot is AMP Validation errors.
> 
> https://rickal.ee/DOu9pBBp
> 
> After I 'Kept' jQuery in AMP Validation, I was able to successfully push to External Location from frontend.

@rickalee Thanks for the review! I do add the dev mode attribute to the jQuery script but am currently looking for a script with an id of `jquery-core-js`, whereas your screenshot is showing an id of `jquery-js`.

The former was what I was seeing when testing so I'll look at this again and see if there's a more resilient way to look for jQuery that will work even if that ID changes (though I'm not sure why the ID is different, something else I can look at)

dkotter:
@rickalee I've added some code that I think makes the handling of dependencies (in our case, jQuery and underscores) a bit more resilient. I couldn't ever reproduce a scenario that gave me a jQuery file with the ID of `jquery-js`, it was always `jquery-core-js`, so I wasn't able to fully test the scenario you ran into so if you could test again and let me know if this now works properly, that would be great.

rickalee:
@dkotter Confirmed that Distributing is working on frontend and not getting any AMP Validation errors. Also confirmed jQuery isn't added to frontend for non-logged in users. 

I think we are set with this feature. However found another character encoding bug with apostrophe/single quote with titles but will open a different ticket. 

https://rickal.ee/KouwpQ59

PR #672, opened 2020-12-07
Fix: HTML Entity Character Encoding

### LINKED ISSUE #668: HTML Entity Character Encoding in post_title on Distributed Posts
Seeing right single quote being encoded on Distributed Post but correct on Source post. 

Cole Haan’s Summer Sale offers up to 75% off sandals, more + free shipping

https://www.codetable.net/decimal/8217

https://rickal.ee/KouwpQ59

dinhtungdu:
This PR fix the issue of distributed posts' title which contain HTML entities.

Fixes #668


jeffpaul:
@rickalee does this PR resolve your related issue?

includes/classes/ExternalConnections/WordPressExternalConnection.php
```diff
@@ -543,7 +543,7 @@ public function push( $post_id, $args = array() ) {
 		 * Now let's push
 		 */
 		$post_body = [
-			'title'                          => get_the_title( $post_id ),
+			'title'                          => html_entity_decode( get_the_title( $post_id ) ),
```
dkotter: @dinhtungdu Wondering if we should pass in the character set we want to use for decoding? I think by default it is UTF-8, which is probably good for most use cases, but some sites may use a different encoding.

Looking at core, most places they use `html_entitiy_decode`, they have it set up like this: `html_entity_decode( $content, ENT_QUOTES, get_bloginfo( 'charset' ) )`

composer.json
```diff
@@ -10,32 +10,24 @@
...
-            }
-        }
-    }
-  ],
   "require": {
     "php": ">=5.6",
     "yahnis-elsts/plugin-update-checker": "^4.4",
-    "georgestephanis/application-passwords": "0.1.1",
-    "ext-json": "*"
+    "ext-json": "*",
```
dkotter: @dinhtungdu Is there a reason the `composer.json` and `composer.lock` files have changes as part of this PR? Doesn't seem like they are needed, so I'm assuming these were just accidentally brought in.

dkotter (changes requested):

composer.json
dinhtungdu: It's my fix to bundle Application password properly. Currently, it locks to 0.1.1. But I agree that it isn't relevant to this PR.

includes/classes/ExternalConnections/WordPressExternalConnection.php
dinhtungdu: Nice catch! Thank you!

dkotter (dismissed):

PR #682, opened 2020-12-12
Docs: Add guide for enabling Application Passwords on 5.6

dinhtungdu:
### Description of the Change
This PR adds a new section for Application Passwords and 5.6 debug guide. This also updates the auth page URL to make the Auth Wizard work on 5.6.

### Verification Process
- Update a distributor development site without HTTPS to 5.6.
- Notice the Application Passwords section disappears on the user edit page.
- Add a new external connection, don't see the wizard.
- Follow the guide in this PR to enable Application Passwords.
- See the Application Passwords section on the user edit page.
- Add a new external connection, see the wizard working.

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,9 +174,9 @@ In WordPress 5.6, Application Passwords is merged into the WordPress core with s
 ```php
 add_filter( 'wp_is_application_passwords_available', '__return_true' );
 
-add_action('wp_authorize_application_password_request_errors', function( $error ) {
+add_action( 'wp_authorize_application_password_request_errors', function( $error ) {
     $error->remove( 'invalid_redirect_scheme' );
-});
+} );
 ```
 
 ## Changelog
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ _Note:_ The latest stable version of the plugin is the _stable_ branch. [Downloa
   * [Running Locally](#running-locally)
   * [Testing](#testing)
   * [Debugging](#debugging)
+  * [Application Passwords](#application-passwords-and-wordpress-56)
 * [Changelog](#changelog)
 * [Contributing](#contributing)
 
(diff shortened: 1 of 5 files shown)


dkotter (dismissed):

dkotter:
Code looks fine here. Tested on both WP 5.5 and 5.6 and on 5.5, everything worked great. I did run into one issue on 5.6 though. If you try and use the wizard to create a connection to another site that isn't on 5.6 yet, you'll end up getting sent to a 404 page (because the `authorize-application.php` page won't exist unless you're on 5.6).

Not sure if this is something we want to try and fix here, checking what WordPress version the external site is running and using the `admin.php?page=auth_app` page if it's not on 5.6? Or if we want to leave this alone and just add a note in the wizard page, mentioning how the external site needs to be on 5.6 (similar to the note we already have in there about how both sites need to be running Distributor 1.6.0 or greater)?

dinhtungdu:
@dkotter Nice catch! I missed that important point. The Application Password endpoint should be based on the remote site, not the origin site. I pushed a solution to fix that.

There is still a case that hasn't been covered: the remote site is updated to WP 5.6 but Distributor isn't. We can't get the WP version or the core Application Passwords status through REST API. A dev note in the readme is the only thing we can do IMO.

dinhtungdu:
Here is what happens if Application Passwords is not available on the remote site:
image

jeffpaul:
@dinhtungdu does that minor docs change help call out what we're requiring or should we be more verbose/explicit?

dkotter (approved):

TimothyBJacobs:
Just wanted to mention that you can check whether the target site supports app passwords by making a request to the REST API Index ( `/wp-json` ) and looking for `application-passwords` in the authentication section. See `rest_add_application_passwords_to_index` for details. That should also be the preferred way to get the URL to redirect the user to instead of hardcoding `authorize-application.php`.

jeffpaul:
@dkotter is there perhaps an update we should make in how Distributor is leveraging App Pwds based on @TimothyBJacobs note above?

dkotter:
Yeah, that seems like a better approach. Probably worth opening an issue to get things updated to follow best practices

PR #701, opened 2021-01-21
Add option to set post_status of pulled content

### LINKED ISSUE #695: Pulled posts are automatically published
Hello, I started using the plugin last week and it was exactly what I needed for a large multisite network. 
However as I started working with it I found some issues, oine of them is that once I pull a post, it is pushed as Published automatically instead of being kept as a draft, as I read is the default behaviour since the latest version.
I tried it also on a clean install  with Wordpress default theme and no plugins and I ran into the same issue. 
I'm using the latest version of Wordpress and the plugin ( 1.6.2). 

I wonder if there is some setting which I'm missing?

Thanks for your help.

dkotter:
### Description of the Change

When pushing content, you have the ability to choose if that content should be sent in a draft state or not (draft is the default). But when pulling content, you don't have that option and items will come over with whatever `post_status` they have (typically published).

Having content automatically be published is probably not the ideal workflow, so this PR adds a new `Pull in as draft` checkbox to the Pull Content screen. This checkbox is checked by default and if checked, if content is pulled in, either individually or in bulk, the `post_status` will be set as draft on those items.

There is a filter around this value so if someone wants the default to be unchecked, they can use that filter to accomplish this.

Screen Shot 2021-01-21 at 4 12 31 PM

### Alternate Designs

We could just force pulled in content to be set as draft, rather than introducing a new checkbox control

### Benefits

Gives users the ability to choose whether pulled in content should be set in a draft state (which will now be the default) or if they want pulled content to inherit the `post_status` of the original item (which is the state if this new checkbox is unchecked).

### Possible Drawbacks

None

### Verification Process

Go to the Pull Content screen and see if the new `Pull in as draft` checkbox shows. Try pulling content individually and in bulk to ensure they come over as draft. Now uncheck that box and run the same tests again, ensuring items come over with their original `post_status`.

### Checklist:

### Applicable Issues

Fixes #695

### DIFF
--- a/assets/css/admin-pull-table.css
+++ b/assets/css/admin-pull-table.css
@@ -9,6 +9,15 @@
 			cursor: default;
 		}
 	}
+
+	& .dt-pull-post-type {
+		display: flex;
+		align-items: center;
+
+		& .dt-as-draft {
+			padding-left: 10px;
+		}
+	}
 }
 
 .wp-list-table .disabled {
--- a/assets/js/admin-pull.js
+++ b/assets/js/admin-pull.js
@@ -1,4 +1,5 @@
 import jQuery from 'jquery';
+import { addQueryArgs } from '@wordpress/url';
 
 const chooseConnection = document.getElementById( 'pull_connections' );
 const choosePostType = document.getElementById( 'pull_post_type' );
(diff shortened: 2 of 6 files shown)


dinhtungdu (commented):
The code looks good to me. The draft checkbox handler works flawlessly. Just some minor adjustments and this is ready to be merged.

- The checkboxes should be aligned middle vertically.
- The Pull row action link is syncing with the checkbox (really nice), but we can make it better by updating the action label. For example: `Pull as draft` if the checkbox is checked.

dkotter:
@dinhtungdu Thanks for the feedback! I've updated the styling on the checkbox to vertically align it (and give it a little more spacing) and I've updated the inline Pull action link text to reflect the value of the draft checkbox.

Screen Shot 2021-02-04 at 11 26 29 AM

dinhtungdu (approved):
🎉

francisco-domo:
Hello @jeffpaul I wanted to download this branch because I need this feature but I see you deleted it... can I find it anywhere else?
Thanks in advance.

dinhtungdu:
Hi @francisco-domo You can download the develop branch to use this feature now.

PR #703, opened 2021-01-22
Fix some Pull UI errors

### LINKED ISSUE #402: PHP Notice: Undefined offset: 0 on pull content page
**Describe the bug**
`Notice: Undefined offset: 0 in /plugins/distributor/includes/pull-ui.php:394`

There is no check if there are ANY available post types.
```php
$connection_now->pull_post_types = \Distributor\Utils\available_pull_post_types( $connection_now, $connection_type );
$connection_now->pull_post_type = $connection_now->pull_post_types[0]['slug'];
```
https://github.com/10up/distributor/blob/1.4.1/includes/pull-ui.php#L394

**Steps to Reproduce**

1. Go to Pull Content page wp-admin/admin.php?page=pull
2. See PHP Notice

**Expected behavior**
No PHP Notice

dkotter:
### Description of the Change

As reported in #402, if a connection has no supported post types for pulling, a PHP notice is shown (if debug is turned on). Fixing this, I also realized that if a connection has no post types available, we default to still showing content from the `post` post type, which seems non-ideal. So this PR fixes that as well, so if there are no post types available for pulling, we show the `No items found` message.

Also in addition, while testing this, I was able to reproduce another PHP notice on this screen. We have a helper function to output any pull error messages and this only accounted for external connections, so when using an internal connection, a PHP notice is shown. And finally, it's possible for internal and external connections to have the same ID and the way we were storing these error messages wasn't accounting for that. So I've made a change so transient key we use now is different for both internal and external connections, to avoid any potential overlap.

### Alternate Designs

None

### Benefits

No more PHP notices will be shown and this fixes a few additional bugs.

### Possible Drawbacks

None

### Verification Process

1. Add this filter to your current theme: `add_filter( 'dt_available_pull_post_types', '__return_empty_array' );`
2. Go to the Pull screen and ensure that 1) no PHP notice is shown and 2) no posts are shown as available for pulling
3. Remove that filter and go back to the Pull screen and ensure posts are now showing

### Checklist:

### Applicable Issues

Fixes #402

### DIFF
--- a/includes/classes/PullListTable.php
+++ b/includes/classes/PullListTable.php
@@ -358,7 +358,7 @@ class PullListTable extends \WP_List_Table {
 	public function prepare_items() {
 		global $connection_now;
 
-		if ( empty( $connection_now ) ) {
+		if ( empty( $connection_now ) || empty( $connection_now->pull_post_type ) ) {
 			return;
 		}
 
--- a/includes/pull-ui.php
+++ b/includes/pull-ui.php
@@ -221,6 +221,7 @@ function process_actions() {
 			if ( 'external' === $_GET['connection_type'] ) {
 				$connection = \Distributor\ExternalConnection::instantiate( intval( $_GET['connection_id'] ) );
 				$new_posts  = $connection->pull( $posts );
+				$error_key  = "external_{$connection->id}";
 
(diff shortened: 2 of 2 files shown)


dinhtungdu (approved):
LGTM

PR #716, opened 2021-02-10
Reset the `last_changed_sites` option when a new site is created

### LINKED ISSUE #639: Doesn't initialize properly on sites created with WP CLI
**Describe the bug**

The multisite connection doesn't load on sites created with the WP CLI

**Steps to Reproduce**

1. Create WP Multisite Install
2. Install and Network Activate Distributor Plugin.
3. Run `$ wp site create --slug=example --title=Example`
4. Navigate to http://example.example.com/wp-admin/admin.php?page=pull and see the error: "No connections to Pull from, Create One?".

Note: Going to the 'Edit Site' page and clicking 'Save Changes' solves this problem.

**Expected behavior**

Should work the same as when sites are created using the web forms.

**Screenshots**

Error page
image

**Environment information**
 - Device: Virtual Machine: Homestead v9.5.1 on Virtualbox 6.1.12
 - OS: Ubuntu
 - Browser and version: Firefox
 - WordPress version: 5.5.1
 - WP CLI version: 2.4.0

dkotter:
### Description of the Change

We currently have some caching around our internal sites list that should be updated anytime a new site is created. An issue was reported where if a site was created using WP-CLI, no connections would show until that site was manually saved in the network settings. This seems to point to a caching issue.

Looking through the code that fires when a new site is created via WP-CLI, there is a possibility that the current way we update our cache may not fire correctly, under certain circumstances.

This PR adds a new hook onto `wp_insert_site` that fires our same cache clearing function we currently use that is hooked to `clean_site_cache`. In theory, this second hook _should_ always fire, while the first one may not fire in certain circumstances. For most circumstances though, both of these hooks will probably end up firing.

### Alternate Designs

None

### Benefits

The internal connections list should be more accurate now when a new site is created

### Possible Drawbacks

For most setups, I think we'll end up with both hooks firing, which isn't a huge deal but we'll basically clear the cache twice.

### Verification Process

Within an existing multisite, create a new site, go to that site and go to the Pull screen. Ensure connections show on that screen. Run this test both with manually creating a site and using WP-CLI.

Can also test from the perspective of an existing site. So create a new site, go to an existing site, go to the Pull screen and ensure the newly created site shows in the list. 

### Checklist:

### Applicable Issues

#639

### DIFF
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -496,6 +496,7 @@ class NetworkSiteConnection extends Connection {
 		add_action( 'wp_trash_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'separate_syndicated_on_delete' ) );
 		add_action( 'untrash_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'connect_syndicated_on_untrash' ) );
 		add_action( 'clean_site_cache', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'set_sites_last_changed_time' ) );
+		add_action( 'wp_insert_site', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'set_sites_last_changed_time' ) );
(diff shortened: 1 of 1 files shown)


dinhtungdu (approved):
LGTM 👍🏼

Issue #610, opened 2020-07-09
Bulk Skip function erroneously pulls posts


jeffpaul:
**Describe the bug**
In working to update the screenshots for the [Distributor microsite](https://distributorplugin.com/) today I ran into and issue when pulling content across External Connections and using bulk skip it instead pulls the content.  Individual row links for skip work correctly though.  A workaround for now would be to delete the mistakenly pulled post.

**Steps to Reproduce**

1. Go to Distributor > Pull Content
2. Select multiple posts
3. Change bulk dropdown to "Skip"
4. Click "Apply"
5. See that posts now appear in "Pulled" section as well as locally referenced posts in the All Posts list table.

**Expected behavior**

5. See that posts now appear in "Skipped" section.

**Screenshots**

**Environment information**
 - Device: Mac mini (2018)
 - OS: macOS 10.15.5
 - Browser and version: Firefox 78.0.1
 - WordPress version: 5.4.2
 - <details><summary>Site Health Info:</summary>
```
### wp-core ###

version: 5.4.2
site_language: en_US
user_language: en_US
timezone: +00:00
permalink: /%postname%/
https_status: false
user_registration: 0
default_comment_status: open
multisite: false
user_count: 3
dotorg_communication: true

### wp-paths-sizes ###

wordpress_path: /Users/jeffpaul/Local Sites/distributor-wpe/app/public
wordpress_size: 39.45 MB (41366248 bytes)
uploads_path: /Users/jeffpaul/Local Sites/distributor-wpe/app/public/wp-content/uploads
uploads_size: 56.27 KB (57625 bytes)
themes_path: /Users/jeffpaul/Local Sites/distributor-wpe/app/public/wp-content/themes
themes_size: 3.96 MB (4153984 bytes)
plugins_path: /Users/jeffpaul/Local Sites/distributor-wpe/app/public/wp-content/plugins
plugins_size: 6.35 MB (6660681 bytes)
database_size: 2.19 MB (2293760 bytes)
total_size: 52.01 MB (54532298 bytes)

### wp-active-theme ###

name: Twenty Twenty (twentytwenty)
version: 1.4
author: the WordPress team
author_website: https://wordpress.org/
parent_theme: none
theme_features: automatic-feed-links, custom-background, post-thumbnails, custom-logo, title-tag, html5, align-wide, responsive-embeds, customize-selective-refresh-widgets, editor-color-palette, editor-font-sizes, editor-styles, widgets, menus, editor-style
[...]

PR #717, opened 2021-02-12
Set empty action in bulk action dropdown

dhanendran:
### Description of the Change
We have the `Pull` action as the default action on the bulk dropdown, whenever we click the `Apply` button the page sends both top and bottom form to the server irrespective of the button we clicked. Since both `action` and `action2` has values, WP always takes the value of `action'. Now I have added a empty option as like the post list page to fix this issue. 

### Checklist:

### Applicable Issues

https://github.com/10up/distributor/issues/610

### DIFF
--- a/includes/classes/PullListTable.php
+++ b/includes/classes/PullListTable.php
@@ -527,6 +527,7 @@ class PullListTable extends \WP_List_Table {
 	public function get_bulk_actions() {
 		if ( empty( $_GET['status'] ) || 'new' === $_GET['status'] ) { // @codingStandardsIgnoreLine Nonce not required.
 			$actions = [
+				'-1'             => esc_html__( 'Bulk Action', 'distributor' ),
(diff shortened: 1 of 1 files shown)


includes/classes/PullListTable.php
```diff
@@ -527,6 +527,7 @@ public function column_cb( $post ) {
 	public function get_bulk_actions() {
 		if ( empty( $_GET['status'] ) || 'new' === $_GET['status'] ) { // @codingStandardsIgnoreLine Nonce not required.
 			$actions = [
+				'-1'             => esc_html__( 'Bulk Action', 'distributor' ),
```
dkotter: Minor change but I think the text here should be `Bulk actions` to match what WordPress uses elsewhere

dkotter:
@dhanendran Left one minor comment but otherwise, this looks good. Thanks!

dhanendran:
@dkotter I have addressed your comment.

Issue #392, opened 2019-06-05
Update Pull Content screen


jeffpaul:
**Is your enhancement related to a problem? Please describe.**
Making batch manipulation easier to understand and establish patterns for scale for setups with lots of connected sites.  Having `Skipped` items treated like new allows to pull things late which were accidentally skipped for some reason.

**Describe the solution you'd like**
From the designs below, ignore the changing header as we'll keep the existing header as-is.
- [x] `Pull` and `Skip` batch controls
- [x] include "All" option in post type dropdown to view all available content to pull across all available post types
- [x] ability to "un-skip" an item // include view of all `Skipped` posts and ability to `Unskip` a post to have it return to a state of being able to `Pull` or `Skip` again as well as direct option to `Pull` from the `Skipped` view.  This should be a row action as well as bulk action.

**Designs**
General screen flow:
1. https://cldup.com/s6qnqaojZg.png
1. https://cldup.com/-rGHbjj8L1.png
1. https://cldup.com/tCPn2kZ3Wm.png

**Describe alternatives you've considered**

**Additional context**
Relates to #40.

Props GH:@oszkarnagy / WP:[@oszkarnagy](https://profiles.wordpress.org/oszkarnagy/) for the UX analysis and recommendation.

mrhasbean:
Any update on this functionality? Very challenging at present with staff profiles missing on sites because they were accidentally skipped and no way of overriding that at present.

jeffpaul:
@mrhasbean this work isn't planned for a couple releases in the future.  Have you tried going to the source site and pushing them to the site(s) that are missing that content?

mrhasbean:
Hi Jeffrey,

Thanks for the reply. The source site is the main (parent) site of the multisite installation and I don’t get a “Push” option there, otherwise I would use that. Was wondering why I don’t see that option?

Cheers,
Steve

jeffpaul:
@mrhasbean the push menu is available in the admin bar for any published content ([screenshot](https://camo.githubusercontent.com/0b29e5649392ddffaee975ff58dad393e95d080e/68747470733a2f2f6469737472696275746f72706c7567696e2e636f6d2f77702d636f6e74656e742f7468656d65732f6469737472696275746f722d7468656d652f6173736574732f696d672f707573682d73637265656e73686f742e6a7067)).

emisrw:
[...]

PR #725, opened 2021-02-23
Introduce 'View all' post type filter on pull screen

elliott-stocks:
### Description of the Change

Introduce a 'View all' filter on the pull screen to allow users to view all post types by default instead of filtering manually. Also added a new column to the pull screen table to show which post type is currently in view - useful for when all are being displayed.

This handles one of the planned changes outlined in #392.

### Alternate Designs
N/A

### Benefits
Allows users to view all available post types on the pull screen without having to filter separately.

### Verification Process

1. Visit the pull screen and verify that "View all" is selected by default - all post types should be viewable
2. Test filtering various post types
3. Ensure correct post type is visible within the added "Post Type" column

### Checklist:

- [x]  My change requires a change to the documentation.

### Applicable Issues

#392

### DIFF
--- a/includes/classes/PullListTable.php
+++ b/includes/classes/PullListTable.php
@@ -422,10 +422,15 @@ class PullListTable extends \WP_List_Table {
 
 		$current_page = $this->get_pagenum();
 
-		if ( empty( $connection_now->pull_post_type ) || 'all' === $connection_now->pull_post_type ) {
-			$post_type = wp_list_pluck( $connection_now->pull_post_types, 'slug' );
+		// Support 'View all' filtering for internal connections.
+		if ( is_a( $connection_now, '\Distributor\InternalConnections\NetworkSiteConnection' ) ) {
+			if ( empty( $connection_now->pull_post_type ) || 'all' === $connection_now->pull_post_type ) {
+				$post_type = wp_list_pluck( $connection_now->pull_post_types, 'slug' );
+			} else {
+				$post_type = $connection_now->pull_post_type;
+			}
 		} else {
-			$post_type = $connection_now->pull_post_type;
+			$post_type = $connection_now->pull_post_type ? $connection_now->pull_post_type : 'post';
 		}
 
 		$remote_get_args = [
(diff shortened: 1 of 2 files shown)


includes/classes/PullListTable.php
```diff
@@ -258,7 +259,7 @@ public function column_date( $post ) {
 	public function column_default( $item, $column_name ) {
 		switch ( $column_name ) {
 			case 'name':
-				return $item['post_title'];
+				return $item->post_title;
```
dkotter: Looking at this, doesn't seem we even use this output, as the `name` column gets rendered in the `_column_name` callback function. Might be worth just removing this case.

includes/classes/PullListTable.php
```diff
@@ -258,7 +259,7 @@ public function column_date( $post ) {
...
 			case 'url':
```
dkotter: I know this is unrelated to any changes here but also doesn't seem we ever use a `url` column and if so, probably worth removing this case statement here as we're modifying other things in this UI

dkotter:
@elliott-stocks Overall looks good, left a few minor comments.

But there is an issue in regards to external connections. If you select `all` for the post type value (which is now the default) internal connections handle this fine but external connections break, because we now pass an array of post types and it expects a single post type.

As an example, here's one place that will throw a PHP error when `all` is selected and we try and pull from an external connection: https://github.com/10up/distributor/blob/develop/includes/classes/ExternalConnections/WordPressExternalConnection.php#L171

elliott-stocks:
> @elliott-stocks Overall looks good, left a few minor comments.
> 
> But there is an issue in regards to external connections. If you select `all` for the post type value (which is now the default) internal connections handle this fine but external connections break, because we now pass an array of post types and it expects a single post type.
> 
> As an example, here's one place that will throw a PHP error when `all` is selected and we try and pull from an external connection: https://github.com/10up/distributor/blob/develop/includes/classes/ExternalConnections/WordPressExternalConnection.php#L171

Thanks, @dkotter !

Do we want to enable this for external connections? I think it works well for internal but not too sure about external - happy to disable the "view all" for external if you agree.

dkotter:
> Do we want to enable this for external connections?

I guess my expectation would be that this would work on both. It would be a little weird from a user perspective to lose options on that Pull screen if you switch connection types.

But that said, because of the way we handle running a query on an external connection, not sure there's a straight forward way to make this all work, at least without having to refactor that query portion. I guess in theory we could loop through the post types and run individual queries for each of those and then combine and reduce those results to the most recent. There are performance issues to consider there though, especially for sites that have lots of post types.

includes/classes/PullListTable.php
jeffpaul: I agree that even some light cleanup is worthwhile while we're working on this section of the code.

jeffpaul:
@dkotter it could be worthwhile to add just for Internal Connections now and once we get through the larger work on https://github.com/10up/distributor/issues/123 we could come back to enable this for External Connections then?

dkotter:
> @dkotter it could be worthwhile to add just for Internal Connections now and once we get through the larger work on #123 we could come back to enable this for External Connections then?

If we just want to add it for Internal Connections now, that's fine with me. I do have concerns both about the user expectations that both connection types would behave the same and it does make the code a little messier, as I think we would need to have conditionals in a few places to change behavior depending on the connection type.

But I'm happy to review any changes to this PR if we want to just apply this to Internal Connections for now.

includes/classes/PullListTable.php
elliott-stocks: @dkotter Updated.

includes/classes/PullListTable.php
elliott-stocks: @dkotter Updated.

elliott-stocks:
Thanks, both!

@dkotter I've pushed updates to address the default column logic and to limit the new filter to internal connections.

dkotter (approved):

PR #726, opened 2021-02-26
Added support for core auto-updates for registered sites

### LINKED ISSUE #656: Add support for core auto-updates for registered sites
Now that WP core supports enabling auto-updates for plugins, we should look to support that functionality for registered Distributor sites.

**Designs**
Here's a screenshot from a sample install showing enable/disable states for plugins on WP.org on an unregistered Distributor site.  We'll want to ensure the yellow notice box extends fully to the right column.
Screen Shot 2020-10-20 at 12 40 56 PM

Otherwise for registered sites we'll want to allow for the enable/disable states in that far right column.

n/a

**Additional context**
n/a

dhanendran:
### Description of the Change

Added support for plugin auto-update for registered user.

### Possible Drawbacks

After updating to this version, this change might not be visible immediately as WP has already collected the data for existing plugins and stored it in transient for 12 hours (I guess). So either we need to manually delete the transient or wait for 12 hours.

### Verification Process

- After checking out to this branch, manually delete the transient.
- `_site_transient_update_plugins` meta key name in sitemeta table needs to be deleted

### Checklist:

### Applicable Issues

https://github.com/10up/distributor/issues/656


includes/settings.php
```diff
@@ -454,3 +456,91 @@ function sanitize_settings( $settings ) {
...
+ *
+ * @param stdClass $transient Transient cache object.
+ *
+ * @return stdClass
+ */
+function auto_update_plugins_setting( $transient ) {
+	if ( ! DT_IS_NETWORK ) {
+		$valid_license = \Distributor\Utils\get_settings()['valid_license'];
+	} else {
+		$valid_license = \Distributor\Utils\get_network_settings()['valid_license'];
```
dkotter: For both this line and the line above, we don't need the `\Distributor` namespace added, as we have a call at the beginning of this file to `use Distributor\Utils`. So both of those lines can be simplified to just `Utils\get_settings()` and `Utils\get_network_settings()`

includes/settings.php
```diff
@@ -454,3 +456,91 @@ function sanitize_settings( $settings ) {
...
+	}
+
+	if ( ! $valid_license ) {
+		return $transient;
+	}
+
+	$current_version = defined( 'DT_VERSION' ) ? DT_VERSION : null;
+	$update = check_for_updates( $current_version );
```
dkotter: Minor comment but the equals sign on this line should line up with the one above it

includes/settings.php
```diff
@@ -454,3 +456,91 @@ function sanitize_settings( $settings ) {
...
+
+/**
+ * Compare latest release version from plugin with the current installed version.
+ *
+ * @param string $current_version Current installed version.
+ *
+ * @return false|object
+ */
+function check_for_updates( $current_version ) {
+	$request = wp_safe_remote_get( 'https://api.github.com/repos/10up/distributor/releases?per_page=1' );
```
dkotter: In other places we make remote requests, we make if we are on VIP first in order to use the right function. We probably want to do that here as well. So something like:
```
if ( function_exists( 'vip_safe_wp_remote_get' ) && Utils\is_vip_com() ) {
    vip_safe_wp_remote_get(...);
} else {
    wp_safe_remote_get(...);
}

File: includes/settings.php [hunk 4]
```diff
@@ -454,3 +456,91 @@ function sanitize_settings( $settings ) {
...
+
+	if ( is_wp_error( $request ) ) {
+		return false;
+	}
+
+	$request = wp_remote_retrieve_body( $request );
```
dkotter: If we wanted to be extra careful here, we could check the response code to ensure that is a 200 response, using `wp_remote_retrieve_response_code`

includes/settings.php
```diff
@@ -454,3 +456,91 @@ function sanitize_settings( $settings ) {
...
+	$request = json_decode( $request );
+	$release  = ( $request[0] )? $request[0] : false;
```
dkotter: Some minor formatting issues here. Equals sign should be aligned to the one above it and there should be a space before the `?`

includes/settings.php
```diff
@@ -454,3 +456,91 @@ function sanitize_settings( $settings ) {
...
+}
+
+/**
+ * Compare latest release version from plugin with the current installed version.
+ *
+ * @param string $current_version Current installed version.
+ *
+ * @return false|object
+ */
+function check_for_updates( $current_version ) {
```
dkotter: So I know we bundle the Plugin Update Checker and use that to check for normal updates. Curious to know if we could utilize that here instead of having to maintain our own function? I think it ends up doing mostly the same as this function does, as far as the remote request goes, so may not provide any value but would keep things inline to how we handle normal updates.

dkotter (changes requested):

dhanendran:
@dkotter using an update checker library is a great suggestion. I have updated my code to use them now.

includes/settings.php
```diff
@@ -454,3 +455,24 @@ function sanitize_settings( $settings ) {
...
+ */
+function auto_update_plugin() {
+	if ( ! DT_IS_NETWORK ) {
+		$valid_license = Utils\get_settings()['valid_license'];
+	} else {
+		$valid_license = Utils\get_network_settings()['valid_license'];
+	}
+
+	if ( $valid_license ) {
+		\Puc_v4_Factory::buildUpdateChecker(
```
dkotter: This class should always exist but we may want to add a `class_exists( 'Puc_v4_Factory' )` check here, as we do that same thing in other places

dkotter (changes requested):

dkotter:
@dhanendran This looks a lot cleaner now, using that update class. I've left one minor comment but otherwise looks good to me code-wise. Have you tested this with the new approach and are seeing auto-updates work? I tried testing locally but I could never get auto-updates to fire, though may have just not waited long enough.

dhanendran:
@dkotter looks like similar code already present in https://github.com/10up/distributor/blob/develop/distributor.php#L156-L188. Not sure why it is not working though, I will check the existing implementation first.

jeffpaul:
@dhanendran any updates on your research on this topic?

dhanendran:
@jeffpaul unfortunately no, as I was busy with other project priorities. I will try to check this in next couple of weeks.

sksaju:
This PR is ready for your review @dkotter @jeffpaul

Screenshot 2022-05-25 at 4 38 39 AM

dkotter (approved):

Issue #583, opened 2020-05-21
UI bug: switching set up process doesn't clear fields


jeffpaul:
**Describe the bug**
From @helen's [comment](https://github.com/10up/distributor/pull/368#issuecomment-612242826) on #368:
UI bug - if I click "Manually Setup Connection" but then decide to switch to WordPress.com Application, the fields do not disappear. [See screenshot](https://share.getcloudapp.com/bLueYZNR) for what I mean. I think this should probably reset entirely, so hide the fields when you switch, but if you switch _back_ to `Username/Password` it should give you the wizard thing again.

**Steps to Reproduce**

**Expected behavior**

**Screenshots**

**Environment information**
 - Device: 
 - OS: 
 - Browser and version: 
 - Distributor version: 
 - Theme and version: 
 - Other installed plugin(s) and version(s):

**Additional context**

jeffpaul:
Punting to 1.6.1 as something to resolve in a minor release, happy to get a PR reviewed and merged if someone's able to work on this in the meantime.

dhanendran:
@jeffpaul I can work on this issue.

dhanendran:
PR: https://github.com/10up/distributor/pull/727

PR #727, opened 2021-03-01
Hide manual connection when switching to WP.COM

dhanendran:
### Description of the Change

Hiding manual connection setup fields when switching to wp.com connection mode.

### Verification Process

- Checkout to this branch
- Run `npm run watch`
- Create new external connection
- Try to change the connection type

### Checklist:

### Applicable Issues

https://github.com/10up/distributor/issues/583

### DIFF
--- a/assets/js/admin-external-connection.js
+++ b/assets/js/admin-external-connection.js
@@ -451,6 +451,12 @@ jQuery( externalConnectionTypeField ).on( 'change', () => {
 	const slug = externalConnectionTypeField.value;
 
 	wpbody.className = slug;
+
+	if ( 'wp' === slug ) {
+		jQuery( '.external-connection-setup, .hide-until-authed' ).show();
+	} else {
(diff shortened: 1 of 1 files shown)


dkotter:
@dhanendran @jeffpaul Code looks good here, thanks!

Testing things out, this addresses this point:
> if I click "Manually Setup Connection" but then decide to switch to WordPress.com Application, the fields do not disappear

But also mentioned in that same issue:

> I think this should probably reset entirely, so hide the fields when you switch, but if you switch back to Username/Password it should give you the wizard thing again.

The fields are hidden when you switch but if you switch back to `Username/Password`, it doesn't give the wizard prompt again. Not sure if we want to address that here or in a separate PR?

jeffpaul:
If @dhanendran you're able to handle the second part of Darin's comment I'd be happy to have that roll in with this PR change?

dhanendran:
@dkotter I have addressed that concern now. Can you review this again?

dkotter (approved):

PR #728, opened 2021-03-02
Add ability to Unskip or Pull items from the Skipped tab on the Pull screen

dkotter:
### Description of the Change

At the moment, if you Skip an item from the Pull screen, that item goes to a Skipped tab. From here, the only option a user has is to view that item. So if someone were to accidentally Skip an item or change their mind later, there isn't an easy way for them to later pull that item.

This PR adds the same Pull options to the Skipped tab, both in the bulk actions and in the inline row options. This allows someone to directly Pull an item from the Skipped tab.

In addition, it adds a new Unskip option, both in the bulk actions and in an inline row option. If this Unskip option is used, the Skipped item is sent back to the New screen and can then be Pulled from there later (or Skipped again if desired).

Note: by default, we show the 10 most recent items on the new screen. If someone were to Unskip an item that doesn't belong to one of those 10 items, this item will still be Unskipped but won't show up on the first page of results. Pagination will have to be used to find this item.

### Alternate Designs

None

### Benefits

Users can now Unskip or directly Pull previously Skipped items

### Possible Drawbacks

None

### Verification Process

1. Go to the Pull screen and from the New tab, Skip an item
2. Go to the Skipped tab and find this Skipped item
3. Test using the inline row option and Unskip this item
4. Verify that item now shows back up in the New tab
5. Repeat using the Bulk option
6. In addition, test if the Pull option works as expected from the Skipped tab following the same directions as above but using the Pull option instead of the Unskip option

### Checklist:

### Applicable Issues

Addresses part of #392

### DIFF
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -346,12 +346,7 @@ class NetworkSiteConnection extends Connection {
 	public function log_sync( array $item_id_mappings, $blog_id = 0, $overwrite = false ) {
 		$blog_id = 0 === $blog_id ? $this->site->blog_id : $blog_id;
 
-		$sync_log = get_option( 'dt_sync_log', array() );
-
-		$current_site_log = [];
-		if ( ! empty( $sync_log[ $blog_id ] ) && false === $overwrite ) {
-			$current_site_log = $sync_log[ $blog_id ];
-		}
+		$current_site_log = $this->get_sync_log( $blog_id );
 
 		foreach ( $item_id_mappings as $old_item_id => $new_item_id ) {
 			if ( empty( $new_item_id ) || is_wp_error( $new_item_id ) ) {
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -338,17 +338,18 @@ class NetworkSiteConnection extends Connection {
 	 *
 	 * This let's us grab all the IDs of posts we've PULLED from a given site
 	 *
-	 * @param array $item_id_mappings Mapping to log; key = origin post ID, value = new post ID.
-	 * @param int   $blog_id Blog ID
+	 * @param array   $item_id_mappings Mapping to log; key = origin post ID, value = new post ID.
+	 * @param int     $blog_id Blog ID
+	 * @param boolean $overwrite Whether to overwrite the sync log for this site. Default false.
 	 * @since 0.8
 	 */
-	public function log_sync( array $item_id_mappings, $blog_id = 0 ) {
+	public function log_sync( array $item_id_mappings, $blog_id = 0, $overwrite = false ) {
 		$blog_id = 0 === $blog_id ? $this->site->blog_id : $blog_id;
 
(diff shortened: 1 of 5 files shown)


includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -338,19 +338,15 @@ public function pull( $items ) {
...
 	 *
-	 * @param array $item_id_mappings Mapping to log; key = origin post ID, value = new post ID.
-	 * @param int   $blog_id Blog ID
+	 * @param array   $item_id_mappings Mapping to log; key = origin post ID, value = new post ID.
+	 * @param int     $blog_id Blog ID
+	 * @param boolean $overwrite Whether to overwrite the sync log for this site. Default false.
 	 * @since 0.8
 	 */
-	public function log_sync( array $item_id_mappings, $blog_id = 0 ) {
+	public function log_sync( array $item_id_mappings, $blog_id = 0, $overwrite = false ) {
```
elliott-stocks, line 346: @jeffpaul I think the logic for `$overwrite` is missing from this function. Currently the post ID is set to false which means it still shows on the skipped screen.

If an empty array is passed to `$item_id_mappings` it will fallback to `$current_site_log` instead, but it should overwrite.

elliott-stocks (changes requested):

includes/classes/InternalConnections/NetworkSiteConnection.php
dkotter: @elliott-stocks Ah, nice catch. I had refactored this a bit and in the process I ended up breaking things. I believe it should be working properly now

includes/pull-ui.php
```diff
@@ -326,6 +330,52 @@ function( $remote_post_id ) use ( $post_type, $post_status ) {
...
+
+			foreach ( $posts as $post_id ) {
+				if ( array_key_exists( $post_id, $sync_log ) ) {
+					unset( $sync_log[ $post_id ] );
+				}
+			}
+
+			$connection->log_sync( $sync_log, intval( $_GET['connection_id'] ), true );
+
+			setcookie( 'dt-unskipped', 1, time() + DAY_IN_SECONDS, ADMIN_COOKIE_PATH, COOKIE_DOMAIN, is_ssl() );
```
elliott-stocks, line 374: @dkotter VIP will flag this due to caching, although as it's in the admin it may not be an issue. Testing locally on VIP seems fine, but it may fail PHPCS.

includes/classes/InternalConnections/NetworkSiteConnection.php
elliott-stocks: Looks good!

elliott-stocks (approved):

elliott-stocks:
@dkotter @jeffpaul This works well locally - approved.

Added a note about the cookie logic as this will be flagged by VIP, but as it's in the admin I think it _should_ be fine.

includes/pull-ui.php
dkotter: Yeah, I debated changing how this works but this is how we already handle skipped or pulled items, by setting a cookie. So I decided it was best to just follow the pattern we already have in place.

dinhtungdu (approved):
Tested and this is looking good to me 👍🏼

Issue #735, opened 2021-03-10
Test against WordPress 5.7


jeffpaul:
**Is your enhancement related to a problem? Please describe.**
Once WordPress 5.7 is released, we'll want to test Distributor to see if any incompatibility issues arise.

**Describe the solution you'd like**
- [x] test Distributor on WordPress 5.7
- [ ] open issues for any incompatibilities noted in testing
- [ ] resolve issues identified in testing
- [x] bump "tested up to" version
- [ ] if code changes needed due to incompatibilities, ship a plugin release

**Designs**
n/a

**Describe alternatives you've considered**
none

**Additional context**
Related: https://github.com/10up/distributor/issues/568, https://github.com/10up/distributor/issues/624.

antoinebougr:
Hi

I m facing a fatal error after wordpress 5.7 update. For the moment I disabled distributor to make wordpress work again
![distributor](https://user-images.githubusercontent.com/80766046/111318658-8b311280-8665-11eb-9d1b-a2d6b52a3edc.png)

dkotter:
@antoinebougr Thanks for the report. This doesn't seem to be related to the latest WordPress update but does seem like this may be related to the latest Distributor release. I have not been able to reproduce this error on either WP 5.6 or 5.7, with the latest release of Distributor. Can you let me know what version of PHP you are using? Based on the error you shared, seems like it may be a compatibility issue with a certain version of PHP.

PaddyWhacks:
Just updated Multisite to WordPress 5.7 and distributor to 1.6.3. Rolling back to 1.6.2 works. Same error message as @antoinebougr - 

Fatal error: Declaration of Distributor\ExternalConnection::log_sync() must be compatible with Distributor\Connection::log_sync(array $item_id_mappings, $id, Distributor\bool $overwrite) in /home/dotnewsb/public_html/wp-content/plugins/distributor/includes/classes/ExternalConnection.php on line 17

There has been a critical error on this website. Please check your site admin email inbox for instructions.

Learn more about troubleshooting WordPress.

dkotter:
@PaddyWhacks Do you happen to know what version of PHP you're site is running? Thanks!

dkotter:
@PaddyWhacks @antoinebougr Just tested on various PHP versions and found the issue. With the latest release of Distributor and a site running PHP <7.2 (so 7.1, 7.0 or 5.6) a fatal error occurs. I just fixed this in PR #742, if you want to test that out.

[...]

PR #742, opened 2021-03-19
Fix a PHP fatal error with the `log_sync` function

### LINKED ISSUE #734: Site crashed on the 1.6.3 version
**Describe the bug**

Site crashed on the 1.6.3 version

**Steps to Reproduce**

**Expected behavior**

**Screenshots**

**Environment information**
 - Device: 
 - OS: 
 - Browser and version: 
 - WordPress version: 
 
 - Plugins and version: 
 - Theme and version: 
 
 - <details><summary>Site Health Info:</summary>
   
   </details>

**Additional context**

dkotter:
### Description of the Change

In the latest Distributor release, we added an optional parameter to the `log_sync` method that is used to overwrite the log. The base method we extended had a type declaration on that optional value to `bool` but the two places we extended this method didn't have that same type declaration. In PHP versions 7.0 and 7.1, this will cause a fatal error to be thrown because those function signatures don't match.

In addition, PHP 5.6 does not support scalar type declarations, which `bool` is considered a scalar type. Because we still support PHP 5.6, this PR removes the `bool` type declaration so no errors will happen in 5.6, 7.0 or 7.1.

### Alternate Designs

None

### Benefits

No PHP fatals on PHP version <7.2

### Possible Drawbacks

None

### Verification Process

- Install the latest version of Distributor on a site running PHP <7.2
- Should get a fatal error
- Checkout this PR in same environment
- Everything should work now

### Checklist:

### Applicable Issues

Fixes #734. Related #735

### DIFF
--- a/includes/classes/Connection.php
+++ b/includes/classes/Connection.php
@@ -48,7 +48,7 @@ abstract class Connection {
 	 * @param boolean $overwrite Whether to overwrite the sync log. Optional.
 	 * @since 0.8
 	 */
-	abstract public function log_sync( array $item_id_mappings, $id, bool $overwrite );
+	abstract public function log_sync( array $item_id_mappings, $id, $overwrite );
 
 	/**
 	 * Get the sync log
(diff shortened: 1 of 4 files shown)


helen (approved):
LGTM

Issue #141, opened 2018-06-26
Gutenberg: Extra Refresh to Push Content In Gutenberg


dkotter:
By default, we only allow published content to be Distributed. In the Classic Editor, when you publish an item, the page refreshes and then the admin menu item is added by Distributor, which then allows us to push that item to other connections.

In Gutenberg, there are no page refreshes, so if you publish an item, the item is published but the page doesn't refresh, so the custom menu item is never added. This means you have to manually refresh the page before you can push that item anywhere. Not a huge deal but also not the most intuitive approach. Not sure the best way around this but would be nice, if possible, to listen for the post publish event and then add the dropdown menu.

helen:
I think for the new editor we should probably explore using a side panel instead of the admin bar. Generally we probably need to explore other push UI besides/in addition to the admin bar because it's possible to have it off or not have enough space especially on smaller screens.

helen:
This belongs in a more major 1.7-type of milestone, can call it good enough for a point release in #750 but want to get into this and remove a lot of that hover code for the next release.

peterwilsoncc:
@jeffpaul Do you recall why you added the help wanted tag on this? It looks like it's fixed by the merged PRs #750 and #790.

dkotter:
@peterwilsoncc I can't speak to the tag but there is still an issue here. Those referenced PRs definitely make the integration of Distributor and the block editor nicer but the don't fully solve the issue I opened here.

To recap, by default we only allow published items to be distributed so the distribution menu isn't even loaded if an item is in a non-published state. When an item is published in the block editor, because this doesn't refresh the page the menu is never shown.

We do show this now (coming from those referenced PRs), which lets the user know they need to refresh themselves: 
<img width="289" alt="Distributor refresh message" src="https://user-images.githubusercontent.com/916738/174126589-4b213393-3418-46ee-a7ce-fae1bbcfcdc8.png">

But I think there's still opportunity here for improvement, whether we lazy load in the menu after publish, whether we load it in at the beginning and only show it after publish, some other approach?

Also, and I didn't test this but I wonder if we have the same issue in reverse, for an item that is published and then gets put back in draft. I believe the distribution menu is probably still active until the page refreshes, though I'd have to test to see if it allows you to still distribute that item or not.

peterwilsoncc:
Thanks for clarifying Darin. 

For 1.7.0 I think I will keep the existing PRs as the bug fix and aim to remove the refresh in either 2.0.0 or 2.1.0

PR #750, opened 2021-04-08
Add better support for the block editor's fullscreen mode

### LINKED ISSUE #597: Distributor admin bar menu unavailable in Gutenberg fullscreen mode (current WP default)
**Describe the bug**
In testing something unrelated, I noticed that WP 5.4 and the new default of fullscreen mode for Gutenberg that post-publish you don’t get the admin bar anymore so the Distributor menu is no longer visible in Gutenberg when it used to be.

**Steps to Reproduce**

1. Go to Gutenberg editor of a Post
2. Publish a post
3. See that admin bar is no longer present
4. :sad-panda:

**Expected behavior**
I know we want to avoid cluttering up the sidebar with more cognitive load and maintaining 2 UIs, so some options to consider are:
- integrating Distributor push actions into the Gutenberg slotfills versus relying on the admin bar
- something in the Gutenberg view to trigger showing the existing UI
- ...something else I haven't considered?

**Screenshots**

**Environment information**
 - Device:  Mac mini (2018)
 - OS:  MacOS 10.15.5
 - Browser and version:  Firefox 77.0.1
 - WordPress version: 5.4.2
 - Distributor version:  1.5.0
 - Theme and version:  Twenty Twenty 1.4
 - Other installed plugin(s) and version(s):

**Additional context**

Back in August 2019 a [Gutenberg issue was opened](https://github.com/WordPress/gutenberg/issues/16988) looking to add a SlotFill to the Toolbar in the Gutenberg header.  Later that month our very own @ryanwelcher [pushed up a PR](https://github.com/WordPress/gutenberg/pull/17198) to try and solve that issue, but it appears the Gutenberg team either doesn't appreciate the issue at hand or don't agree with the approach of the PR as there's no clear direction from that team on how to resolve the linked issue (either by iterating on the PR or via a completely separate approach).

Then in April of this year, a [similar issue was opened](https://github.com/WordPress/gutenberg/issues/21739) which then referenced back to the earlier issue and PR from August 2019.  Then again in May of this year, a [related issue was opened](https://github.com/WordPress/gutenberg/issues/22178) focused more on missing the New/View Page options from the adminbar.

In reviewing Gutenberg PRs to see what other options may currently be explored to adjust the fullscreen mode all that seems to exist is a [merged (and then reverted) PR](https://github.com/WordPress/gutenberg/pull/21121) from March of this year that attempted to make the WP logo toggle the WP Admin sidebar view on/off from fullscreen mode.  A [second PR attempt](https://github.com/WordPress/gutenberg/pull/22191) on this WP logo toggle from May of this year is still open and appears on hold and unlikely to be merged in time to make it into WP 5.5.

dkotter:
### Description of the Change

The default block editor mode is fullscreen, which hides the top admin bar. This admin bar is where Distributor puts the distribution tools. So when in fullscreen mode, it can be hard to figure out how to actually distribute something.

This PR attempts to solve that by adding a new Distributor panel to the editor with a toggle option. This panel will hold the current `Distributed to` and `Distributed on` messages that previously showed in the Post Status panel and will now show a new button to toggle the showing of the admin bar (if needed) and the showing of the Distributor menu content.

This option will show both in normal mode and in fullscreen mode on a published item that isn't a previously Distributed item (an item received from another site). On a non-published item, we output a simple message letting users know they need to publish first. On a newly published item, since the admin bar only updates on page refreshes, instead of showing the toggle we show a message letting users know they need to refresh.

Note: no copy or designs are final. Open to suggestions on all of that.

Screenshots of all scenarios will follow.

### Screenshots
Post received through Distribution, shows the same in both normal and fullscreen mode:
distributed-post-full-and-normal

Post that has been Distributed, in normal and fullscreen mode
Screen Shot 2021-04-09 at 10 42 40 AM

Post with no Distributions, in normal and fullscreen mode
Screen Shot 2021-04-09 at 10 52 14 AM

Non-published post
non-published-post

Newly published post, prior to a page refresh
Screen Shot 2021-04-09 at 10 45 56 AM

### Alternate Designs

Could attempt to move the full Distribution capabilities from the admin bar to this new panel.

### Benefits

When using the block editor in fullscreen mode, it should now be easier to figure out how to distribute something

### Possible Drawbacks

None

### Verification Process

1. Checkout this PR
2. Test the various scenarios:
   - Edit a published item that hasn't been distributed yet, checking both normal mode and fullscreen mode
   - Edit a published item that has been distributed, checking both normal mode and fullscreen mode
   - Edit an item that hasn't been published yet, checking both normal mode and fullscreen mode
   - Edit an item that was received through distribution, checking both normal mode and fullscreen mode
3. Ensure the new Distributor panel shows properly in all scenarios and toggling to the admin bar works properly in fullscreen mode

### Checklist:

### Applicable Issues

#597


assets/js/gutenberg-plugin.js
```diff
@@ -0,0 +1,154 @@
...
+			{ __( 'Distributed on: ', 'distributor' ) }
+			<strong> { dtGutenberg.syndicationTime } </strong>
+		</span>
+	);
+};
+
+/**
+ * Create the Distributor icon
+ */
+const DistributorIcon = () => {
```
dinhtungdu: We have an SVG file in the assets folder, I think we should reuse it instead of duplicating it here.

dkotter: This has been addressed. Had to test a few different webpack loaders to get this working but I believe I've landed on a working solution now.

dinhtungdu (dismissed):
LGTM 💯

jeffpaul:
@helen any further UX tweaks you'd like to see here before this gets merged in?

jeffpaul:
Noting this might also handle #141, so will want to test for that in reviewing this PR.

helen:
This is working for me and I think does at least address the immediate problem - the one thing I think it needs before merge is to close the popup as soon as you click outside of the area, otherwise it's not super intuitive how you can close it (either click the button again or else mouse into the area and back out) and you can end up clicking around fruitlessly wondering if it'll ever close, and it does let you click into blocks, etc.

dkotter:
@helen @dinhtungdu 

> close the popup as soon as you click outside of the area

Just pushed a fix for this

dinhtungdu:
@darin, @helen Clicking outside the area not only closes the popup but also selects the element under the mouse. For example when the popup is open after clicking on `Distribut Post`, click on the editor area selects the block then change the sidebar content, which may be confusing. The situation goes worse if `Move to Trash` button is under the cursor.

dkotter:
@dinhtungdu I've made some more changes here on how we handle the overlay. Previously this was just added as styles to the menu item. I've now added an actual overlay div that we show/hide when needed. This is then used to capture clicks so you don't end up clicking on any elements that show beneath that overlay.

This is a bigger change here though so would appreciate some testing. It seemed to work fine in my tests, both in the admin and in the front-end, and both with hovering over the menu, entering the menu via the keyboard and using the new sidebar button.

assets/js/gutenberg-plugin.js
```diff
@@ -0,0 +1,167 @@
+import { wp, dtGutenberg } from 'window';
+import PluginIcon from '../img/icon.svg'; // eslint-disable-line no-unused-vars
+
+const { Icon } = wp.components; // eslint-disable-line no-unused-vars
```
dinhtungdu, line 4: Look like we haven't added Eslint rules for react yet. I think it's better to do it in this PR  so we don't have to add these comments to disable Eslint.

dinhtungdu (dismissed):
This is working great in my test. But there are some minor issues with this PR:
- We don't show the push menu on mobile/small screens, so please considering hiding the sidebar button on small screens.
- It throws three errors in the console. It only happens with `feature/gutenberg-fullscreen-support`. I don't get any error with `develop`. 
Screen Shot 2021-08-06 at 11 16 41

dkotter:
@dinhtungdu 

> We don't show the push menu on mobile/small screens, so please considering hiding the sidebar button on small screens.

I've added some CSS to hide the Distributor panel on smaller screens

> It throws three errors in the console

I actually don't see any errors. Any other information you can provide around where the errors are happening?

assets/js/gutenberg-plugin.js
dkotter: @dinhtungdu Any recommendations on the best way to add these rules without impacting other things? We currently use an older version of the `@10up/eslint-config` and I updated that to the latest version (following instructions [here](https://github.com/10up/10up-toolkit/blob/develop/packages/eslint-config/README.md)), but I find that brings over a lot of pretty strict rules, some of which I don't agree with (which is fine, that's just my opinion).

But what this does mean is all the JS files will need a bunch of minor formatting changes applied in order to pass these new rules. Not sure we want all of that cleanup work on this PR, so wondering if there's either a simpler solution for now to add better React eslint rules or if we tackle this in a separate PR, one built around updating our entire build/lint process?

jeffpaul: I'm fine with (1) resolving those eslint failures in a separate PR and (2) not worrying about those until the next release (aka let's not block this issue/release by those eslint issues).

helen:
There are two things going on here that I would like to see ironed out:

* The admin toolbar remains visible after the dropdown closes, i.e. the `is-showing-distributor` class is not being removed
* The overlay shows up before the dropdown because of the hoverIntent stuff on the toolbar, which feels really weird and can also lead to the overlay showing up and then never being removed again if you mouse over the Distributor item without triggering the dropdown.

assets/js/gutenberg-plugin.js
dinhtungdu: > But what this does mean is all the JS files will need a bunch of minor formatting changes applied in order to pass these new rules.

I didn't think about this. Let's fix the eslint in other PR.

dinhtungdu:
> I actually don't see any errors. Any other information you can provide around where the errors are happening?

I tested on a multisite installation. Only Distributor is activated (network-wide). The WP version is 5.8. I'm using `apache2` with PHP `7.3.24`. Please check this screencast: https://share.getcloudapp.com/Jruxg2pn!

dkotter:
@helen 

> The admin toolbar remains visible after the dropdown closes, i.e. the is-showing-distributor class is not being removed

This should be fixed now

> The overlay shows up before the dropdown because of the hoverIntent stuff on the toolbar

I took another stab at this and I think it's better now, though still not perfect. As you mention, core uses `hoverintent` on the admin bar to determine when to show things. That has a timeout on it so you have to actually hover over something for a certain period of time before it triggers. This can result in us showing the overlay even though the dropdown isn't showing or cause a delay between the two.

I've added a delay on our code (if things aren't already showing) to try and deal with this. I feel it's definitely better but open to other ideas on how to solve this.

dkotter:
@dinhtungdu Thanks for the additional details. I have a very similar set up but still don't get those warnings. But looking at the  stack trace it gives, I think I found the issue and I've pushed a fix for that. Let me know if that clears those up for you or not.

helen (approved):
I hope we'll be able to come up with a better long-term solution someday where we can remove a lot of this because it's pretty funky, but it's working for what we need for the moment.

dkotter:
@helen I agree and after doing some more thinking, I did think of a few ways to simplify this logic. That is in a new PR, #790

PR #767, opened 2021-05-28
Fix: Post author uses API user instead of author

### LINKED ISSUE #751: Post author uses API user instead of author.
**Describe the bug**
We have selected the option in Setting: For linked distributed posts, replace the author name and link with the original site name and link.

but when a new post is distributed the Author is the API user and not the post author. 

**Steps to Reproduce**
1.create new post
2. distributed it

**Expected behavior**
created post to be distributed with the original Author

**Screenshots**

**Environment information**
 - Device: PC
 - OS: Windows 10
 - Browser and version: Chrome Version 89.0.4389.128 
 - WordPress version: 5.7.1

dinhtungdu:
### Description of the Change

Originally, when `override_author_byline` is false, Distributor uses the user of the remote site (which is used for authorizing the connection) as the post author.

This PR adds the ability to display the original author for distributed posts. This is enabled by default and can be turned off by a filter to restore the original behavior. I decided to make it default because using the original author makes more sense to me.

This PR also fixes the issue where the original site name and URL are `null` when pulling external posts.

### Alternate Designs
n/a

### Benefits
As mentioned above.

### Possible Drawbacks
This PR will cause inconsistent behavior for the old posts which don't have the original author data.

### Verification Process
1. Switch to this PR on both original and remote sites.
2. Make sure `override_author_byline` isn't checked on both sites.
3. On the remote site, create a post with a unique author, which is not the current logged-in user or has the same username as any user on the remote site.
4. Push the post to the remote site.
5. See the author name and URL of the distributed post on the frontend set to the original author.
6. Do the similar for pull, see the original author used on the front end.

### Checklist:

### Applicable Issues
Closes #751

### Changelog Entry

Fixed: prepare correct original site name and URL for pull action.
Added: show the original author as the author of distributed posts.

### DIFF
--- a/includes/rest-api.php
+++ b/includes/rest-api.php
@@ -265,7 +265,7 @@ function register_endpoints() {
 		array(
 			'get_callback'    => function( $post_array ) {
 				$site_url = get_post_meta( $post_array['id'], 'dt_original_site_url', true );
-				return empty( $site_url ) ? home_url() : esc_html( $site_url );
+				return empty( $site_url ) ? home_url() : esc_url_raw( $site_url );
 			},
 			'update_callback' => function( $value, $post ) { },
 			'schema'          => array(
--- a/includes/rest-api.php
+++ b/includes/rest-api.php
@@ -248,7 +248,8 @@ function register_endpoints() {
 		'distributor_original_site_name',
 		array(
 			'get_callback'    => function( $post_array ) {
-				return esc_html( get_post_meta( $post_array['id'], 'dt_original_site_name', true ) );
+				$site_name = get_post_meta( $post_array['id'], 'dt_original_site_name', true );
+				return empty( $site_name ) ? get_bloginfo( 'name' ) : esc_html( $site_name );
 			},
 			'update_callback' => function( $value, $post ) { },
 			'schema'          => array(
@@ -263,7 +264,8 @@ function register_endpoints() {
 		'distributor_original_site_url',
 		array(
 			'get_callback'    => function( $post_array ) {
-				return esc_url_raw( get_post_meta( $post_array['id'], 'dt_original_site_url', true ) );
+				$site_url = get_post_meta( $post_array['id'], 'dt_original_site_url', true );
+				return empty( $site_url ) ? home_url() : esc_html( $site_url );
 			},
 			'update_callback' => function( $value, $post ) { },
 			'schema'          => array(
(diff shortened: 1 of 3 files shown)


includes/rest-api.php
```diff
@@ -263,7 +264,8 @@ function register_endpoints() {
 		'distributor_original_site_url',
 		array(
 			'get_callback'    => function( $post_array ) {
-				return esc_url_raw( get_post_meta( $post_array['id'], 'dt_original_site_url', true ) );
+				$site_url = get_post_meta( $post_array['id'], 'dt_original_site_url', true );
+				return empty( $site_url ) ? home_url() : esc_html( $site_url );
```
dkotter: Seems like we should probably have the escaping here be `esc_url_raw` instead of `esc_html`

dinhtungdu: Yeah, it's my mistake. I fixed it in https://github.com/10up/distributor/pull/767/commits/1e49b581bfaa0701f0885a36a7f1d0eed96b5441

ldollard:
@dinhtungdu let me know what you need me to do, not really a programmer, but am willing to either insert the code or replace a file and do testing on my sites for this.

dinhtungdu:
@ldollard Thank you! To test this feature, please follow these steps:
1. Go to [this page](https://github.com/10up/distributor/tree/fix/751), click on the green `Code` button, choose `Download ZIP`.
2. Extract your downloaded zip file, copy the `includes` folder inside it.
3. Replace the `includes` folders of your installed Distributor plugin with the one from the zip file.
4. Test to see if your issue is fixed.

Thank you again for your help! Please test this on localhost or staging sites only!

ldollard:
Hi Tung Du,

I tested it from my master site to a remote site and it sort of works. 

See the screenshots:

https://we.tl/t-7GsPQeM2ig

If I pick an author in this case Terry, publish the post then distribute it, the post gets sent, but the author on the post in the remote site on the backend still shows as the distributor user that connects the sites, but on the front end it shows a mashup with the distributor user and the pushed author from the master site.

The avatar you see is the avatar of the distributor user not the author that got assigned.

thanks

dinhtungdu:
I'm closing this PR because this doesn't fully solve the issue of distributing the original post author. The author doesn't only have a name and URL but also can have Avatar, bio, and custom user metadata. The author boxes displayed on the front-end are also different from site to site, so we don't have a fixed set for the field to distribute. The approach in this PR is similar to Co-Authors Plus, which shouldn't be done in Distributor core IMO.

This PR can be used as an inspiration for custom development or even a Distributor addon.

Issue #773, opened 2021-07-01
Fatal error


Drmzindec:
When a site isnt a multi site and you want to edit an imported article it does a check in the below file and gets a Fatal error.

```
Fatal error: Uncaught Error: Call to undefined function Distributor\SyndicatedPostUI\switch_to_blog()
in /home/ubuntu/test/wp-content/plugins/distributor-develop/includes/syndicated-post-ui.php on line 541
Call stack:
    Distributor\S\enqueue_gutenberg_edit_scripts()
    wp-includes/class-wp-hook.php:292
    WP_Hook::apply_filters()
    wp-includes/class-wp-hook.php:316
    WP_Hook::do_action()
    wp-includes/plugin.php:484
    do_action()
    wp-admin/edit-form-blocks.php:409
    require()
    wp-admin/post.php:187
```

jeffpaul:
@Drmzindec looking at the file location for the plugin files it looks like you may have downloaded a ZIP of the `develop` branch for your use.  If that's the case, you're better off using the built version from the `stable` branch or the ZIP file referenced on a [specific release](https://github.com/10up/distributor/releases).

Drmzindec:
@jeffpaul I have used the latest stable branch and the error is the same:

```
Fatal error: Uncaught Error: Call to undefined function Distributor\SyndicatedPostUI\switch_to_blog()
in /home/ubuntu/test/wp-content/plugins/distributor-1.6.4/includes/syndicated-post-ui.php on line 541

Call stack:

    Distributor\S\enqueue_gutenberg_edit_scripts()
    wp-includes/class-wp-hook.php:292
    WP_Hook::apply_filters()
    wp-includes/class-wp-hook.php:316
    WP_Hook::do_action()
    wp-includes/plugin.php:484
    do_action()
    wp-admin/edit-form-blocks.php:409
    require()
    wp-admin/post.php:187
```

The lines in the file:

```
switch_to_blog( $original_blog_id );
$original_location_name = get_bloginfo( 'name' );
restore_current_blog();
```

The site i am running this on is not a multisite so the above line that produces the fatal error will never be able to run successfully.

However something like the below will solve the issue:

```
if (is_multisite()) {
    switch_to_blog($original_blog_id);
    $original_location_name = get_bloginfo('name');
    restore_current_blog();
}
```

dkotter:
@Drmzindec Thanks for the detailed report.

I tried to reproduce myself in a brand-new single site install and did not encounter the same problem.

[...]

PR #780, opened 2021-07-15
Ensure we are on a multisite before using `switch_to_blog`

dkotter:
### Description of the Change

As detailed in #773, if you use Distributor on a multisite and then convert some of those sites into single install sites, you can run into PHP fatal errors. The issue here is around our usage of `switch_to_blog`. We had checks in place to ensure we only used that if the proper Distributor meta was in place but in the above scenario, you can have that meta but actually not be on a multisite.

This PR fixes that by adding `is_multisite()` checks to any function that uses `switch_to_blog` after checking for the existence of Distributor meta. This should prevent fatal errors from occurring in the above detailed scenario.

### Alternate Designs

None

### Benefits

No PHP fatal errors will happen when using Distributor on sites that used to be part of a multisite

### Possible Drawbacks

None

### Verification Process

1. Create a WordPress multisite with two sites
2. Distribute some content from site A to site B
3. Convert site B to a stand-alone site
4. Edit one of the items that was distributed and see that no errors occur

### Checklist:

### Applicable Issues

#773 

### Changelog Entry

> Fixed - ensure we are on a multisite before using `switch_to_blog`

### DIFF
--- a/includes/push-ui.php
+++ b/includes/push-ui.php
@@ -542,7 +542,7 @@ function menu_content() {
 	$original_blog_id = get_post_meta( $post->ID, 'dt_original_blog_id', true );
 	$original_post_id = get_post_meta( $post->ID, 'dt_original_post_id', true );
 
-	if ( ! empty( $original_blog_id ) && ! empty( $original_post_id ) && ! $unlinked ) {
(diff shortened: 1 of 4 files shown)


helen (approved):
LGTM

PR #789, opened 2021-09-01
Fix stable release ZIP action

helen:
### Description of the Change

Workflows have to be present in the checkout, which means they have to be kept in stable branch. This is not a big deal because it's hidden anyway, but just to be super precise about it this adds a `.gitattributes` file to leave that out when creating the ZIP. Bonus: this now properly creates the ZIP so it unzips into a directory on the command line.

### Alternate Designs

You'll see in the commits I tried an rsync thing which was fine but realized it was overly complex.

### Benefits
Actual ZIP file attached to releases which we can reference from now on and get download stats for.

### Possible Drawbacks

n/a

### Checklist:

- [ ] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+/.github export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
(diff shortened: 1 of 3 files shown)


helen:
Alright I've tested this by running the various steps locally on my machine and then unzipping on the command line to check the directory structure, I think we are good. The final test will be... the release 😅

PR #800, opened 2021-09-23
Add filters to control terms and meta distribution for internal connections

dinhtungdu:
This PR adds two new filters that allow themes/plugins to control the term and meta distribution for internal connections.

### DIFF
--- a/includes/classes/ExternalConnections/WordPressExternalConnection.php
+++ b/includes/classes/ExternalConnections/WordPressExternalConnection.php
@@ -454,11 +454,17 @@ class WordPressExternalConnection extends ExternalConnection {
 			}
 
 			if ( ! empty( $post_array['meta'] ) ) {
-				\Distributor\Utils\set_meta( $new_post, $post_array['meta'] );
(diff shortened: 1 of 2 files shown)


dkotter (dismissed):

dkotter:
This looks good to me but just wanted to ask a couple questions:

1. This adds new filters to turn off the distribution of terms and/or meta when pushing content from an internal connection. Should we also add similar filters when pulling? I see we have already both `dt_push_post_media` and `dt_pull_post_media`, wondering if these two new filters should also have corresponding `pull` versions?
2. If we add `pull` versions, should we look to add these same filters when pulling content from external connections? We do already have the same `dt_pull_post_media` filter on external connections. I think it becomes more difficult to add the same filters when pushing from an external connection, just based on how that is handled, but could probably add those when pushing

dinhtungdu:
@dkotter Thanks so much for the head-up. I added `pull` versions of those filters, please take another look!

dkotter (approved):

dkotter:
Looks good!

PR #803, opened 2021-09-24
Add proper error `<ul>` so errors can be properly shown

### LINKED ISSUE #795: Uncaught TypeError: can't access property "innerText", T is null
Ive had this issue on some of my sites recently, the plugin was working fine without any issues and still is on some 7/8 of my sites.

`Uncaught TypeError: can't access property "innerText", T is null - push.js:127:2`

This happens on version 1.6.4 and 1.6.5 when distributing from the front end.

dkotter:
### Description of the Change

In our templates that output the connection dropdown, used when Pushing content, we were missing an unordered list element that we look for when outputting error messages. It appears this accidentally got removed when changes were made to better support [AMP](https://github.com/10up/distributor/pull/665).

Without this element, if an error occurs during pushing, that error message will not be shown and a console error will happen (as reported in #795). This PR fixes this by adding that error element back to each template that needs it.

### Alternate Designs

None

### Benefits

Error messages will properly show and no console errors will happen

### Possible Drawbacks

None

### Verification Process

Easy way to create an error during push is to modify the [`ajax_push`](https://github.com/10up/distributor/blob/1.6.5/includes/push-ui.php#L224) function and at the top of that, temporarily add a `wp_send_json_error`. Then go and try pushing some content. An error message should show and it should match whatever message as manually added.

### Checklist:

### Applicable Issues

Closes #795

### Changelog Entry

> Fixed - ensure error messages are shown properly if an error happens during a push

### DIFF
--- a/includes/push-ui.php
+++ b/includes/push-ui.php
@@ -594,6 +594,7 @@ function menu_content() {
 				<div class="loader-messages messages">
 					<div class="dt-error">
 						<?php esc_html_e( 'There was an issue loading connections.', 'distributor' ); ?>
+						<ul class="details"></ul>
 					</div>
 				</div>
 			</div>
--- a/templates/show-connections-amp.php
+++ b/templates/show-connections-amp.php
@@ -82,6 +82,7 @@
 			</div>
 			<div class="dt-error">
 				<?php esc_html_e( 'There were some issues distributing the post.', 'distributor' ); ?>
+				<ul class="details"></ul>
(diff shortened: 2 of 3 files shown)


dinhtungdu (approved):

Drmzindec:
Thank you for this update. The issue was that the original user who authorized the request was removed from the 1 site and the plugin couldnt verify. This update helped solve the problem. @dkotter

PR #807, opened 2021-10-12
Passing VIPCS check

### LINKED ISSUE #206: Ensure plugin passes WordPress VIP standards
This feedback is from a code review by @tomjn:

- The Update checker uses htmlentities in places were esc_html should be used
- The Update checkers uses @ operators to swallow a warning parse_url generates, but that warning was eliminated in PHP 5.3.3. If it can be bumped to PHP v5.3 minimum rather than 5.2 then these can be eliminated

### LINKED ISSUE #799: Enhancement: Support WP VIP Code Standards
During a commit of Distributor to a WP VIP Go repository, the VIP Bot flagged a number of issues with the plugin codebase. _[For privacy reasons, the site in question is not being identified here - contact me directly for further details.]_

Include the VIP code standards in tests, ensure that where practical code is updated for compatibility; where impractical PHPCS ignore rules are added.

**Designs**
N/A

N/A (The only alternative is not to use VIP Go)

**Additional context**
N/A

mehulkaklotar:
### Description of the Change

This PR ensures Distributor passes the phpcs check using WordPress-VIP-Go standard.

Excluded tests folder in this check.

### Verification Process

Make sure to run `composer install` and then run linting bu running `composer lint`

OR

Install VIPCS and run this command:

`phpcs --standard=WordPress-VIP-Go -s --extensions=php --ignore=node_modules,vendor,tests`

### Checklist:

### Applicable Issues

Fixes #799


includes/push-ui.php
```diff
@@ -7,6 +7,8 @@
 
 namespace Distributor\PushUI;
 
+use function Aws\filter;
```
dkotter: I think this was added by mistake

dkotter (changes requested):

includes/push-ui.php
mehulkaklotar: Yes, sorry. fixed now.

includes/external-connection-cpt.php
```diff
@@ -233,12 +235,8 @@ function ajax_verify_external_connection() {
 		exit;
 	}
 
-	$auth = array();
-	if ( ! empty( $_POST['auth'] ) ) {
-		$auth = $_POST['auth'];
-	}
-
-	$current_auth = get_post_meta( intval( sanitize_key( $_POST['endpointId'] ) ), 'dt_external_connection_auth', true );
+	$auth         = filter_input( INPUT_POST, 'auth', FILTER_REQUIRE_ARRAY );
```
dinhtungdu: `$auth` was assigned to `false` in my test, which causes a fatal error for the `array_merge` call below.

image

includes/external-connection-cpt.php
```diff
@@ -381,10 +383,8 @@ function save_post( $post_id ) {
 		// Create an instance of the connection to test connections
 		$external_connection_class = \Distributor\Connections::factory()->get_registered()[ sanitize_key( $_POST['dt_external_connection_type'] ) ];
 
-		$auth = array();
-		if ( ! empty( $_POST['dt_external_connection_auth'] ) ) {
-			$auth = $_POST['dt_external_connection_auth'];
-		}
+		$auth = filter_input( INPUT_POST, 'dt_external_connection_auth', FILTER_REQUIRE_ARRAY );
```
dinhtungdu: This line has the same issue as 238th.

dinhtungdu (commented):
@mehulkaklotar Thanks for the PR! The changes look good but I'm having issues with `filter_input` and `FILTER_REQUIRE_ARRAY` flag, please check the review comments for more detail. I'm not sure you have the same issue. I tested on PHP 8.0 and WordPress 5.8.1.

mehulkaklotar:
> @mehulkaklotar Thanks for the PR! The changes look good but I'm having issues with `filter_input` and `FILTER_REQUIRE_ARRAY` flag, please check the review comments for more detail. I'm not sure you have the same issue. I tested on PHP 8.0 and WordPress 5.8.1.

@dinhtungdu Yes, I get the sae warning as you. I have reverted those line back to original. Better to just ignore phpcs for the now. New changes pushed.

dinhtungdu (commented):
@mehulkaklotar I'm sorry for the late re-review. Is there any clear benefit of using `filter_input` over WordPress sanitize functions? Personally, I prefer using WordPress functions if possible.

mehulkaklotar:
@dinhtungdu Yes, actually I prefer to be less dependent of WP for these utility functions. If you want, I can change it. Mostly it is used for urls and nonces with some other few fields.

dinhtungdu:
@mehulkaklotar Can you explain why do you prefer to be less dependent on WP for these utility functions? Given that Distributor is a WordPress plugin.

> If you want, I can change it.

I think we want the best and most sensible solution, not the one that I want : ).

mehulkaklotar:
@dinhtungdu Actually if you look at [this line](https://github.com/10up/distributor/pull/807/files#diff-c9864a237f6f3f3b3956fd9b30b7b46adee6ea100bf5fad5684b68de30ba055dR230) for example, it is a integer value we need from POST variable as checked [here](https://github.com/10up/distributor/pull/807/files#diff-c9864a237f6f3f3b3956fd9b30b7b46adee6ea100bf5fad5684b68de30ba055dL283), so when you use `sanitize_text_field`, it will always return ANY sanitized value, but filter_input with `FILTER_VALIDATE_INT` will always return integer if passed correctly, otherwise gives false. so that it fails or be successful correctly.

dinhtungdu:
Did some research on `filter_input`, we have an open issue considering adding it to our Engineering Best Practices [here](https://github.com/10up/Engineering-Best-Practices/issues/132).

There is also an issue on xwp/stream raised the reliable issue of using `filter_input` https://github.com/xwp/stream/issues/254 which is also followed by

> Interesting. Yeah, and Nacin said that the filter_* functions aren't allowed in core because of un-guaranteed availability. I've removed them from Widget Customizer.

Looking at this issue, as long as we provide the third parameter for `filter_input`, it's fine to use it. But the un-guaranteed availability issue is ~a deal blocker to me~ something we should take into account.

Pinging @dkotter here for another perspective.

mehulkaklotar:
@dinhtungdu I agree with you on not using `filter_input` as it is not described in our 10up engineering best practices and even from WP's un-guaranteed availability. It seems to be an older comments from 2016 and it is still available. 

WPCS's lint gives errors when running phpcs if variable is sanitized even via `sanitize_title` or `sanitize_text_field` function from the core. so we need to decide if we want to put "so many phpcs ignores" in the code for this or we can keep `filter_input` always using with 3rd parameter.

Pinging @dkotter for another perspective here.

dkotter:
> so we need to decide if we want to put "so many phpcs ignores" in the code for this or we can keep `filter_input` always using with 3rd parameter

I know we use `filter_input` in quite a few other 10up plugins (for instance, the Classifai PR that was mentioned in #799 uses them). I personally think it's fine to use as long as we're passing in a valid 3rd parameter (otherwise no actual filtering is done). We also need to ensure we don't break anything by unintentionally adding more strict validation (for instance, URL validation using this approach may be more strict than what we were doing before). Definitely don't want to introduce a bug by trying to fix PHPCS issues.

I can't find much information on the reliability of `filter_input`. It seems like an issue with older PHP versions but we do claim to still support PHP 5.6 here. I know we want to bump that higher soon and I'm hoping most, if not all users of Distributor are on at least PHP 7.0 and from what I can tell, there's no reliability issues there.

So I guess my opinion is I'm fine using `filter_input`, especially if that means we don't need a bunch of ignore statements littering our code.

peterwilsoncc:
> WPCS's lint gives errors when running phpcs if variable is sanitized even via `sanitize_title` or `sanitize_text_field` function from the core. so we need to decide if we want to put "so many phpcs ignores" in the code for this or we can keep `filter_input` always using with 3rd parameter.

I'm not seeing this, is it possible you had forgotten to run `wp_unslash()` beforehand?

To test I changed the license key check in `includes/settings.php` to the following and confirmed no warnings were shown by either the WP or VIP standards:

```php
// Please ignore that these are the wrong sanitization functions, I was testing.
$email = sanitize_text_field( wp_unslash( $_POST['dt_settings']['email'] ) );
$license_key = sanitize_title( wp_unslash( $_POST['dt_settings']['license_key'] ) );
```

peterwilsoncc:
I'm going to close this pull request in favour of the alternative approach in #901.

The alternative approach requires fewer `//phpcs` suppression comments and makes use of the WP Core sanitization functions.

PR #811, opened 2021-10-18
Change how the `New` tab is generated on the Pull screen for external connections

### LINKED ISSUE #808: Previously pulled posts appear in new tab
**Describe the bug**
When having two separate WP instances connected via an External Connection and using the Pull Content screen, previously pulled posts appear in the New tab view.  Note that in the case that was demo'd to me that the Bulk Pull option was used and not individual post pull actions.

**Steps to Reproduce**

**Expected behavior**
Previously pulled posts should only show in the Pulled tab view and not in the New tab view.

**Screenshots**

Screen Shot 2021-10-13 at 9 33 19 AM

**Environment information**
 - Device: 
 - OS: 
 - Browser and version: 
 - WordPress version: 
 
 - Plugins and version: 
 - Theme and version: 
 
 - <details><summary>Site Health Info:</summary>
   
   </details>

**Additional context**

### LINKED ISSUE #809: Performance issues on Pull Content screen when origin site has 100k+ posts
**Describe the bug**
When having two separate WP instances connected via an External Connection and using the Pull Content screen where the origin site selected has over 100,000 posts (in this specific case ~230,000 posts) there are performance issues when trying to filter the New/Pulled/Skipped posts as well as paginate within those tabbed sections.

**Steps to Reproduce**

**Expected behavior**
Pagination and filtering on the Pull Content screen should work regardless of the number of posts on an origin site.

**Screenshots**

Screen Shot 2021-10-13 at 9 39 12 AM

**Environment information**
 - Device: 
 - OS: 
 - Browser and version: 
 - WordPress version: 
 
 - Plugins and version: 
 - Theme and version: 
 
 - <details><summary>Site Health Info:</summary>
   
   </details>

**Additional context**
This issue relates to #808 as the New tab view is polluted with previously pulled posts which makes the New tab view have more posts rendering than necessary.  Once 808 is resolved, this performance issue may relax a bit but should still be something we investigate and work to handle for as best we can for sites with a large amount of content to filter through and pull.

dkotter:
### Description of the Change

When building the content that shows on the `New` tab of the Pull screen, we determine what content has already been pulled and what content has already been skipped. This is then trimmed down to 200 items and used in a `post__not_in` query.

This can cause two different problems:
1. Since we limit this to 200 items, if you have more than 200 items, you could end up seeing already pulled or skipped items on the `New` tab
2. You might see performance problems if you have lots of items that need excluded. This will mostly impact external connections and even more so, WordPress.com VIP sites, since we use a lower timeout value in those requests

This PR attempts to fix both of those problems in the following ways:
1. Create a new custom endpoint that is used to generate the content for the `New` tab on the Pull screen
2. Utilize this endpoint if we are on the Pull screen of an external connection and we have no previously pulled or skipped items
3. Make a `POST` request to this endpoint (instead of `GET`), which allows us to pass data via the `body` parameter instead of in the URL
4. Remove the 200 item limit on excluded items, since those are now passed in the `body`

Because we remove that 200 item limit, we won't run into issues with previously pulled or skipped items showing on the `New` tab. Fixes #808 

Because we are using a custom endpoint with a `POST` request, fixes potential timeout issues when on a WordPress.com VIP site because we no longer have the strict 3 second timeout limit. Fixes #809 

### Alternate Designs

Ultimately we are still running a `post__not_in` query, which may not be the most performant. Other approaches were considered but ultimately rejected:

1. Instead of using `post__not_in`, increase our `posts_per_page` value and then remove excluded items after. This can still cause performance issues and causes pagination issues
2. To get around the performance issues of a large `posts_per_page`, could set an upper limit and break out into new queries once that limit is reached. Still have pagination problems and could still have performance issues because we are running more queries
3. To fix pagination issues, would have to potentially run multiple extra queries to determine proper offset. For instance, if someone wants page 10 of the results, we can't assume page 10 of the query matches what we want. We would have to run queries for each page leading up to page 10 to determine what offset we need. This would have performance implications for higher paginated pages
4. Instead of excluding items based on their post ID, have a piece of meta associated with all external pulled items that we can use to exclude. This would fix things going forward but isn't backwards compatible

### Benefits

All excluded items are properly excluded from the `New` tab. Better performance in certain situations

### Possible Drawbacks

I can't think of any but maybe some drawback around switching from a `GET` request to a `POST` request?

### Verification Process

1. Set up a site with at least one internal connection and one external connection
2. Ensure the Pull screen works for both
3. Try pulling and skipping content from both
4. Ensure things still work. Be sure to test pagination, search, skipping, pulling, etc.
5. Ideal is to test with a large amount of content pulled in, though this can be tedious to do

### Checklist:

### Applicable Issues

#808 
#809 

### Changelog Entry

> Fixed - change how the Pull screen `New` tab is populated for external connections

### DIFF
--- a/includes/classes/ExternalConnections/WordPressExternalConnection.php
+++ b/includes/classes/ExternalConnections/WordPressExternalConnection.php
@@ -165,20 +165,6 @@ class WordPressExternalConnection extends ExternalConnection {
 			}
 		}
 
-		/**
-		 * Filter the remote_get query arguments
-		 *
-		 * @since 1.0
-		 * @hook dt_remote_get_query_args
-		 *
-		 * @param  {array}  $query_args The existing query arguments.
-		 * @param  {array}  $args       The arguments originally passed to `remote_get`.
-		 * @param  {object} $this       The authentication class.
-		 *
-		 * @return {array} The existing query arguments.
-		 */
-		$query_args = apply_filters( 'dt_remote_get_query_args', $query_args, $args, $this );
-
 		// When running a query for the Pull screen with excluded items, make a POST request instead
 		if ( empty( $id ) && isset( $args['post__not_in'] ) && isset( $args['dt_pull_list'] ) ) {
 			$query_args['post_type'] = isset( $post_type ) ? $post_type : 'post';
@@ -256,6 +242,20 @@ class WordPressExternalConnection extends ExternalConnection {
 		if ( ! empty( $posts_per_page ) ) {
 			$args_str .= 'per_page=' . (int) $posts_per_page;
 		}
+
+		/**
+		 * Filter the remote_get query arguments
+		 *
+		 * @since 1.0
+		 * @hook dt_remote_get_query_args
+		 *
+		 * @param  {array}  $query_args The existing query arguments.
+		 * @param  {array}  $args       The arguments originally passed to `remote_get`.
+		 * @param  {object} $this       The authentication class.
+		 *
+		 * @return {array} The existing query arguments.
+		 */
+		$query_args = apply_filters( 'dt_remote_get_query_args', $query_args, $args, $this );
 
 		foreach ( $query_args as $arg_key => $arg_value ) {
 			if ( is_array( $arg_value ) ) {
@@ -402,6 +402,7 @@ class WordPressExternalConnection extends ExternalConnection {
 					/**
 					 * Filter the timeout used when calling `remote_post`
 					 *
+					 * @since 1.6.7
 					 * @hook dt_remote_post_timeout
 					 *
 					 * @param int $timeout The timeout to use for the remote post. Default `45`.
@@ -410,7 +411,18 @@ class WordPressExternalConnection extends ExternalConnection {
 					 * @return int The timeout to use for the remote_post call.
 					 */
 					'timeout' => apply_filters( 'dt_remote_post_timeout', 45, $args ),
-					'body'    => $args,
+					/**
+					 * Filter the remote_post query arguments
+					 *
+					 * @since 1.6.7
+					 * @hook dt_remote_post_query_args
+					 *
+					 * @param {array}  $args The request arguments.
+					 * @param {WordPressExternalConnection} $this The current connection object.
+					 *
+					 * @return {array} The query arguments.
+					 */
+					'body'    => apply_filters( 'dt_remote_post_query_args', $args, $this ),
 				)
 			)
 		);
--- a/includes/classes/ExternalConnections/WordPressExternalConnection.php
+++ b/includes/classes/ExternalConnections/WordPressExternalConnection.php
@@ -165,6 +165,33 @@ class WordPressExternalConnection extends ExternalConnection {
 			}
 		}
 
+		/**
+		 * Filter the remote_get query arguments
+		 *
+		 * @since 1.0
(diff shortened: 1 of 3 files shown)


jeffpaul:
@dkotter its probably worthwhile to add some explicit comments in here as to why we took this approach especially for the case where VIP or someone else might flag it as a concern (thus there would be some comments for them to reference and better understand the _why_ maybe even with a link to the PR or issue with our analysis and stats on the topic).

includes/classes/ExternalConnections/WordPressExternalConnection.php
```diff
@@ -370,6 +383,86 @@ public function remote_get( $args = array() ) {
...
+		foreach ( $posts as $post ) {
+			$post['full_connection'] = ! empty( $response_headers['X-Distributor'] );
+
+			$formatted_posts[] = $this->to_wp_post( $post );
+		}
+
+		$total_posts = ! empty( $response_headers['X-WP-Total'] ) ? $response_headers['X-WP-Total'] : count( $formatted_posts );
+
+		return apply_filters(
+			'dt_remote_post',
```
dinhtungdu, line 480: We may want to document this filter. The filter name is bit vague to me but I can't think about any better one 😅

includes/rest-api.php
```diff
@@ -338,6 +405,125 @@ function check_post_types_permissions() {
...
+	if ( ! empty( $request['search'] ) ) {
+		$args['s'] = urldecode( $request['search'] );
+	}
+
+	if ( ! empty( $request['exclude'] ) ) {
+		$args['post__not_in'] = $request['exclude'];
+	}
+
+	// Filter documented in includes/classes/ExternalConnections/WordPressExternalConnection.php
+	$query = new \WP_Query( apply_filters( 'dt_remote_get_query_args', $args, $request ) );
```
dinhtungdu: Can you explain why do we need this filter in the callback, which will be executed on the remote sites, not from the site making the request?

Also, how can we distinguish between normal `remote_get` calls and pull content calls? I'm thinking about checking for the existence of the third argument.

dinhtungdu (commented):
@dkotter Awesome work! 🎉 This is looking good to me. I just have some minor comments, please take a look : )

includes/classes/ExternalConnections/WordPressExternalConnection.php
dkotter: Thanks, added docs here. And I just based this name of our existing `dt_remote_get` filter that runs in our `remote_get` method. So figured since this is a new `remote_post` method, `dt_remote_post` made sense.

includes/rest-api.php
dkotter: I've gone ahead and removed this filter as it isn't needed. We run the same filter before we end up making our remote POST request, so this would just result in filtering the same values twice.

dkotter: And after thinking about it a bit more, not sure it makes sense to run the first filter, as that's meant to filter the remote_get query args. I've moved that code to run after our remote post call and added a new filter in the remote post method around those arguments. So to answer your second question, if you only want to filter the arguments for a Pull call, you'd use this new filter instead

dinhtungdu (commented):
LGTM 🎉

helen (approved):
:shipit:

dinhtungdu (approved):
Not sure why I didn't approve this.

PR #845, opened 2022-02-02
Fix a PHP fatal error when updating a distributed item that uses the block editor

### LINKED ISSUE #844: Fatal error when updating an item that has been distributed on WP 5.9
### Describe the bug

On WordPress 5.9, I can Pull content and Push content, both from internal and external connections. But when trying to update an item (in this case a post) that has already been distributed, the save process fails and I get an error in the editor. This prevents me from making any updates to items that have been distributed, not to mention no updates will be syndicated (since no updates can happen).

Looking at my debug log, the culprit here is this:
```
PHP Fatal error:  Uncaught TypeError: call_user_func_array(): Argument #1 ($callback) must be a valid callback, function "_disable_block_editor_for_navigation_post_type" not found or invalid function name in /var/www/html/wp-includes/class-wp-hook.php:307
Stack trace:
#0 /var/www/html/wp-includes/plugin.php(189): WP_Hook->apply_filters()
#1 /var/www/html/wp-content/plugins/distributor/includes/utils.php(953): apply_filters()
#2 /var/www/html/wp-content/plugins/distributor/includes/subscriptions.php(285): Distributor\Utils\dt_use_block_editor_for_post_type()
#3 /var/www/html/wp-includes/class-wp-hook.php(309): Distributor\Subscriptions\send_notifications()
...
```

Distributor has a utility function called `dt_use_block_editor_for_post_type` which mirrors the same functionality the core `use_block_editor_for_post_type` function has (but allows us to support pre-WP 5.0 installs). The last line of this function calls a core filter, `use_block_editor_for_post_type`. In WP 5.9, a function was hooked to that filter (`_disable_block_editor_for_navigation_post_type`) which lives within the `wp-admin/includes/post.php` directory.

This directory is not loaded in some contexts (I believe REST API contexts) so when our `dt_use_block_editor_for_post_type` function is called during a post update (specifically in the block editor) and it subsequently calls the `use_block_editor_for_post_type` filter, the `_disable_block_editor_for_navigation_post_type` function is registered to that filter but the function won't exist since the file isn't loaded, leading to the fatal error.

### Steps to Reproduce

1. Be on a WP 5.9 install
2. Create a new post using the block editor
2. Distribute that post to at least one connection
3. Try updating that post
4. Notice an update error happens and no updates are syndicated

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

### LINKED ISSUE #847: Could not determine remote post id
Attempt to push Post (not as draft) results in "Could not determine remote post id" error message, post is not pushed.

As we only use Distributor once in a couple of weeks, I don't know what changes on either end might have caused this (pushing worked before). Updated Distributor to 1.6.8 on both ends, error still present. Source WP is 5.8.2, target was 5.7.5 upgraded to 5.9 with no change to this behavior.

I may be helpful in tracing the cause of the error if you please explain to me to what exactly does distributor respond with it (because pushing creates the remote post, I don't know, why shouldn't it be able to determine its id).

Nothing in debug.log nor PHP err log.

Thank you!

dkotter:
### Description of the Change

As described in #844, WordPress 5.9 added a new function hooked to the `use_block_editor_for_post_type` filter. We run this filter in our `dt_use_block_editor_for_post_type` function, which runs when an item that has already been distributed is saved (among other places). If this item uses the block editor, saves happen via a REST API request and the new function WP 5.9 hooks doesn't exist in that context.

This PR fixes that by checking if the `use_block_editor_for_post_type` function exists (which is in the same file as the newly hooked function, `_disable_block_editor_for_navigation_post_type `. If it does exist, we apply the `use_block_editor_for_post_type`. If it doesn't exist, we apply a newly added filter, `dt_use_block_editor_for_post_type`. This ensures this response can be filtered no matter the context.

Closes #844 

### Alternate Designs

Instead of doing `function_exists( 'use_block_editor_for_post_type' )`, we could just `remove_action( 'use_block_editor_for_post_type', '_disable_block_editor_for_navigation_post_type' );`. My concern with this approach is while it solves the immediate issue, there could be other functions in the future hooked up that cause the same problem and we would have to remove those hooks individually as well.

Another approach I considered is using `function_exists` but if it doesn't exist, just return `true` instead of passing `true` through a new filter. I figured having a new filter gives developers the most flexibility.

### Possible Drawbacks

Any code that is hooked to the `use_block_editor_for_post_type` will no longer run when the `dt_use_block_editor_for_post_type` function is called during a REST API request. This can be solved by instead using the new `dt_use_block_editor_for_post_type` filter

### Verification Process

1. Be on a WP 5.9 install
2. Create a new post using the block editor
3. Distribute that post to at least one connection
4. Try updating that post
5. No errors should happen and updates should be syndicated

### Checklist:

### Changelog Entry

> Fixed - Ensure content updates work for distributed items that use the block editor in WordPress 5.9+

### Credits

Props @dkotter

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -949,8 +949,24 @@ function dt_use_block_editor_for_post_type( $post_type ) {
 		return false;
 	}
 
-	// Filter documented in WordPress core.
-	return apply_filters( 'use_block_editor_for_post_type', true, $post_type );
+	// In some contexts this function doesn't exist so we can't reliably use the filter.
+	if ( function_exists( 'use_block_editor_for_post_type' ) ) {
+		// Filter documented in WordPress core.
+		return apply_filters( 'use_block_editor_for_post_type', true, $post_type );
+	}
+
+	/**
+	 * Filters whether an item is able to be edited in the block editor.
+	 *
+	 * @since 1.6.9
+	 * @hook dt_use_block_editor_for_post_type
+	 *
+	 * @param bool   $use_block_editor Whether the post type uses the block editor. Default true.
+	 * @param string $post_type        The post type being checked.
+	 *
+	 * @return bool Whether the post type uses the block editor.
+	 */
+	return apply_filters( 'dt_use_block_editor_for_post_type', true, $post_type );
 }
 
 /**


cadic (changes requested):
@dkotter with possible drawbacks you've desrcibed, it worth a documentation update to help 3-rd parties who might already rely on `use_block_editor_for_post_type` filter.

dkotter:
> @dkotter with possible drawbacks you've desrcibed, it worth a documentation update to help 3-rd parties who might already rely on `use_block_editor_for_post_type` filter.

Thanks for the review @cadic! I agree we'll want to properly document this, do you have suggestions on the best place for that? Not sure we want this in the README, was thinking maybe this is just a note that we add to our release changelog (which would happen separately from this PR)?

cadic:
@dkotter agree, the note in the release changelog is enough for this.

cadic (approved):

PR #853, opened 2022-02-16
PHP 8 compatibility

dkotter:
### Description of the Change

This PR updates our PHPCS testing to ensure we run PHP compat tests on PHP 5.6+, hopefully flagging any issues with code that isn't compliant with our minimums. It also updates our PHPUnit tests to run on PHP 7.1+, again hopefully catching issues that don't match our supported versions. Ideally we would run on all versions from 5.6+ (or at least 5.6, 7.0 and 8.0) but we are using WP Mock as a dependency and it has issues running on 5.6, so until those are addressed, we are limited here.

We also remove WP Acceptance testing in this PR, as that's been notoriously flaky and we are looking to replace those tests with cypress.

Note: the PHPCompatibility tool does not yet fully support PHP 8. So while the hope coming out of this PR is to have some automated scanning to ensure all existing code and any newly introduced code works on all supported PHP versions, until that tool adds full 8.0+ support, we can't guarantee that.

### Alternate Designs

None

### Possible Drawbacks

More tests running

### Verification Process

Ensure newly added checks are passing

### Checklist:

### Changelog Entry

> Added - More robust PHP testing

### Credits

Props @dkotter


.github/workflows/test.yml
```diff
@@ -12,51 +16,44 @@ on:
...
 
   phpunit:
-    runs-on: ubuntu-latest
+    name: ${{ matrix.php }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        # We claim to support from 5.6+ but WP Mock only supports 7.1+
+        # Also an issue with WP Mock not working on PHPUnit 9.5+, which PHP 7.3+ needs.
```
peterwilsoncc: Testing locally on PHP 7.4 I managed to resolve this by adding `'phpunit/phpunit':'>=7.0 <9.0'` to this projects composer file.

Further bumping to PHP 8.0 still threw some errors, `Mockery\Exception\NoMatchingExpectationException`, but two more versions of PHP seems better than nothing.

dkotter: Ah, nice catch. Adding that brought 7.3 and 7.4 and I was able to fix the couple broken tests so we also have 8.0 and 8.1 now. Thanks!

peterwilsoncc (approved):
Thanks Darin.

PR #877, opened 2022-04-27
Feature/858 Ability to set roles allowed to Pull

### LINKED ISSUE #858: Add ability to set roles allowed to Pull content when creating an External Connection

When setting up/editing an External Connection currently, you can choose which roles are allowed to Push content. But there is no way to set which roles can Pull content. This can be changed via a [filter](https://github.com/10up/distributor/issues/626#issuecomment-673276684) but seems like we should have the same settings for Pull as we do for Push.

### Designs

Screen Shot 2022-03-08 at 3 44 53 PM

_No response_

### Code of Conduct

faisal-alvi:
### Description of the Change

- This PR introduces a new setting to set user roles who are allowed to pull content.
- It also respects the `apply_filters( 'dt_pull_capabilities', 'manage_options' )` filter.
- If the current user has a cap coming from the filter, the current user will be allowed to pull despite the user roles settings.
- If the current user does not has a cap coming from the filter but has a user role that is allowed in the setting, then the current user can pull content.
- By default, `administrator` and `editor` roles are allowed to Pull content.
- The `dt_capabilities` filter is removed completelty.

Closes #858.

### Alternate Designs

image
[image link](https://www.screencast.com/t/A7Db62H8NLI)

### Possible Drawbacks

- The `dt_capabilities` filter is removed completely, even if users stop using the filter, they will still see the main menu displayed. So far, the main `Distributor` menu was displayed via a filter (default to `manage_options`). But this PR removes this filter and displays the menu for all users by using the `read` cap.

https://github.com/10up/distributor/blob/08e6127e9a88c70560e59885171cf837540f2ce1/includes/external-connection-cpt.php#L663-L664

- Also removes the filter while displaying the submenu `Distributor > Pull Content` and displays it for all users by using the `read` cap. 

https://github.com/10up/distributor/blob/08e6127e9a88c70560e59885171cf837540f2ce1/includes/pull-ui.php#L154-L155

This was required because unlike the `Push Allowed` user role setting, we have to individually check each external connection to see if the current user can `Pull Content` or not. And to do so, we have to show the `Distributor > Pull Content` menu to all user roles.

### Verification Process

- Make sure you have at least 1 external connection set up.
- Make sure you have a few posts listed in `Distributor > Pull Content` and ready to be pulled. (If not, visit your external connection site, create a couple of posts and Distribute them to your original site, then you will see those posts are listed and ready to be pulled.)
- Please login as admin.
- Visit `Distributor > External Connections` and edit your external connection.
- See the `Roles Allowed to Pull` setting at the bottom.
- `Administrator` and `Editor` roles should be checked by default.
- Create an editor role user if does not exist already.
- Login to the editor role user in a separate browser (incognito recommended).
- See only the `Distributor > Pull Content` submenu is visible under the `Distributor`.
- Visit `Distributor > Pull Content`.
- See the posts listed - ready to be pulled.
- Try to pull 1 post, you should be able to pull successfully.
- Now go back to the browser/window having the admin role logged in.
- Visit `Distributor > External Connections` and edit your external connection.
- Untick the `Editor` role under the `Roles Allowed to Pull` setting at the bottom.
- Switch the browser/window having the editor role logged in.
- Visit `Distributor > Pull Content`.
- No posts should be listed!
- **Now let's check the filter:**
- Add the following code in your theme's `functions.php` file.
```
add_filter( 'dt_pull_capabilities', function() {
	return 'edit_others_posts';
} );
```
- Now even if the `Editor` role is unticked under the `Roles Allowed to Pull`, you should be able to see posts listed and ready to be pulled under the `Distributor > Pull Content`. This is because the editor role is now allowed to pull content via the filter.

### Checklist:

### Changelog Entry

> Added - Ability to set user roles to Pull Content
> Changed - The `Distributor > Pull Content` menu is now be visible for all user roles.
> Removed - The `dt_capabilities` & `dt_pull_capabilities` filters are removed while displaying the menus.

### Credits

Props @faisal-alvi @dkotter


includes/debug-info.php
```diff
@@ -184,9 +184,10 @@ function get_formatted_external_connnections() {
 	}
 
 	foreach ( $external_connections->posts as $external_connection_id ) {
-		$external_connection_type          = get_post_meta( $external_connection_id, 'dt_external_connection_type', true );
-		$external_connection_status        = get_post_meta( $external_connection_id, 'dt_external_connections', true );
-		$external_connection_allowed_roles = get_post_meta( $external_connection_id, 'dt_external_connection_allowed_roles', true );
+		$external_connection_type               = get_post_meta( $external_connection_id, 'dt_external_connection_type', true );
+		$external_connection_status             = get_post_meta( $external_connection_id, 'dt_external_connections', true );
+		$external_connection_push_allowed_roles      = get_post_meta( $external_connection_id, 'dt_external_connection_allowed_roles', true );
```
dkotter: Looks like spacing is off slightly here

includes/external-connection-cpt.php
```diff
@@ -654,13 +681,12 @@ function add_menu_item() {
...
 		 *
 		 * @since 1.0.0
-		 * @hook dt_capabilities
 		 *
 		 * @param {string} 'manage_options' The capability allowed to view external connections.
 		 *
 		 * @return {string} The capability allowed to view external connections.
 		 */
-		apply_filters( 'dt_capabilities', 'manage_options' ),
+		 'read',
```
dkotter: I think we're fine to remove this filter but we should also remove the filter docblock above this, as that's no longer needed

includes/pull-ui.php
```diff
@@ -76,6 +76,13 @@ function setup_list_table() {
 		}
 	}
 
+	// Current user roles.
+	$current_user_roles = (array) wp_get_current_user()->roles;
+
+	// Check if user has cap to pull any external connection.
+	$pull_capabilities        = apply_filters( 'dt_pull_capabilities', 'manage_options' );
```
dkotter, line 91: We should move the filter docblock to above this, so this gets added properly to our doc site

includes/pull-ui.php
```diff
@@ -76,6 +76,13 @@ function setup_list_table() {
...
+	$can_pull_all_connections = current_user_can( esc_html( $pull_capabilities ) );
```
dkotter: I don't believe we need the `esc_html` here

includes/pull-ui.php
```diff
@@ -145,13 +165,12 @@ function action_admin_menu() {
...
 		 *
 		 * @since 1.0.0
-		 * @hook dt_pull_capabilities
 		 *
 		 * @param {string} 'manage_options' The capability allowed to pull content.
 		 *
 		 * @return {string} The capability allowed to pull content.
 		 */
-		apply_filters( 'dt_pull_capabilities', 'manage_options' ),
+		 'read',
```
dkotter: We can remove the filter docblock here

dkotter (changes requested):

includes/pull-ui.php
faisal-alvi: It was added to take care of the spam strings coming from the filter, removed for now.

dkotter:
I had not considered initially that this becomes more complicated because you may have some connections you want certain user roles to have access to (like Editors) and others you may not want that same role to have access to. The way we're achieving that in this PR is to show the Distributor menu to everyone, even if someone doesn't have access to any connections. I understand why but not sure I like this.

In addition, we should probably only allow Admins by default, not admins and editors, as we currently only support admins by default. So I don't think we'd want to change that here.

In addition, if you're user role has no permissions you still get the normal Pull screen, with the ability to add a new connection. I'd suggest instead they get some sort of error message and they shouldn't have the ability to add new connections. So seems we may be introducing a few regressions here and not sure that's worth it for the benefit of making it easier to choose which roles can Pull. May be worth just closing this out and leaving things as-is, where users can utilize existing filters to modify this behavior themselves

faisal-alvi:
> I had not considered initially that this becomes more complicated because you may have some connections you want certain user roles to have access to (like Editors) and others you may not want that same role to have access to. The way we're achieving that in this PR is to show the Distributor menu to everyone, even if someone doesn't have access to any connections. I understand why but not sure I like this.

I think we can achieve "not displaying" Pull screen for every user role by storing the Pull abilities in the `option` key(s) and only checking that key, which would hold the user role only if it can Pull from at least one external connection.

> In addition, we should probably only allow Admins by default, not admins and editors, as we currently only support admins by default. So I don't think we'd want to change that here.

Agree.

> In addition, if you're user role has no permissions you still get the normal Pull screen, with the ability to add a new connection.

No, unless you are an admin, you will not have the ability to add a new connection. Please let me know if that is visible in any case so we can fix it. The following screenshot shows the screen for an "editor" role, where they don't have access to `Distributor > External Connections`.

image

> May be worth just closing this out and leaving things as-is, where users can utilize existing filters to modify this behavior themselves

Also agree with closing this out.

PR #882, opened 2022-05-20
Remove `-dev` from our version number

### LINKED ISSUE #820: Document or remove `-dev` from the PHP version number
The `-dev` suffix has proven to be a little confusing at release time and may not even be necessary now. We should evaluate and then, if it is necessary, document it in code and documentation better. My recollection is that it was added primarily to be able to warn users that they're running a development version and need to run the build from time to time. We could probably achieve the same result but checking if one of the dev version files exists instead.

A less-confusing release process.

**Designs**
n/a

Leaving it alone

**Additional context**

dkotter:
### Description of the Change

In #210, we added `-dev` to our version number in order to output a warning for users that may be running this plugin straight from a non-built branch. Later we added more code around this to remove that `-dev` during release. There's nothing wrong with this approach but does cause some confusion during the release process and having to remember when to add that `-dev` or when to remove it.

This PR changes that up to instead look for the existence of the `composer.lock` file. If that exists, we then output the same warning message. This file should only exist in branches that aren't built, for instance `stable` won't contain that file. Since we're not relying on that `-dev` string anymore, I also removed any code that was in place to support that.

Closes #820

### Alternate Designs

Could leave as-is and just ensure we better document why we have this

### Possible Drawbacks

None

### Verification Process

1. Check this branch out
2. Ensure you still see a warning message when visiting any Distributor specific pages (like the settings page or Pull screen)
3. For more testing, run `npm run release`, which will build a `release` directory
4. Move this directory to the root of the `plugins` directory, temporarily replacing the  current `distributor` directory
5. You should now be running the "released" version of Distributor. Ensure you don't see that warning anymore and also don't have any other issues

### Checklist:

### Changelog Entry

> Changed - Update how we check if someone is running a development version of Distributor


CONTRIBUTING.md
```diff
@@ -29,17 +29,17 @@ The `develop` branch is the development branch which means it contains the next
...
 1. Check release: Wait for the [Publish Release Action](https://github.com/10up/distributor/actions?query=workflow%3A%22Publish+Release%22) to complete, and then check the latest release to ensure that the ZIP has been attached as an asset.  Download the ZIP and inspect the contents to be sure they match the contents of the `stable` branch.
-1. Version bump (again): In the `develop` branch (`cd ../ && git checkout develop`) bump the version number in `distributor.php` and `readme.txt` to `X.Y.(Z+1)-dev`.  It's okay if the next release might be a different version number; that change can be handled right before release in the first step, as might also be the case with `@since` annotations.
+1. Version bump (again): In the `develop` branch (`cd ../ && git checkout develop`) bump the version number in `distributor.php` and `readme.txt` to `X.Y.(Z+1)`.  It's okay if the next release might be a different version number; that change can be handled right before release in the first step, as might also be the case with `@since` annotations.
```
jeffpaul: We probably don't need this step at all then either, yeah?

dinhtungdu (dismissed):
Tested and worked flawlessly! I agree with @jeffpaul that we may not need the second version bump step anymore. Otherwise, this is LGTM!

PR #888, opened 2022-06-03
Adds slashes to a string or recursively adds slashes to strings within an array for the JSON value while migrating metadata

### LINKED ISSUE #724: JSON decode failure while migrating metadata due to wp_unslash function
Issue:
- WordPress stores metadata passing through wp_unslash function
   - https://github.com/WordPress/WordPress/blob/master/wp-includes/meta.php#L53
   - https://github.com/WordPress/WordPress/blob/master/wp-includes/meta.php#L185

- Due to this whatever data is stored using the above functions will be converted to unescaped JSON which is causing issues while decoding the JSON data.

Possible Solutions:
- We can create a hook while we are storing the metadata
- We can add conditional logic before adding metadata, that if the string is JSON then we can use `wp_slash` before using add_post_meta/update_post_meta
- We can use the `global $wpdb` custom query to insert data

I will be happy to have your's input on this and will be happy to contribute too.

Thanks

### LINKED ISSUE #738: Distributed post have broken accents
When distributing a post that consists of some diacritical characters the pushed version seems to have them converted to unicode and removed the slashes. I cannot find the proper place where this may take place.

The content is edited via Elementor.
Distributed to multisite website (different domain).

What is weird is that sometimes this happens but sometimes it pushes the content correctly.

Example:
image

sksaju:
### Description of the Change

Adds slashes to a string or recursively adds slashes to strings within an array for the JSON value while migrating metadata

This PR fixes the issues https://github.com/10up/distributor/issues/738 and https://github.com/10up/distributor/issues/724

### Verification Process

- After checking out this branch, try to migrate a post that included JSON metadata and try to migrate Elementor Post

### Checklist:

### Credits

Props @

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -202,7 +202,7 @@ function set_meta( $post_id, $meta ) {
 			json_decode( $meta_value );
 			// Adds slashes to a string or recursively adds slashes to strings within an array for the JSON
 			if ( json_last_error() === JSON_ERROR_NONE || '_elementor_data' === $meta_key ) {
-				$meta_value = \wp_slash( $meta_value );
+				$meta_value = wp_slash( $meta_value );
 			}
 
 			if ( $has_prev_value ) {
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -199,13 +199,10 @@ function set_meta( $post_id, $meta ) {
 				$meta_value = maybe_unserialize( $meta_value );
 			}
 
-			// Adds slashes to a string or recursively adds slashes to strings within an array to avoid JSON decode failure while migrating metadata
-			$meta_value = wp_slash( $meta_value );
-
 			if ( $has_prev_value ) {
-				update_post_meta( $post_id, $meta_key, $meta_value, $prev_value );
+				update_post_meta( $post_id, wp_slash( $meta_key ), wp_slash( $meta_value ), $prev_value );
 			} else {
-				add_post_meta( $post_id, $meta_key, $meta_value );
+				add_post_meta( $post_id, wp_slash( $meta_key ), wp_slash( $meta_value ) );
 			}
 		}
 	}
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -199,6 +199,12 @@ function set_meta( $post_id, $meta ) {
 				$meta_value = maybe_unserialize( $meta_value );
 			}
 
+			json_decode( $meta_value );
+			// Adds slashes to a string or recursively adds slashes to strings within an array for the JSON
+			if ( json_last_error() === JSON_ERROR_NONE || '_elementor_data' === $meta_key ) {
+				$meta_value = \wp_slash( $meta_value );
+			}
+
 			if ( $has_prev_value ) {
 				update_post_meta( $post_id, $meta_key, $meta_value, $prev_value );
 			} else {
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -195,18 +195,12 @@ function set_meta( $post_id, $meta ) {
 				$prev_value = maybe_unserialize( $existing_meta[ $meta_key ][ $meta_placement ] );
 			}
 
-			$is_json = false;
-
 			if ( ! is_array( $meta_value ) ) {
-				json_decode( $meta_value );
-				$is_json    = json_last_error() === 0 ? true : false;
 				$meta_value = maybe_unserialize( $meta_value );
 			}
 
-			// Adds slashes to a string or recursively adds slashes to strings within an array for the JSON
-			if ( function_exists( 'wp_slash' ) && ( $is_json || '_elementor_data' === $meta_key ) ) {
-				$meta_value = wp_slash( $meta_value );
-			}
+			// Adds slashes to a string or recursively adds slashes to strings within an array to avoid JSON decode failure while migrating metadata
+			$meta_value = wp_slash( $meta_value );
 
 			if ( $has_prev_value ) {
 				update_post_meta( $post_id, $meta_key, $meta_value, $prev_value );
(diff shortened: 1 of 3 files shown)


includes/utils.php
```diff
@@ -195,10 +195,19 @@ function set_meta( $post_id, $meta ) {
...
+			$is_json = false;
+
 			if ( ! is_array( $meta_value ) ) {
+				json_decode( $meta_value );
+				$is_json    = json_last_error() === 0 ? true : false;
 				$meta_value = maybe_unserialize( $meta_value );
 			}
 
+			// Adds slashes to a string or recursively adds slashes to strings within an array for the JSON
+			if ( function_exists( 'wp_slash' ) && ( $is_json || '_elementor_data' === $meta_key ) ) {
```
dkotter: I don't think we need the `function_exists` check here, as that function has been in WordPress since 3.6.0

includes/utils.php
```diff
@@ -195,10 +195,19 @@ function set_meta( $post_id, $meta ) {
 				$prev_value = maybe_unserialize( $existing_meta[ $meta_key ][ $meta_placement ] );
 			}
 
+			$is_json = false;
+
 			if ( ! is_array( $meta_value ) ) {
+				json_decode( $meta_value );
+				$is_json    = json_last_error() === 0 ? true : false;
```
dkotter: I get we are trying to determine here if the value is json, just wondering if there's any edge cases that this might have issues with? For instance, seems like the value `true` will pass as valid json (which may be fine).

Also want to ensure we won't ever have PHP errors by trying to run `json_decode` on data that can't be decoded. I know passing an `array` to that will throw an error, though that isn't a risk here because of the `if` statement. But are there any other types of data that might cause problems here?

dkotter: Also might be worth having a code comment here to explain what we're doing. I could see this confusing future developers trying to understand what's going on

includes/utils.php
dkotter: Also, wonder if we should introduce a new filter here that could be used to turn slashing on? I'm sure there's other types of data we aren't accounting for here that could use slashing and don't really want to have to manually support all those use cases (like we're supporting `_elementor_data` here). So adding a filter would allow others to hook in and turn this on as needed

dkotter (changes requested):

peterwilsoncc:
I may be missing something but I think the solution here is to slash all meta data in the call to `update_post_meta()`, this is similar to the approach taken in the rest api. ([ref 1](https://github.com/WordPress/wordpress-develop/blob/30ff995eb0aa2660ef4b08b9e3c5f7b0126c702b/src/wp-includes/rest-api/fields/class-wp-rest-meta-fields.php#L391-L392) and [ref 2](https://github.com/WordPress/wordpress-develop/blob/30ff995eb0aa2660ef4b08b9e3c5f7b0126c702b/src/wp-includes/rest-api/fields/class-wp-rest-meta-fields.php#L340))

includes/utils.php
dkotter: Or as [Peter mentions](https://github.com/10up/distributor/pull/888#issuecomment-1156007012), could just slash all meta which would simplify things here quite a bit

sksaju: Yeah! Thanks, I'll update the PR this week :)

sksaju: > I don't think we need the `function_exists` check here, as that function has been in WordPress since 3.6.0

@dkotter I've removed the check but now PHPUnit says undefined function! here https://github.com/10up/distributor/runs/6985524501?check_suite_focus=true 
I've also tried with `\wp_slash` to avoid namespace issue but couldn't fix

peterwilsoncc: @sksaju To fix the tests, you'll need to mock the `wp_slash` function with an `\WP_Mock::userFunction( 'wp_slash', $args_array )` in the test suite. 

I've not used the mocking library yet but will try to figure out what the args array should be.

peterwilsoncc: I've added mocks for `wp_slash()` in 4f88fff9f9e35e866b302a1ce977146ad8aaeab3 which seems to fix the tests.

@dkotter Would you mind validating my use of the mocks?

includes/utils.php
```diff
@@ -199,6 +199,9 @@ function set_meta( $post_id, $meta ) {
 				$meta_value = maybe_unserialize( $meta_value );
 			}
 
+			// Adds slashes to a string or recursively adds slashes to strings within an array to avoid JSON decode failure while migrating metadata
+			$meta_value = wp_slash( $meta_value );
```
peterwilsoncc: Both the meta key and the meta value need to be slashed. 

To be more defensive with the code, I suggest slashing the data within the function calls, ie

```php
update_post_meta( $post_id, wp_slash( $meta_key ), wp_slash( $meta_value ), $prev_value );
// and
add_post_meta( $post_id, wp_slash( $meta_key ), wp_slash( $meta_value ) );
```

Although slashing the data at the moment is safe, escaping later protects from future mistakes if data that requires slashing is added between here and the calls to the core functions.

sksaju: Thank you so much for your feedback @peterwilsoncc, I've updated the PR

includes/utils.php
dkotter: > I've added mocks for `wp_slash()` in [4f88fff](https://github.com/10up/distributor/commit/4f88fff9f9e35e866b302a1ce977146ad8aaeab3) which seems to fix the tests.
> 
> @dkotter Would you mind validating my use of the mocks?

This looks good. If wanted, you could set these up as [passthru functions](https://github.com/10up/wp_mock#passthru-functions), since in essence that's what we're doing here. But this is equivalent to that anyway.

peterwilsoncc (dismissed):
Thanks for following up, this all looks good to me.

dkotter (approved):

PR #890, opened 2022-06-08
Fix issue 889 - Unicode characters not escaped

### LINKED ISSUE #889: Unicode characters are not escaped during syndication
### Describe the bug

During syndication, both pull and push, if a WordPress Gutenberg block enabled post content has Unicode characters in block attributes, I expect the Unicode characters to be slashed so that WordPress does not remove the slashes when pulling the syndicated posts. At the moment, WordPress strips out the slashes, so it creates invalid text strings in the post content of the syndicated post.

### Steps to Reproduce

1. Create a custom Gutenberg block with the following attributes title, description. 
2. Add some test content and save it so that the post content is now available in the database with the title and description attributes.
3. Replace the attributes with following code snippet which has encoded content in the title and description attributes of the post content.
4. Make the post available for syndication
5. Pull the post on the multisite.
6. WordPress strips of the slashes creating the character encoding issue as shown in the screenshot below.

image

```
 {"title":"Our 11 Objectives","description":"\u003cstrong\u003eEnvironment\u003c/strong\u003e \u003cbr\u003e\u003cbr\u003e\u003cstrong\u003eObjective 01\u003c/strong\u003e \u003cbr\u003eBe carbon neutral for our own operations no later than 2025 and across our supply chain by 2030, and committed to becoming a net zero business well ahead of 2050\u003cbr\u003e\u003cbr\u003e\u003cstrong\u003eObjective 02\u003c/strong\u003e \u003cbr\u003eTransition to 100% renewable electricity by 2025, and electric vehicles by 2030\u003cbr\u003e\u003cbr\u003e\u003cstrong\u003eObjective 03 \u003c/strong\u003e\u003cbr\u003eHelp our clients to save 10m tons of Co2 eq by 2030\u003cbr\u003e\u003cbr\u003e\u003ca Find out more on environmental sustainability\u003c/a\u003e"}

```

### Environment information

_No response_

### WordPress information

WordPress is managed by the WordPress VIP

### Code of Conduct

amalajith:
### Description of the Change

Add slashes to the post content to fixe the issue with Unicode characters in block attributes not escaped and getting stripped out during syndication

Closes #889 

### Verification Process

After adding the `wp_slash()`, the syndicated content has properly escaped slashes in the Unicode characters and the content  is now properly rendered. 

image

### Checklist:

### Credits

Props @


includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -86,7 +86,7 @@ public function push( $post_id, $args = array() ) {
 		// Distribute raw HTML when going from Gutenberg enabled to Gutenberg enabled.
 		$remote_using_gutenberg = \Distributor\Utils\is_using_gutenberg( $post );
 		if ( $using_gutenberg && $remote_using_gutenberg ) {
-			$new_post_args['post_content'] = $post->post_content;
+			$new_post_args['post_content'] = wp_slash($post->post_content);
```
dkotter:
```suggestion
			$new_post_args['post_content'] = wp_slash( $post->post_content );
```

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -263,6 +263,10 @@ public function pull( $items ) {
 			if ( ! empty( $item_array['post_status'] ) ) {
 				$post_array['post_status'] = $item_array['post_status'];
 			}
+			
+			if( ! empty($post_array['post_content'])){
+				$post_array['post_content'] = wp_slash($post_array['post_content']);
```
dkotter:
```suggestion
			if ( ! empty( $post_array['post_content'] ) ) {
				$post_array['post_content'] = wp_slash( $post_array['post_content'] );
```

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -263,6 +263,10 @@ public function pull( $items ) {
 			if ( ! empty( $item_array['post_status'] ) ) {
 				$post_array['post_status'] = $item_array['post_status'];
 			}
+
```
dkotter:
```suggestion

```

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -263,6 +263,9 @@ public function pull( $items ) {
 			if ( ! empty( $item_array['post_status'] ) ) {
 				$post_array['post_status'] = $item_array['post_status'];
 			}
+			if ( ! empty( $post_array['post_content'] ) ) {
```
dkotter:
```suggestion

			if ( ! empty( $post_array['post_content'] ) ) {
```

dkotter:
@amalajith Thanks for the PR! I made a few formatting changes but otherwise looks good.

But a couple things that may improve this a bit:

1. Thinking that instead of slashing the `post_content` in both the `push` and `pull` methods, might be better to handle that in the [`prepare_post` utility](https://github.com/10up/distributor/blob/1.6.9/includes/utils.php#L1045). This would avoid duplicate code
2. The fix here only applies to Internal connections. We should probably fix this for External connections as well. Is that something you can add to this PR? Looks like a few places we would have to add that: https://github.com/10up/distributor/blob/1.6.9/includes/classes/ExternalConnections/WordPressExternalConnection.php#L701, https://github.com/10up/distributor/blob/1.6.9/includes/classes/ExternalConnections/WordPressExternalConnection.php#L1053 and https://github.com/10up/distributor/blob/1.6.9/includes/utils.php#L987
3. Is there any concern with slashing the `post_content`? I know that solves issues with things like unicode but just wondering if this could introduce bugs in other scenarios?

cadic (commented):
Just wondering, should we also wrap post_content with wp_slash for non-gutenberg content?

https://github.com/10up/distributor/blob/aa36d6c8dddfa4f8bc25dfef9d18a8dc81c41967/includes/classes/InternalConnections/NetworkSiteConnection.php#L72-L80

dkotter:
> Just wondering, should we also wrap post_content with wp_slash for non-gutenberg content?

Yeah, I think having it for both is probably the way to go, as long as that doesn't have any unintended consequences. And as mentioned in my [comment](https://github.com/10up/distributor/pull/890#issuecomment-1151691248), if we move this code to the `prepare_post` utility method, that will take care of both Push and Pull and both block and non-block content. The one thing I just noticed here though is that the Internal Push method runs content through `get_processed_content`, which External connections use as well (and External connections don't use the `prepare_post` method). So if we add slashing to the `get_processed_content` method to solve External connections, we will end up slashing content twice for Internal pushes, so we'll need to handle that.

peterwilsoncc:
> > Just wondering, should we also wrap post_content with wp_slash for non-gutenberg content?
> 
> Yeah, I think having it for both is probably the way to go, as long as that doesn't have any unintended consequences. 

It shouldn't have any consequences, WP expects the following to be slashed and unslashes it within `wp_insert_post()`:
* post_author
* post_date
* post_date_gmt
* post_content
* post_content_filtered
* post_title
* post_excerpt
* post_status
* post_type
* comment_status
* ping_status
* post_password
* post_name
* to_ping
* pinged
* post_modified
* post_modified_gmt
* post_parent
* menu_order
* post_mime_type
* guid

[see source code](https://github.com/WordPress/wordpress-develop/blob/ac3de7c5b3c6eeeda485e00cc0d94da5b40e023f/src/wp-includes/post.php#L4331-L4377)

I wasn't able to reproduce the bug using external connections, so either the data is being slashed there already or this bug doesn't affect both connection types.

I was using this mini-plugin for testing -- it needs to be active on all sites & creates a test post on each site [test-block.zip](https://github.com/10up/distributor/files/9080988/test-block.zip)

peterwilsoncc:
I've done some additional testing locally.

I think I'll move the slashing to include the entire post object rather than just the content. This will ensure the bug is fixed in it's entirety rather than risk the same bug appearing in other fields, such as title.

dkotter (approved):

PR #895, opened 2022-06-17
Prevent fatal caused by filter name clash.

### LINKED ISSUE #879: Conflicting dynamic pre_post_link filter can result in a fatal PHP error
### Describe the bug

`pre_post_link` is a WordPress core filter used to [filter the permalink structure for a post before token replacement occurs](https://developer.wordpress.org/reference/hooks/pre_post_link/).

It provides the `$permalink` string, `$post` object, and `$leavename` boolean when fired within `get_permalink()` in WordPress core.

One of the (I believe custom) arguments [that Distributor passes](https://github.com/10up/distributor/blob/6d20a4ce5f7cea4472d48b8f30cab1b57cacaa5e/includes/classes/ExternalConnections/WordPressExternalConnection.php#L562) to `wp_insert_post()` when pulling content from an external connection is `link`, which contains the URL of the external content being pulled and is stored as post meta.

When `wp_insert_post()` fires, all post arguments are sanitized through `sanitize_post()`. In `sanitize_post()`, each field is processed through `sanitize_post_field()`. In `sanitize_post_field()`, non-`post_`-prefixed fields are [passed through a dynamic filter](https://github.com/WordPress/wordpress-develop/blob/5612331504e22efe3e17440580d887279f68913e/src/wp-includes/post.php#L2830), `pre_post_{$field}`.

In the case of the `link` argument used by Distributor, this becomes `pre_post_link`, which conflicts with the filter of the same name in WordPress core.

Other code filtering `pre_post_link` must now defend against the possibility that only one argument is passed, resulting in a fatal error if the function signature expects 2 or 3. It must also be cautious that the `$link` value is formatted like `https://example.org/slug/` rather than a permalink structure like `/%category%/%postname%/`.

It's disappointing that WordPress allows this conflict to happen, but it seems like something that can be worked around through code. It would make sense to me to change the custom post argument to something like `dt_original_post_url` or `dt_link` so that it's prefixed and/or matches the meta key it's stored with. If this can't happen [early in the process](https://github.com/10up/distributor/blob/b11ac0f6183fffaf10ccf09fda6875aa4552f0bf/includes/classes/ExternalConnections/WordPressExternalConnection.php#L1047), then maybe it can be adjusted just in time before passing to `wp_insert_post()`.

### Steps to Reproduce

Add a filter like:

```
add_filter( 'pre_post_link', 'jf_filter_post_link', 10, 3 );

function jf_filter_post_link( $permalink, $post, $leavename ) {
  // do things...

  return $permalink;
}
```

Then attempt to pull a piece of external content via Distributor -> Pull Content in the dashboard.

_No response_

### Environment information

N/A

### WordPress information

WP 5.9.3 with a lot of plugins and things going on. :)

### Code of Conduct

peterwilsoncc:
### Description of the Change

This change ensures only arguments accepted by `wp_insert_post()` are passed to the function during a push or pull distribution of content.

This addresses a bug in which the dynamic hooks fired within the function call were clashing with other WP hooks and causing a fatal error. 

Closes #879

### Alternate Designs

I considered using the hooks `dt_pull_post_args` and `dt_push_post_args` to reduce the items to the allow list but discovered the latter filter is also used for making an HTTP request when pushing a post so could not be used in this fashion.

### Possible Drawbacks

Breaks plugins expecting _unofficial_ parameters be passed to `wp_insert_post()`. I've done a review of the function to see if there are any hooks that pass both the post ID and the unmodified `$postarr` and can't see anything.

### Verification Process

WIP

### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/10up/.github/blob/trunk/CONTRIBUTING.md) document.

### Changelog Entry

Fixed: Prevent conflict with `pre_post_link` filter.

### Credits

Props @jeremyfelt, @peterwilsoncc

### DIFF
--- a/includes/classes/ExternalConnections/WordPressExternalConnection.php
+++ b/includes/classes/ExternalConnections/WordPressExternalConnection.php
@@ -559,7 +559,8 @@ class WordPressExternalConnection extends ExternalConnection {
 			 *
 			 * @return {array} The post data to be inserted.
 			 */
-			$new_post = wp_insert_post( apply_filters( 'dt_pull_post_args', $post_array, $item_array['remote_post_id'], $post, $this ) );
+			$new_post_args = Utils\post_args_allow_list( apply_filters( 'dt_pull_post_args', $post_array, $item_array['remote_post_id'], $post, $this ) );
+			$new_post      = wp_insert_post( $new_post_args );
 
 			update_post_meta( $new_post, 'dt_original_post_id', (int) $item_array['remote_post_id'] );
 			update_post_meta( $new_post, 'dt_original_source_id', (int) $this->id );
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -111,7 +111,8 @@ class NetworkSiteConnection extends Connection {
 
 		add_filter( 'wp_insert_post_data', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'maybe_set_modified_date' ), 10, 2 );
 		// Filter documented in includes/classes/ExternalConnections/WordPressExternalConnection.php
-		$new_post_id = wp_insert_post( apply_filters( 'dt_push_post_args', $new_post_args, $post, $args, $this ) );
+		$new_post_args = Utils\post_args_allow_list( apply_filters( 'dt_push_post_args', $new_post_args, $post, $args, $this ) );
+		$new_post_id   = wp_insert_post( $new_post_args );
 
 		remove_filter( 'wp_insert_post_data', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'maybe_set_modified_date' ), 10, 2 );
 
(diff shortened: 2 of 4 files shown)


peterwilsoncc:
Needs some tests

- the `pre_post_link` filter does not run
- push/pull posts contain the `link` & other meta data
- terms are assigned

peterwilsoncc:
> Needs some tests
> 
> * the `pre_post_link` filter does not run
> * push/pull posts contain the `link` & other meta data
> * terms are assigned

Added a test for the basic utility function but will need some help with the other items. I am not sure the acceptance tests are running so it might not be worth adding them.

dinhtungdu (approved):
LGTM! Thanks for the PR and the test.

> I am not sure the acceptance tests are running so it might not be worth adding them.

You're right! WP Acceptance is not working 😅

jeffpaul:
Noting that yes we should look to replace WP Acceptance with Cypress e2e tests here, I'll look to see that we have an issue for that with at least some basic coverage in an upcoming release.

PR #898, opened 2022-06-21
Enhancement/860 Removed Built-in post types for External Connections + Added a new filter.

### LINKED ISSUE #860: Trim down the list of default supported post types for External Connections

After setting up an external connection, you see a table of post types and the permissions for those post types. We default to showing all post types that have `show_in_rest` set to `true`.

On more recent versions of WordPress, this shows post types like Navigation Menus and Templates (among others) that we probably don't need to show, as we can't Push or Pull those.

There's also not a filter around these post types, so there's no way for others to change this list if they want. I propose we remove any default post types that will never be Pushed or Pulled and add a new filter around this so others can change this list before it gets output.

Screen Shot 2022-03-08 at 4 05 42 PM

### Code of Conduct

faisal-alvi:
### Description of the Change

- This PR trims down the list of post types coming with a WordPress setup which we can not Push or Pull.
- Also introduces a new filter `dt_excluded_post_types_from_permission_list` so users can also specify/add post types which they want to hide.

Closes #860

### Alternate Designs

_before_

image
[image link](https://www.screencast.com/t/9ycSFb6fCMm6)

_after_

image
[image link](https://www.screencast.com/t/9bM9QQRUAz)

### Possible Drawbacks

does not seem any.

### Verification Process

1. Create a new connection or edit an existing one.
2. See the list under the “Post types permissions” title.
3. You won’t find any of these:
    1. Media
    2. Navigation Menu Items
    3. Reusable blocks
    4. Templates
    5. Templates Parts
    6. Navigation Menus
4. Test the filter now, and add the following code in your theme’s `functions.php` file.

```php
add_filter('dt_excluded_post_types_from_permission_list', 'dt_excluded_post_types_from_permission_list', 10, 1);
function dt_excluded_post_types_from_permission_list( $list ) {
    $hide_from_list[] = 'post';
    return $hide_from_list;
}
```
5. Re-check the list under the “Post types permissions” title.
6. You will see the “Posts” is not listed anymore.
7. Try to add/modify more post types in the filter code above, and see if those should not be listed as well.

### Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/10up/.github/blob/trunk/CONTRIBUTING.md) document.

### Changelog Entry

> Changed - Removed Built-in post types for External Connections.
> Added - Added a new filter `dt_excluded_post_types_from_permission_list ` to add post types that should not be listed under “Post types permissions”.

### Credits

Props @dkotter @faisal-alvi @peterwilsoncc @dinhtungdu

### DIFF
--- a/includes/external-connection-cpt.php
+++ b/includes/external-connection-cpt.php
@@ -539,28 +539,32 @@ function meta_box_external_connection_details( $post ) {
 					<th><?php esc_html_e( 'Can push?', 'distributor' ); ?></th>
 				</thead>
 				<tbody>
-					<?php foreach ( $post_types as $post_type ) : ?>
+					<?php
+					// Hide the built-in post types except 'post' and 'page'.
+					$hide_from_list = get_post_types( [ '_builtin' => true, 'show_in_rest' => true ] );
+					unset( $hide_from_list['post'], $hide_from_list['page'] );
+					$hide_from_list = array_values( $hide_from_list ); // Default is keyed by the post type 'post' => 'post', etc
+
+
+					// Hide 'dt_subscription' post type.
+					$hide_from_list[] = 'dt_subscription';
+
+					/**
+					 * Filter to update the list of post types that should be hidden from the "Post types permissions" list.
+					 *
+					 * @since x.x.x
+					 * @hook dt_hide_post_types_from_permission_list
+					 *
+					 * @param {array} The list of post types..
+					 *
+					 * @return {bool} The updated array with the list of post types that should be hidden.
+					 */
+					$hide_from_list = apply_filters( 'dt_hide_post_types_from_permission_list', $hide_from_list );
+
+					foreach ( $post_types as $post_type ) : ?>
 						<?php
-						/**
-						 * Filter to update the list of post types that should be hidden from the "Post types permissions" list.
-						 *
-						 * @since x.x.x
-						 * @hook dt_hide_post_types_from_permission_list
-						 *
-						 * @param {array} The list of post types..
-						 *
-						 * @return {bool} The updated array with the list of post types that should be hidden.
-						 */
-						$hide_from_list = apply_filters( 'dt_hide_post_types_from_permission_list', array() );
-
-						// Strictly hide 'dt_subscription' post type.
-						$hide_from_list[] = 'dt_subscription';
-
-						// Don't show:
-						// 1. Post types returning from a filter and 'dt_subscription'.
-						// 2. All _builtin post types except 'post' and 'page'.
-						if ( in_array( $post_type->name, $hide_from_list, true )
-							 || ( $post_type->_builtin && 'post' !== $post_type->name && 'page' !== $post_type->name ) ) {
+						// Don't show the given post types.
+						if ( in_array( $post_type->name, $hide_from_list, true ) ) {
 							continue;
 						}
 						?>
--- a/includes/external-connection-cpt.php
+++ b/includes/external-connection-cpt.php
@@ -561,7 +561,7 @@ function meta_box_external_connection_details( $post ) {
 					 * @since x.x.x
 					 * @hook dt_hide_post_types_from_permission_list
 					 *
-					 * @param {array} The list of post types..
+					 * @param {array} The list of hidden post types.
 					 *
 					 * @return {bool} The updated array with the list of post types that should be hidden.
 					 */
--- a/includes/external-connection-cpt.php
+++ b/includes/external-connection-cpt.php
@@ -544,7 +544,7 @@ function meta_box_external_connection_details( $post ) {
 					$hide_from_list = get_post_types(
 						array(
 							'_builtin'     => true,
-							'show_in_rest' => true
+							'show_in_rest' => true,
 						)
 					);
 					unset( $hide_from_list['post'], $hide_from_list['page'] );
--- a/includes/external-connection-cpt.php
+++ b/includes/external-connection-cpt.php
@@ -541,7 +541,26 @@ function meta_box_external_connection_details( $post ) {
 				<tbody>
 					<?php foreach ( $post_types as $post_type ) : ?>
 						<?php
-						if ( 'dt_subscription' === $post_type->name ) {
+						/**
+						 * Filter to update the list of post types that should be hidden from the "Post types permissions" list.
+						 *
+						 * @since x.x.x
+						 * @hook dt_hide_post_types_from_permission_list
+						 *
+						 * @param {array} The list of post types..
+						 *
+						 * @return {bool} The updated array with the list of post types that should be hidden.
+						 */
+						$hide_from_list = apply_filters( 'dt_hide_post_types_from_permission_list', array() );
+
+						// Strictly hide 'dt_subscription' post type.
+						$hide_from_list[] = 'dt_subscription';
+
+						// Don't show:
+						// 1. Post types returning from a filter and 'dt_subscription'.
+						// 2. All _builtin post types except 'post' and 'page'.
+						if ( in_array( $post_type->name, $hide_from_list, true )
+							 || ( $post_type->_builtin && 'post' !== $post_type->name && 'page' !== $post_type->name ) ) {
 							continue;
 						}
 						?>
--- a/includes/external-connection-cpt.php
+++ b/includes/external-connection-cpt.php
@@ -540,32 +540,7 @@ function meta_box_external_connection_details( $post ) {
 				</thead>
 				<tbody>
 					<?php
-					// Hide the built-in post types except 'post' and 'page'.
-					$hide_from_list = get_post_types(
-						array(
-							'_builtin'     => true,
-							'show_in_rest' => true,
-						)
-					);
-					unset( $hide_from_list['post'], $hide_from_list['page'] );
-
-					// Default is keyed by the post type 'post' => 'post', etc; hence using `array_values`.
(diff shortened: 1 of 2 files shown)


includes/external-connection-cpt.php
```diff
@@ -541,7 +541,26 @@ function meta_box_external_connection_details( $post ) {
...
+						$hide_from_list = apply_filters( 'dt_hide_post_types_from_permission_list', array() );
+
+						// Strictly hide 'dt_subscription' post type.
+						$hide_from_list[] = 'dt_subscription';
+
+						// Don't show:
+						// 1. Post types returning from a filter and 'dt_subscription'.
+						// 2. All _builtin post types except 'post' and 'page'.
+						if ( in_array( $post_type->name, $hide_from_list, true )
+							 || ( $post_type->_builtin && 'post' !== $post_type->name && 'page' !== $post_type->name ) ) {
```
peterwilsoncc: This has the effect of strictly hiding all built in post types. I can see a use case in which a site owner may wish to include resuable blocks or attachments in the list of types.

Instead, how about setting the default list prior to the filter:
```php
$hide_from_list = get_post_types( ['_builtin' => true, 'show_in_rest' => true] );
unset( $hide_from_list['post'], $hide_from_list['page'] );
$hide_from_list = array_values( $hide_from_list ); // Default is keyed by the post type 'post' => 'post', etc
```

This will make `dt_hide_post_types_from_permission_list` as flexible as possible.

peterwilsoncc (commented):
I've added a note about reordering the code to allow additional flexibility. It will also make the `if... continue` condition simpler.

Is it possible to add a test to ensure the list is correct?

includes/external-connection-cpt.php
dinhtungdu: > This will make `dt_hide_post_types_from_permission_list` as flexible as possible.

I agree. Also, I think we should create a utility for this task. Putting all ~login~ logics inside HTML code (which is inside some `foreach` loops) makes it a bit hard to read IMO.

faisal-alvi: > This has the effect of strictly hiding all built in post types. I can see a use case in which a site owner may wish to include resuable blocks or attachments in the list of types.

@peterwilsoncc that is what I thought in the beginning but even if a site owner displays built-in post types in the list, those post types can't be pulled/pushed anyways. As per @dkotter's https://github.com/10up/distributor/issues/860#issue-1163273171:

> On more recent versions of WordPress, this shows post types like Navigation Menus and Templates (among others) that we probably don't need to show, as we can't Push or Pull those.

So there is no point to show those post types in the list, isn't it? Can you please share a use case why a site owner would want to include built-in post types even when he/she can't pull/push them?

faisal-alvi: > I agree. Also, I think we should create a utility for this task. Putting all login inside HTML code (which is inside some foreach loops) makes it a bit hard to read IMO.

@dinhtungdu excuse me, can you please elaborate more? are you suggesting moving the code to a separate file or something?

dinhtungdu: > > I agree. Also, I think we should create a utility for this task. Putting all login inside HTML code (which is inside some foreach loops) makes it a bit hard to read IMO.
> 
> @dinhtungdu excuse me, can you please elaborate more? are you suggesting moving the code to a separate file or something?

Sorry, it's my typo, I was saying logics, not login 😂 (edited). I suggested extracting this logic into a utility function, something like `is_post_type_allowed()`.

But looking at the existing utilities, we already have `distributable_post_types()`. Do you think we can use that utility?

https://github.com/10up/distributor/blob/dcf23f33a0cd80bcf3be30f74066b291e80dbf62/includes/utils.php#L282-L300

faisal-alvi: @dinhtungdu that's interesting! Only the difference is we are using `show_in_rest => true`,

https://github.com/10up/distributor/blob/dcf23f33a0cd80bcf3be30f74066b291e80dbf62/includes/external-connection-cpt.php#L452-L457

And the utility function `distributable_post_types()` uses `public => true`.

https://github.com/10up/distributor/blob/dcf23f33a0cd80bcf3be30f74066b291e80dbf62/includes/utils.php#L283

Any reason why you have used `show_in_rest => true` (https://github.com/10up/distributor/commit/13614204987df5f85459a37236404cca0b8e6c26)? if you think we can use `public => true` instead, then IMO, it's fine to use the utility function. In that case, we also do not need to introduce a new filter with this PR.

dkotter: > > This has the effect of strictly hiding all built in post types. I can see a use case in which a site owner may wish to include resuable blocks or attachments in the list of types.
> 
> @peterwilsoncc that is what I thought in the beginning but even if a site owner displays built-in post types in the list, those post types can't be pulled/pushed anyways. As per @dkotter's [#860 (comment)](https://github.com/10up/distributor/issues/860#issue-1163273171):
> 
> > On more recent versions of WordPress, this shows post types like Navigation Menus and Templates (among others) that we probably don't need to show, as we can't Push or Pull those.
> 
> So there is no point to show those post types in the list, isn't it? Can you please share a use case why a site owner would want to include built-in post types even when he/she can't pull/push them?

While it's true that we don't support out of the box pushing or pulling of those content types, some sites may want to add support for those themselves, utilizing hooks. For instance, I know we've had reports of sites that added support for attachments and reusable blocks with existing hooks. So while my opinion is it doesn't make sense to show those in that external connection screen for most sites (since those aren't things you can distribute without doing custom work) we still want to keep this as flexible as possible to support all of those potential use cases.

faisal-alvi: Thanks for the explanation @dkotter, super clear now! I will update the code accordingly.

faisal-alvi:
> Is it possible to add a test to ensure the list is correct?

@peterwilsoncc  IMO, the test is not required here. Also, no tests are written for the function `meta_box_external_connection_details()` which contains the code of this PR. So probably we should think about writing the test for the whole function in a separate ticket.

includes/external-connection-cpt.php
```diff
@@ -539,9 +539,36 @@ function meta_box_external_connection_details( $post ) {
...
+					// Hide 'dt_subscription' post type.
+					$hide_from_list[] = 'dt_subscription';
+
+					/**
+					 * Filter to update the list of post types that should be hidden from the "Post types permissions" list.
+					 *
+					 * @since x.x.x
+					 * @hook dt_hide_post_types_from_permission_list
+					 *
+					 * @param {array} The list of post types..
```
peterwilsoncc:
```suggestion
					 * @param {array} The list of hidden post types.
```

Nitpick. I'll merge to avoid delaying approval.

peterwilsoncc (dismissed):
LGTM, thank you.

includes/external-connection-cpt.php
```diff
@@ -539,9 +539,36 @@ function meta_box_external_connection_details( $post ) {
...
+					 * Filter to update the list of post types that should be hidden from the "Post types permissions" list.
+					 *
+					 * @since x.x.x
+					 * @hook dt_hide_post_types_from_permission_list
+					 *
+					 * @param {array} The list of hidden post types.
+					 *
+					 * @return {bool} The updated array with the list of post types that should be hidden.
+					 */
+					$hide_from_list = apply_filters( 'dt_hide_post_types_from_permission_list', $hide_from_list );
```
dinhtungdu: [Optional] I still think we should extract this logic into a utility function to make it easier to read, something like `get_excluded_post_types_from permission_list()`.

faisal-alvi: Agree, implemented!

includes/utils.php
```diff
@@ -299,6 +299,41 @@ function distributable_post_types() {
...
+	 * Filter to update the list of post types that should be hidden from the "Post types permissions" list.
+	 *
+	 * @since 1.7.0
+	 * @hook dt_hide_post_types_from_permission_list
+	 *
+	 * @param {array} The list of hidden post types.
+	 *
+	 * @return {bool} The updated array with the list of post types that should be hidden.
+	 */
+	return apply_filters( 'dt_hide_post_types_from_permission_list', $hide_from_list );
```
dinhtungdu: Do you think we should update this filter name to match with the function name, or vice versa, update the function name to match the filter name?

dinhtungdu (dismissed):
Thanks for the update! I have a minor comment around the naming, otherwise, this is LGTM so pre-approving.

includes/utils.php
peterwilsoncc: Consistency is nice. I think renaming the the filter `dt_excluded_post_types_from_permission_list` would be good.

dinhtungdu (dismissed):
Thanks so much for the updates! :shipit:

peterwilsoncc (dismissed):
This looks good to me.

I do wonder if `dt_subscription` should be force hidden (ie, re-added to the hide array prior to the `return` but not a blocker.

dinhtungdu:
> I do wonder if dt_subscription should be force hidden (ie, re-added to the hide array prior to the return but not a blocker.

Make sense to me 👍🏽

faisal-alvi:
@peterwilsoncc @dinhtungdu thanks so much for wonderful suggestions. Please let me know if any changes are required. The last one implemented in: https://github.com/10up/distributor/pull/898/commits/b21c4e4b7679471deb19c2f86dc4570ca4bc1841

dkotter (approved):

PR #901, opened 2022-06-27
Meet VIP Coding Standards (alternative approach)

### LINKED ISSUE #799: Enhancement: Support WP VIP Code Standards

peterwilsoncc:
### Description of the Change

Alternative approach to PR #807. 

This uses the WP built in sanitization functions rather than `filter_input()`. To avoid false positives I have:

* created a custom sanitization function for verifying connections -- these don't work in namespaces so it is in the global, unfortunately.
* created a wrapper for both `wp_remote_get()` and `vip_safe_wp_remote_get()` to reduce false positives for the VIP sniffs reporting we were using the former
* added `wp_verify_nonce()` as unslashing and sanitizing function because it doesn't need either (opinionated)
* added `wp_trim_words()` as a santitizing function as it effectively is (also opinionated)

Closes #799 

The initial sniff report (running on develop) is  [sniffs.txt](https://github.com/10up/distributor/files/8988829/sniffs.txt) -- this reports on the combined 10up/WPVIP coding standards.

### Alternate Designs

See #807. 

### Possible Drawbacks

Incorrect sanitization: I am testing the sanitization out as I go but please check I haven't made anything too strict. 

### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/10up/.github/blob/trunk/CONTRIBUTING.md) document.

### Credits

Props @mehulkaklotar, @peterwilsoncc, @dkotter, @dinhtungdu


dinhtungdu (commented):
@peterwilsoncc this is looking good to me. `Utils\remote_http_get` makes a lot of sense to me 💯.

I have some questions about your PR.
- I run `phpcs` and got this report, is that expected?

<details>
<summary>`phpcs` report</summary>

```
% ./vendor/bin/phpcs includes templates distributor.php --standard="WordPress-VIP-Go" 

FILE: ...up/Sites/mu/app/public/wp-content/plugins/distributor/includes/push-ui.php
--------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
--------------------------------------------------------------------------------
 241 | WARNING | Detected usage of a non-sanitized input variable:
     |         | $_POST['connections']
--------------------------------------------------------------------------------


FILE: ...butor/includes/classes/ExternalConnections/WordPressExternalConnection.php
--------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
--------------------------------------------------------------------------------
 172 | WARNING | Having more than 100 posts returned per page may lead to
     |         | severe performance problems.
--------------------------------------------------------------------------------


FILE: ...p/Sites/mu/app/public/wp-content/plugins/distributor/includes/rest-api.php
--------------------------------------------------------------------------------
FOUND 0 ERRORS AND 3 WARNINGS AFFECTING 3 LINES
--------------------------------------------------------------------------------
 183 | WARNING | Having more than 100 posts returned per page may lead to
     |         | severe performance problems.
 416 | WARNING | Having more than 100 posts returned per page may lead to
     |         | severe performance problems.
 529 | WARNING | This comment is 39% valid code; is this commented out code?
--------------------------------------------------------------------------------


FILE: ...public/wp-content/plugins/distributor/includes/external-connection-cpt.php
--------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
--------------------------------------------------------------------------------
 363 | WARNING | Detected usage of a non-sanitized input variable:
     |         | $_POST['dt_external_connection_details']
--------------------------------------------------------------------------------


FILE: .../app/public/wp-content/plugins/distributor/includes/syndicated-post-ui.php
--------------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
--------------------------------------------------------------------------------
 289 | WARNING | Detected usage of a non-sanitized input variable:
     |         | $_GET['_wpnonce']
 336 | WARNING | Detected usage of a non-sanitized input variable:
     |         | $_GET['_wpnonce']
--------------------------------------------------------------------------------

Time: 10.9 secs; Memory: 18MB

```

</details>

- Do you think we should add a composer script to lint the codebase against VIP WPCS.
- In addition to the composer script, we can also set up a GitHub Actions workflow for VIP WPCS like our Apple Maps Block [does](https://github.com/10up/maps-block-apple/blob/develop/.github/workflows/linting.yml). What do you think?

includes/external-connection-cpt.php
```diff
@@ -165,7 +167,10 @@ function setup_list_table() {
 			$sendback = remove_query_arg( array( 'trashed', 'untrashed', 'deleted', 'locked', 'ids' ), wp_get_referer() );
 
 			$deleted  = 0;
-			$post_ids = array_map( 'intval', $_REQUEST['post'] );
+			$post_ids = array();
+			if ( ! empty( $_REQUEST['post'] ) ) {
+				$post_ids = array_map( 'intval', $_REQUEST['post'] );
+			}
```
dkotter, line 173: I'm fine with this but we could also make this a one-liner, though maybe it's harder to read?

```
$post_ids = ! empty( $_REQUEST['post'] ) ? array_map( 'intval', $_REQUEST['post'] ) : array();
```

includes/utils.php
```diff
@@ -1121,3 +1123,30 @@ function post_args_allow_list( $post_args ) {
...
+ * to the second parameter.
+ *
+ * See {@see http://developer.wordpress.org/reference/classes/WP_Http/request/ WP_Http::request} for $args defaults.
+ *
+ * @param  string $url       The URL to request.
+ * @param  array  $args      Optional. An array of arguments to pass to wp_remote_get()/vip_safe_wp_remote_get().
+ * @param  mixed  $fallback  Optional. Fallback value to return if the request fails. Default false. VIP only.
+ * @param  int    $threshold Optional. The number of fails required before subsequent requests automatically return the fallback value. Defaults to 3, with a maximum of 10. VIP only.
+ * @param  int    $timeout   Optional. The timeout for WP VIP requests. Default 3, 1-5 valid. VIP only, use $args for others.
+ * @param  int    $retries   Optional. The number of retries to attempt. Deafult 10. VIP only.
```
dkotter:
```suggestion
 * @param  int    $retries   Optional. The number of retries to attempt. Default 10. VIP only.
```

includes/utils.php
```diff
@@ -1121,3 +1123,30 @@ function post_args_allow_list( $post_args ) {
...
+ *
+ * Wrapper function for wp_remote_get() and vip_safe_wp_remote_get(). The order
+ * of parameters differs from vip_safe_wp_remote_get() to promote the arguments array
+ * to the second parameter.
+ *
+ * See {@see http://developer.wordpress.org/reference/classes/WP_Http/request/ WP_Http::request} for $args defaults.
+ *
+ * @param  string $url       The URL to request.
+ * @param  array  $args      Optional. An array of arguments to pass to wp_remote_get()/vip_safe_wp_remote_get().
+ * @param  mixed  $fallback  Optional. Fallback value to return if the request fails. Default false. VIP only.
```
dkotter: So I know we are setting this to `false` currently but in reading the [documentation](https://docs.wpvip.com/technical-references/code-quality-and-best-practices/retrieving-remote-data/#h-vip_safe_wp_remote_get), sounds like this is expected to be a string (default would be an empty string).

includes/utils.php
```diff
@@ -1121,3 +1123,30 @@ function post_args_allow_list( $post_args ) {
...
+ * @param  int    $threshold Optional. The number of fails required before subsequent requests automatically return the fallback value. Defaults to 3, with a maximum of 10. VIP only.
+ * @param  int    $timeout   Optional. The timeout for WP VIP requests. Default 3, 1-5 valid. VIP only, use $args for others.
```
dkotter: I believe only 1, 2 or 3 is valid here

includes/utils.php
dkotter: May be worth noting here that this value can't be less than 10, at least according to VIP's documentation (and they actually set the default to 20, may be worth matching that?)

includes/utils.php
peterwilsoncc: The time out maximum is complicated according to the code:
* 30 seconds for WP CLI `POST` requests
* 15 seconds for `POST` requests within the admin
* 5 seconds for front end and non POST requests.

I'll expand the notes.
https://github.com/Automattic/vip-go-mu-plugins/blob/fdd99cf535e55eeda46d7a11a4cbb1603d8bc666/vip-helpers/vip-utils.php#L816-L831

includes/utils.php
peterwilsoncc: > May be worth noting here that this value can't be less than 10, at least according to VIP's documentation (and they actually set the default to 20, may be worth matching that?)

I'll update the docs

All the instances in the code base set this to ten, thus the lower default. It may be worth changing it but let's do so in a follow up to keep the code as close to unchanged as possible.

includes/utils.php
peterwilsoncc: Fixed in 7b156bb1f4239aab8f6a9e5a171f6d7124ea0cce

peterwilsoncc:
> I run phpcs and got this report, is that expected?

Thanks @dinhtungdu 

Some of those appear to be conflicts between the 10up CS and the VIP CS. I'll check them out.

* `$_POST['connections']` being unsantized is expected as there is a custom function set in the phpcs.xml file
* The various nonces being unsanitized is expected as I've set `wp_verify_nonce()` as auto sanitized and unslashed as a valid nonce doesn't need either and an invalid nonce shouldn't be sanitized to a correct one. This is opinionated.

I've reached out to a contact at VIP to check if they respect phpcs files.

peterwilsoncc:
Sorry, some further changes following your comments above:

* 0778f7c78984dfe79c93c00d30530294a84bec19 -- renamed `remote_http_get()` to `remote_http_request()` as it works for any method and the plugin uses it for `POST` requests
* 4fee9e7e19ed29282c17abc377111b00469877a3 -- reworded a comment to avoid a false positive in Tung's report above.

dkotter (dismissed):

mehulkaklotar (commented):
@peterwilsoncc This looks good to me! Just to be sure, there are same VIPCS errors as @dinhtungdu got and it shows in the [lint check](https://github.com/10up/distributor/runs/7105270800?check_suite_focus=true) below as well. 
Can you take a look at that as well?

peterwilsoncc:
@dinhtungdu @mehulkaklotar As they're false positives, I think we're going to need to leave it be & remove the additional linting job from the GH actions.

As the vendor libraries also throw errors (also false positives), I think we'll need to accept close enough is good enough. 

I was reading the docs for WP VIP Go yesterday and they [allow customization of the PHPCS config](https://github.com/Automattic/vip-go-skeleton#phpcs-for-checking-coding-standards) these days. I suspect the additional flexibility is due to each site running on it's own container rather than as a custom theme on wordpress.com, as was the case prior to WP VIP Go.

peterwilsoncc:
I've removed the VIPCS check in 997b3a94e0b6fd09630f251a4aae1f848cd86cd9 due to the false positives.

Are folks able to take another look?

mehulkaklotar (approved):
Looks good! 🚀

PR #907, opened 2022-07-15
Fix/906 Distributable post types made consistent

### LINKED ISSUE #906: Consistently determine if a post type is distributable.

Currently Distributor determines whether a post type can be distributed based on a range of searches:

* `get_post_types( [ 'public' => true ] )`
* `get_post_types( [ 'show_in_rest' => true ] )`

In some cases they change based on the distribution type (push or pull); in some cases they change based on the connection type (internal or external).

### Designs

Ideally the search method would be consistent except for when technical considerations require a stricter search.

The most accurate search for post types visible on the front end of a site is `array_filter( get_post_types( [] ), 'is_post_type_viewable' );` -- this will take in to consideration the filter added in WP 5.9 and account for the fact that a post can be `public` but not public. 

As external connections require `show_in_rest` to be `true` for post distribution, the external connections have a further technical limitation and ought to use the search `array_filter( get_post_types( [ 'show_in_rest' => true ] ), 'is_post_type_viewable' );`

For the purpose of allowing sites to set their own policies, Distributor ought to retain the different pushable and pullable specific features.

While this is sorted out, it might be worth removing the `dt_excluded_post_types_from_permission_list` @faisal-alvi added recently from the 1.7.0 release -- just in case this results in the quick adding and deprecation of a function and filter.

_No response_

### Code of Conduct

faisal-alvi:
### Description of the Change

* This PR reverts the changes made in #898.
* Removes the function `get_excluded_post_types_from_permission_list`.
* Removes a filter `dt_excluded_post_types_from_permission_list`.
* Modifies the `distributable_post_types` function. Added a new parameter `$output`, which supports returning the result as per the given type of output.
* In that function, the following post types are excluded from the result by default.
```
                 'attachment',
		'nav_menu_item',
		'wp_block',
		'wp_navigation',
		'wp_template',
		'wp_template_part',
		'dt_ext_connection',
		'dt_subscription'
```

Closes #906

### How to test the Change

1. Create a new connection or edit an existing one.
2. See the list under the “Post types permissions” title.
3. You won’t find any of these:
    1. Media
    2. Navigation Menu Items
    3. Reusable blocks
    4. Templates
    5. Templates Parts
    6. Navigation Menus

### Credits

Props @faisal-alvi 

### Checklist:

- [ ] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).

### DIFF
--- a/includes/external-connection-cpt.php
+++ b/includes/external-connection-cpt.php
@@ -454,7 +454,7 @@ function meta_box_external_connection_details( $post ) {
 
 	$external_connection_status = get_post_meta( $post->ID, 'dt_external_connections', true );
 
-	$post_types = \Distributor\Utils\distributable_post_types( 'is_post_type_viewable' );
+	$post_types = \Distributor\Utils\distributable_post_types( 'objects' );
 
 	$registered_external_connection_types = \Distributor\Connections::factory()->get_registered();
 
--- a/includes/external-connection-cpt.php
+++ b/includes/external-connection-cpt.php
@@ -454,12 +454,7 @@ function meta_box_external_connection_details( $post ) {
 
 	$external_connection_status = get_post_meta( $post->ID, 'dt_external_connections', true );
 
-	$post_types = get_post_types(
-		array(
-			'show_in_rest' => true,
-		),
-		'objects'
-	);
+	$post_types = \Distributor\Utils\distributable_post_types( 'is_post_type_viewable' );
 
 	$registered_external_connection_types = \Distributor\Connections::factory()->get_registered();
 
@@ -544,20 +539,13 @@ function meta_box_external_connection_details( $post ) {
 					<th><?php esc_html_e( 'Can push?', 'distributor' ); ?></th>
 				</thead>
 				<tbody>
-					<?php
-					$hide_from_list = \Distributor\Utils\get_excluded_post_types_from_permission_list();
-
-					foreach ( $post_types as $post_type ) :
-						if ( in_array( $post_type->name, $hide_from_list, true ) ) {
-							continue;
-						}
-						?>
-						<tr>
-							<td><?php echo esc_html( $post_type->label ); ?></td>
-							<td><?php echo in_array( $post_type->name, $external_connection_status['can_get'] ) ? esc_html__( 'Yes', 'distributor' ) : esc_html__( 'No', 'distributor' ); ?></td>
-							<td><?php echo in_array( $post_type->name, $external_connection_status['can_post'] ) ? esc_html__( 'Yes', 'distributor' ) : esc_html__( 'No', 'distributor' ); ?></td>
-						</tr>
-					<?php endforeach; ?>
+				<?php foreach ( $post_types as $post_type ) : ?>
+					<tr>
+						<td><?php echo esc_html( $post_type->label ); ?></td>
+						<td><?php echo in_array( $post_type->name, $external_connection_status['can_get'] ) ? esc_html__( 'Yes', 'distributor' ) : esc_html__( 'No', 'distributor' ); ?></td>
+						<td><?php echo in_array( $post_type->name, $external_connection_status['can_post'] ) ? esc_html__( 'Yes', 'distributor' ) : esc_html__( 'No', 'distributor' ); ?></td>
+					</tr>
+				<?php endforeach; ?>
 				</tbody>
 			</table>
 		</div>
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -307,15 +307,10 @@ function available_pull_post_types( $connection, $type ) {
  * @return array
  */
 function distributable_post_types( $output = 'names' ) {
-	$post_types = get_post_types( [ 'show_in_rest' => true ], $output );
+	$post_types = array_filter( get_post_types(), 'is_post_type_viewable' );
 
 	$exclude_post_types = [
 		'attachment',
-		'nav_menu_item',
-		'wp_block',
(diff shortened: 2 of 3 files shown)


includes/utils.php
```diff
@@ -285,14 +285,28 @@ function available_pull_post_types( $connection, $type ) {
...
+ * @param string $output Optional. The type of output to return.
+ *                       Accepts post type 'names' or 'objects'. Default 'names'.
+ *
  * @since  1.0
  * @return array
  */
-function distributable_post_types() {
-	$post_types = get_post_types( [ 'public' => true ] );
+function distributable_post_types( $output = 'names' ) {
+	$post_types = get_post_types( [ 'show_in_rest' => true ], $output );
```
peterwilsoncc:
```suggestion
	$post_types = get_post_types( [ 'show_in_rest' => true ] );
	$post_types = array_filter( $post_types, 'is_post_type_viewable' );
```

This will ensure only publicly queryable post types are shown -- ie, post types that can be seen by visitors on the front end of the site. 

See comment on filter for why I've removed the output type.

includes/utils.php
```diff
@@ -285,14 +285,28 @@ function available_pull_post_types( $connection, $type ) {
...
+	$exclude_post_types = [
+		'attachment',
+		'nav_menu_item',
+		'wp_block',
+		'wp_navigation',
+		'wp_template',
+		'wp_template_part',
+		'dt_ext_connection',
+		'dt_subscription',
+	];
```
peterwilsoncc:
```suggestion
	$exclude_post_types = [
		'attachment',
		'dt_subscription',
	];
```

This removes unneeded items inf you check for public post types per the comment above.

includes/external-connection-cpt.php
```diff
@@ -454,12 +454,7 @@ function meta_box_external_connection_details( $post ) {
 
 	$external_connection_status = get_post_meta( $post->ID, 'dt_external_connections', true );
 
-	$post_types = get_post_types(
-		array(
-			'show_in_rest' => true,
-		),
-		'objects'
-	);
+	$post_types = \Distributor\Utils\distributable_post_types( 'is_post_type_viewable' );
```
peterwilsoncc:
```suggestion
	$post_types = \Distributor\Utils\distributable_post_types( 'objects' );
```

includes/utils.php
```diff
@@ -302,47 +316,11 @@ function distributable_post_types() {
 	 * @hook distributable_post_types
 	 *
 	 * @param {array} Post types that are distributable. Default all post types except `dt_ext_connection` and `dt_subscription`.
+	 * @param {string} The type of output to return. Accepts post type 'names' or 'objects'. Default 'names'.
```
peterwilsoncc: This is a back-compat break for developers using this filter: they may get objects instead of names without any warning.

Instead, I would leave the filter unchanged (ie, always passing names) and if the function is called with `$output !== 'names'` then convert the post types to objects once the filter is run.

This very thing hit me while branch testing as I had a filter set up to `return [ 'post' => 'post' ];` so the list was throwing PHP errors.

```suggestion
```

includes/utils.php
```diff
@@ -302,47 +316,11 @@ function distributable_post_types() {
...
-	 *
-	 * @return {bool} The updated array with the list of post types that should be hidden.
-	 */
-	$hide_from_list = apply_filters( 'dt_excluded_post_types_from_permission_list', $hide_from_list );
-
-	// Strict Hide 'dt_subscription' post type.
-	$hide_from_list[] = 'dt_subscription';
-
-	return $hide_from_list;
+	return apply_filters( 'distributable_post_types', $post_types, $output );
```
peterwilsoncc:
```suggestion
	$post_types = apply_filters( 'distributable_post_types', $post_types );

	if ( 'names' === $output ) {
		return $post_types;
	}

	return array_map( 'get_post_type_object', $post_types );
```

Per note on docblock

peterwilsoncc (commented):
I've added a few notes.

I've changed the search slightly in `distributable_post_types()` but need to figure out whether `show_in_rest` should be a requirement for internal connections. 

I'll dig in to the code.

peterwilsoncc:
Sorry for the double post but I noticed a number of places in which `get_post_types()` is been called rather than using the utility function. 

The main one I suspect should use the utility function is https://github.com/10up/distributor/blob/0778f7c78984dfe79c93c00d30530294a84bec19/includes/classes/InternalConnections/NetworkSiteConnection.php#L467

https://github.com/10up/distributor/blob/0778f7c78984dfe79c93c00d30530294a84bec19/includes/classes/InternalConnections/NetworkSiteConnection.php#L862

The appearances in `includes/rest-api.php` appear to be OK as it's altering all the post types that can be shown in the rest api.

peterwilsoncc:
I think this is good to go for the changes that can go in to 1.7.0

@faisal-alvi I've made the following changes since taking the PR https://github.com/10up/distributor/pull/907/files/28878153e576b45b6d7ee99d5c362676d4e36a3d..6c3e2557b6d34ecbeedc732032ae81a59875a0ab 

I've created #912 for some further follow up in 2.0.0 -- the changes I think are still needed don't affect any of the UI. 

I've asked Darin to review as this now includes my code.

dkotter (approved):

jmslbam:
Dropping in here :) 
I checked this also and this fixes the issue that was introduce in 1.7.0. I merged this fix and everything worked again 👍  thank you for the fix!

PR #916, opened 2022-08-01
Replace gulp build with wp-scripts.

### LINKED ISSUE #892: Tooling: Replace custom gulp build with WP Scripts

To allow ease of contribution within the WordPress developer system, it would be good to switch from a custom gulp build to `@wordpress/scripts` for asset generation.

A side benefit is that the assert versions will be auto generated based on a hash of the built file/s so the chance of forgetting to bump a constant is reduced.

### Designs

_No response_

**10up Toolkit**

This was considered but as an open-source project I figured WP Scripts was better to make contributing easier.

### Code of Conduct

peterwilsoncc:
### Description of the Change

Replace gulp and various config files with `@wordpress/scripts`.

I think this is ready for review, it should probably get double sign-off as it is quite large. I recommend reviewing this with [whitespace suppressed](https://github.com/10up/distributor/pull/916/files?w=1).

I have put together a second webpack config for the `npm run release` command as I think it is needed to copy files in to the `release` subdirectory for committing to the `stable` branch. As noted inline, I would love an alternative suggestion as it doesn't seem ideal.

Closes #892.

### How to test the Change

* Ensure production file names remain the same
* Test the various builds/npm scripts have effectively the same result (ignoring whitespace, variable munging differences)

### Changelog Entry

> Changed - Build process now uses `@wordpress/scripts` in place of Gulp.

### Credits

Props @peterwilsoncc, @dinhtungdu 

### Checklist:


webpack.config.release.js
```diff
@@ -0,0 +1,35 @@
+const CopyPlugin = require( 'copy-webpack-plugin' );
```
peterwilsoncc, line 1: I'm using this in place of the gulp copy script to copy all the files over during the `release` step.

I don't really like it or think it's a proper use of webpack so suggestions on how to do this without this would be much appreciated.

package.json
```diff
@@ -1,68 +1,68 @@
...
+		"lint": "npm run lint:css; npm run lint:js; npm run lint:pkg-json",
+		"lint:css": "wp-scripts lint-style",
+		"lint:js": "wp-scripts lint-js",
+		"lint:pkg-json": "wp-scripts lint-pkg-json",
+		"makepot": "wpi18n makepot && echo '.pot file updated'",
+		"packages-update": "wp-scripts packages-update",
+		"plugin-zip": "wp-scripts plugin-zip",
+		"start": "wp-scripts start",
+		"test:e2e": "wp-scripts test-e2e",
+		"test:unit": "wp-scripts test-unit-js",
```
dinhtungdu: I think we should remove this because we don't have any e2e or unit tests.

dinhtungdu (changes requested):
I notice some issues testing this PR:
- The build misses some files, for example, `dist/css/admin-pull-table.min.css`.
- Some JS functions are broken, for example, the Pull as draft checkbox doesn't work, see [screenshot](https://user-images.githubusercontent.com/5423135/182517096-3943ecc8-6eb4-4aff-87a4-1b278d53081a.png).
- The post editor is broken, see [screenshot](https://user-images.githubusercontent.com/5423135/182517283-c7b4439c-4a60-4c2a-90c1-7a96004898f9.png). It works fine on `develop`.
- External connection JS doesn't work too, see [screenshot](https://user-images.githubusercontent.com/5423135/182517582-b98d0468-9edd-4c4a-8e98-126d873cc3ac.png).

I will continue testing and report in follow-up comments.

package.json
```diff
@@ -1,68 +1,68 @@
...
+		"check-engines": "wp-scripts check-engines",
+		"check-licenses": "wp-scripts check-licenses",
+		"format": "wp-scripts format",
+		"lint": "npm run lint:css; npm run lint:js; npm run lint:pkg-json",
+		"lint:css": "wp-scripts lint-style",
+		"lint:js": "wp-scripts lint-js",
+		"lint:pkg-json": "wp-scripts lint-pkg-json",
+		"makepot": "wpi18n makepot && echo '.pot file updated'",
+		"packages-update": "wp-scripts packages-update",
+		"plugin-zip": "wp-scripts plugin-zip",
```
dinhtungdu, line 52: Do you think we can leverage this to create the release file?

package.json
peterwilsoncc: Now the Cypress tests have been merged, I'll change the command to run those once I update against `develop`.

package.json
peterwilsoncc: We may be able to do so on `stable` but I will need advice on how to test it.

peterwilsoncc:
I think I've addressed the various issues @dinhtungdu picked up:

* fc21fd1b799e5f353f7cdf96409873e64b6ff656 fixes the block editor -- the svg needs to be a react component now 
* 104ba2322578d4ddb686cefd411db667af3f3437 fixes the name of the CSS file restoring `admin-pull-table.min.css` 
* 4cd3fc8bd40b96ad68771bba58fc7d2dbb933fa8 fixes the external connections screen

WP Scripts doesn't seem to produce code compatible with older versions of WP so I've pulled in 04a15803c6bc6e775025a39122579dbcf441321e from the requirements bump PR as both the version bump and this issue are on the same milestone.

dinhtungdu:
@peterwilsoncc Thanks for the updates! I can confirm all issues I noted are addressed.

I did another round of testing and found some additional issues, please check the screenshots below for more detail.

<details>
<summary>Screenshots</summary>


Screen Shot 2022-08-06 at 00 04 43
Screen Shot 2022-08-06 at 00 03 20

</details>

peterwilsoncc:
Distributor panel restored in 0b53814306ae663dca9d50e6b30ad5484d6eb28c -- I'd reversed a media query.

For the `checked` attribute showing, I think that's expected due to the differences between properties and attributes.

The property shows the current state, the attribute the state at page load. If you run `jQuery('#dt-as-draft').prop( 'checked' )` in the browser console, you should get an accurate result when unchecked.

peterwilsoncc:
@dinhtungdu I think this is ready for another review. I recommend [suppressing white space](https://github.com/10up/distributor/pull/916/files?w=1).

For now I have kept all the built asset file names the same which means the webpack config is a little odd. I've done this for back-compat but if it's decided it isn't needed then that can be done in a smaller follow up commit.

dinhtungdu (approved):
Awesome! 🎉
Thank you so much for this hard work! This PR tested well on my end.

PR #922, opened 2022-08-03
Don't discard the `post_type` argument, as this will force all items to be the type `post`

### LINKED ISSUE #919: Custom post type gets copied as a regular post
### Describe your question

I have 2 sites on a multisite network, with a CPT `nt_course`, but when I "distribute" it, it gets cloned as a `post` , not as a CPT. Any suggestions?

### Code of Conduct

### LINKED ISSUE #921: Pages are getting pulled in as posts
Huge issue. Pages are showing up under Distributor > Pull Content > Pages as "Post Type" = Page...

But when I pull them in as draft, they pull in as Posts.

I have replicated this with two different websites/two different External Connections, and have tried to Pull in the reverse (using the Distributor Content website as the receiver...)

This just happened recently. I can't get any of my Pages to pull in as Pages anymore.

dkotter:
### Description of the Change

In #895 we added some code to ensure only the arguments we want are allowed to be passed in to `wp_insert_post`. The `post_type` argument was left off that list, which now means that argument isn't set and `wp_insert_post` will default to using `post` as the argument there. This means any non-post `post_type` will be imported as a post.

This is a simple fix of just adding `post_type` to the list of approved arguments, so that value won't be lost.

Closes #919, #921

### How to test the Change

Test pulling in a non-post `post_type` item (Page or CPT) from both an internal and external connection. Ensure in both instances the proper `post_type` is retained.

Test pushing a non-post `post_type` item (Page or CPT) from both an internal and external connection. Ensure in both instances the proper `post_type` is retained.

### Credits
Props @dkotter, @pdewouters, @andygagnon, @jmstew3

### Checklist:

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -1140,6 +1140,8 @@ function post_args_allow_list( $post_args ) {
 		'post_excerpt',
 		'post_status',
 		'post_type',
+		'comment_status',
+		'ping_status',
 		'post_password',
 		'post_name',
 		'to_ping',
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -1139,6 +1139,7 @@ function post_args_allow_list( $post_args ) {
 		'post_title',
 		'post_excerpt',
 		'post_status',
+		'post_type',
 		'post_password',
 		'post_name',
 		'to_ping',
--- a/tests/php/UtilsTest.php
+++ b/tests/php/UtilsTest.php
@@ -874,6 +874,7 @@ class UtilsTest extends TestCase {
 	function test_post_args_allow_list() {
 		$post_args = [
 			'post_title'   => 'Test Title',
+			'post_type'    => 'post',
 			'post_content' => 'Test Content',
 			'post_excerpt' => 'Test Excerpt',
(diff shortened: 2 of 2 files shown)


includes/utils.php
```diff
@@ -1139,6 +1139,7 @@ function post_args_allow_list( $post_args ) {
 		'post_title',
 		'post_excerpt',
 		'post_status',
+		'post_type',
```
dkotter, line 1142: Also noticed we don't include the `comment_status` or `ping_status` arguments here, though those are default ones `wp_insert_post` accepts. Were those left out intentionally?

peterwilsoncc: I've added them in 835b1ca38723f1ce4d323b2529c0743857091a4d.

I suspect I lost my spot going through the arguments and they were skipped due to an oversight, along with the post type.

peterwilsoncc (approved):

pdewouters:
thanks @peterwilsoncc !

jmslbam:
Dropping in here unannounced :)
I checked this also and this fixes the issue that was introduce in 1.7.0. I merged this fix and everything worked again 👍 thank you for the fix!

PR #937, opened 2022-09-08
Update `yahnis-elsts/plugin-update-checker` from 4.9 to 4.13

dkotter:
### Description of the Change

The Plugin Update Checker library we use is also used on a few other 10up projects. Some of those are running different versions though, leading to conflicts (especially when using composer to manage all dependencies of a project). This PR updates this library to the latest version, which will also be done in other repos that are using this library.

- [Classifai](https://github.com/10up/classifai/pull/365)
- [10up Experience](https://github.com/10up/10up-experience/pull/116)

Alternatively we could use something like [mozart](https://github.com/coenjacobs/mozart) or [PHP-Scoper](https://github.com/humbug/php-scoper) to add a distinct namespace to this dependency. I'm not opposed to this necessarily but because we're seeing conflicts with other plugins that we fully control, I like the idea of just ensuring all of those projects are using the same version.

### How to test the Change

1. Pull these changes down
2. Within the `distributor.php` file, set the version number to something lower, like 1.6.0
3. Ensure you have a proper license in place
4. Go to the plugins section and ensure you see an update notice for Distributor (may need to click the `Check for updates` link; also ensure the plugin is active)
5. Click on the `View version X.X.X details` link and ensure you see proper details
6. Go to the WordPress Updates page and ensure you see Distributor there as well, with the icon
7. Click on the `View version X.X.X details` link and ensure you see proper details
8. If desired, click the `update now` link and ensure an update works (either from the plugins page or the main updates page). Note though that this will wipe out your version controlled `distributor` directory, so proceed with caution

### Changelog Entry

> Changed - Upgrade the Plugin Update Checker library to 4.13

### Credits

Props @dkotter

### Checklist:


peterwilsoncc (approved):
LGTM 

Git remains clean after running `rm -rf vendor; composer install` so all changes appear to have been committed.

Otherwise I manually tested per the notes in the PR description.

PR #938, opened 2022-09-12
Update distributions after post terms and meta is saved.

### LINKED ISSUE #399: Delay `send_notifications` until after all meta has saved (Yoast SEO problem).

peterwilsoncc:
### Description of the Change

This moves the updating of distributions during edits from the `save_post` hook to the `wp_after_insert_post` hook.

In updates via both the classic and REST API, WordPress fires this hook after all terms and post meta is updated. This prevents the distribution of legacy data described in the issue.

Closes #399 

### How to test the Change

1. Install this mini-plugin that [updates post meta with each save](https://gist.github.com/peterwilsoncc/61eeb5f2337cdf4976608a5123acbb1d) 
2. On the main `develop` branch
    a) Create and distribute a post via the block editor
    b) Update the post
    c) Check the database -- observe the value of the post meta differs between source and distribution (one less)
4. Switch to this branch
5. Resave the post in the block editor
6. observe the post meta matches in both the source and distributed material.

### Changelog Entry

> Fixed - Ensure post meta and terms have saved prior to distribution.

### Credits

Props @peterwilsoncc, @dinhtungdu, @dkotter, @timstl, @faisal-alvi.

### Checklist:

### DIFF
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -579,7 +579,7 @@ class NetworkSiteConnection extends Connection {
 	 */
 	public static function bootstrap() {
 		add_action( 'template_redirect', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'canonicalize_front_end' ) );
-		add_action( 'save_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'update_syndicated' ), 99 );
+		add_action( 'wp_after_insert_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'update_syndicated' ), 99 );
 		add_action( 'before_delete_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'separate_syndicated_on_delete' ) );
 		add_action( 'before_delete_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'remove_distributor_post_from_original' ) );
 		add_action( 'wp_trash_post', array( '\Distributor\InternalConnections\NetworkSiteConnection', 'separate_syndicated_on_delete' ) );
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -704,12 +704,6 @@ class NetworkSiteConnection extends Connection {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


peterwilsoncc:
I tried writing an E2E test for this (using tags rather than meta as the UI is already exposed) but I will need some help from some one more familiar with cyrpress.

faisal-alvi (commented):
@peterwilsoncc Code LGTM but the issue is not regenerating for me in the `develop` branch.

Please check the video: 

https://user-images.githubusercontent.com/25176325/190115685-e4e080ac-e83a-453a-8c9b-e93dd46c94ff.mp4

Let me know if I'm not testing it correctly.

peterwilsoncc:
@faisal-alvi Your tests process is correct, so I am unclear why we are seeing different results. I do have the classic editor installed so I wonder if it's an interaction with that. You'll see in the video that the distributed block on xu-distributor.local shows as a classic block.

Not in the video, I start on this branch so the values are in sync. The values go out of sync once I switch to develop. https://youtu.be/8grzn3YdWXk

faisal-alvi (approved):
@peterwilsoncc thanks for the video. I have retried and this time it regenerated for me in the `develop` and worked fine in the PR branch.

image

PR #951, opened 2022-09-30
Sync empty taxonomies to distributed posts.

### LINKED ISSUE #625: Taxonomy terms don’t sync among sites.
Let’s say I have 2 sites (Site A and B). I create several posts with taxonomy terms on B, then in site A I pulled from B, and I am able to see the posts pulled form B with their taxonomy terms.
Then I update a post on site B with it has 3 taxonomy terms, but I decide to delete one.
Site A is synced with B, and the post is updated showing two taxonomy terms. If I add more terms on B, A is synced too. If I delete others on B, but at least I leave one term, site A is synced too. The problem is if I delete all the terms on B, site A doesn’t sync.

Is there a way to solve this situation with taxonomy terms?

peterwilsoncc:
### Description of the Change

This modifies `prepare_taxonomy_terms()` to include both populated and unpopulated taxonomies when distributing sites.

This allows for the situation in which an original post is updated to remove all terms within a shared taxonomy.

To account for empty arrays being stripped from `$_POST` data (I've not been able to determine where this happens), update notifications are now sent as JSON encoded data to the remote sites REST API.  _This may be a backward compatibility issue for sites using hooks and getting POSTed data rather than using the `$request` object._

Taxonomy updates are limited to those shown in the REST API. 

Closes #625

### How to test the Change

1. Check out the `develop` branch.
2. Create a post and assign it two or more tags.
3. Distribute the post
4. Observe distributed version of post has two tags
5. Delete one tag on original post, update it.
6. Observe distributed version of post has one tag.
7. Delete all tags on original post, update it.
8. Observe distributed version of post has one tag (error).
9. Check out this branch and repeat
10. After deleting tags on the original post, the distributed version should have no tags.

#### Notes

1. E2E tests are failing for trunk due to a bug in `@wordpress/env` -- it was fixed five days ago but is still to be released https://github.com/WordPress/gutenberg/commit/3983fee52b8fa0b9336d6202e8b92d407b99187c 
2. I've had to change some of the tests to work around an issue in wp_mocks. I've created an issue and linked to that within the comments.

### Changelog Entry

> Fixed - Distribute empty taxonomies to external sites.

### Credits

Props @peterwilsoncc, @manolobevia, @jeffpaul, @cadic

### Checklist:

### DIFF
--- a/includes/classes/API/SubscriptionsController.php
+++ b/includes/classes/API/SubscriptionsController.php
@@ -248,6 +248,11 @@ class SubscriptionsController extends \WP_REST_Controller {
 				'media'        => ( isset( $request['post_data']['distributor_media'] ) ) ? $request['post_data']['distributor_media'] : [],
 			];
 
+			// Limit taxonomy updates to those shown in the REST API.
+			$rest_taxonomies = get_taxonomies( [ 'show_in_rest' => true ] );
+			$rest_taxonomies = array_fill_keys( $rest_taxonomies, true );
+			$update['terms'] = array_intersect_key( $update['terms'], $rest_taxonomies );
+
 			update_post_meta( (int) $request['post_id'], 'dt_subscription_update', $update );
 
 			$unlinked = (bool) get_post_meta( $request['post_id'], 'dt_unlinked', true );
--- a/includes/subscriptions.php
+++ b/includes/subscriptions.php
@@ -284,33 +284,42 @@ function send_notifications( $post ) {
 			}
 		}
 
+		/**
+		 * Filter the timeout used when calling `\Distributor\Subscriptions\send_notifications`
+		 *
+		 * @hook dt_subscription_post_timeout
+		 *
+		 * @param {int}     $timeout The timeout to use for the remote post. Default `5`.
+		 * @param {WP_Post} $post    The post object
+		 *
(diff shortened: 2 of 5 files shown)


peterwilsoncc:
The tests here are failing because the following returns `null` rather than an empty array. I've tried figuring out why but need some help from someone else.  

```php
\WP_Mock::userFunction(
	'\Distributor\Utils\prepare_taxonomy_terms', [
		'return' => [],
	]
);
```

I did a var_dump on the following line within `test_send_notifications_remote_post_exists()`

```
/vagrant/content/plugins/distributor/tests/php/SubscriptionsTest.php:514:NULL
```

cadic (approved):
The code looks good.

BTW, I was not able to replicate the original issue (at `develop` branch: after second-the-last tag was deleted in the original post, the distributed post has no tags as well on the step 8).

peterwilsoncc:
@cadic Do you recall if you were testing on an internal (multisite) connection or an external connection?

cadic:
@peterwilsoncc oh, thanks for clarifying that, it does reproduce with the external connection.

peterwilsoncc:
Thanks Max, sorry I missed that step in the reproduction notes.

PR #959, opened 2022-10-19
Allow WP Trunk E2E tests to fail.

### LINKED ISSUE #958: Temporarily allow E2E `trunk` tests to fail.

Due to an upstream bug in `wordpress/env` the E2E tests for WordPress trunk are failing. See https://github.com/WordPress/gutenberg/pull/44887

As a temporary measure, these tests should be allowed to fail without blocking merge/ending the other E2E test runs early.

### Designs

_No response_

_No response_

### Code of Conduct

peterwilsoncc:
### Description of the Change

Due to an upstream bug in `wordpress/env` the E2E tests for WordPress trunk are failing. See https://github.com/WordPress/gutenberg/pull/44887

As a temporary measure, these tests should be allowed to fail without blocking merge/ending the other E2E test runs early.

Closes #958

### Credits

Props @peterwilsoncc 

### Checklist:

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -13,6 +13,8 @@ on:
 jobs:
   cypress:
     name: ${{ matrix.core.name }} / ${{ matrix.php.name }}
+    ## Pending upgrade of `@wordpress/env` package.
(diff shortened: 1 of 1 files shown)


jeffpaul:
This approach works for me, we could also pause the running of the action entirely and avoid a code change in the future.

PR #961, opened 2022-10-19
Change from using the term `blacklist` to `excluded`

### LINKED ISSUE #960: Update usages of `blacklist` to be more inclusive

We have a handful of places that still use the term `blacklist` (mostly in our utilities). We should look to update these to use something like `excluded` instead.

### Designs

_No response_

_No response_

### Code of Conduct

dkotter:
### Description of the Change

There's a handful of places we were using the term `blacklist`. To be more inclusive (and more descriptive), this PR changes those to use the term `excluded`.

There was one utility function and one filter that used the term `blacklist`. I've added deprecations for both of those, though we may be safe to just completely remove/rename the function (more likely though that the filter is being used).

Note also there's a handful of places that will need version numbers updated before we release this. I've left those as `X.X.X` for now.

Closes #960

### How to test the Change

Nothing that really needs tested here, though could test the meta exclusion piece to ensure nothing broke there (though this should just be name changes, no functionality changes).

### Changelog Entry

> Changed - Change uses of `blacklist` and use `exclude` instead
> Deprecated - Deprecate the `Distributor\Utils\blacklisted_meta` function and the `dt_blacklisted_meta` filter

### Credits
Props @dkotter, @jeffpaul 

### Checklist:

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -330,7 +330,7 @@ function distributable_post_statuses() {
  */
 function blacklisted_meta() {
 	_deprecated_function( __FUNCTION__, 'X.X.X', '\Distributor\Utils\excluded_meta()' );
-	excluded_meta();
+	return excluded_meta();
 }
 
 /**
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -137,17 +137,17 @@ function is_dt_debug() {
 /**
  * Given an array of meta, set meta to another post.
  *
- * Don't copy in blacklisted (Distributor) meta.
+ * Don't copy in excluded (Distributor) meta.
  *
  * @param int   $post_id Post ID.
  * @param array $meta Array of meta as key => value
  */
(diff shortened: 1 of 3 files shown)


includes/utils.php
```diff
@@ -322,23 +322,35 @@ function distributable_post_statuses() {
...
+ * @since  X.X.X
+ * @return array
+ */
+function excluded_meta() {
+
 	/**
-	 * Filter meta keys that are blacklisted from distribution.
+	 * Filter meta keys that are excluded from distribution.
 	 *
 	 * @since 1.0.0
```
peterwilsoncc, line 347:
```suggestion
	 * @since 1.0.0
	 * @deprecated x.x.x
```

peterwilsoncc (commented):
LGTM, one note inline which I'll self merge.

peterwilsoncc (approved):

PR #964, opened 2022-11-14
Post abstraction

### LINKED ISSUE #207: Consider a post abstraction
This feedback is from a VIP code review by @tomjn - opening this issue to keep track of it.

> As I mentioned on the previous distributor reviews, I feel that there is still an abstraction that can be made to move post wrangling into its own object, that would make the codebase more manageable, testable, and more reviewable, if not smaller.

peterwilsoncc:
### Description of the Change

Creates a post abstraction to aid with the distribution of posts without the need to duplicate code across each connection type.

Closes #207

### How to test the Change

WIP

### Changelog Entry

> Added - Post abstraction class.

### Credits

Props @peterwilsoncc, @dkotter 

### Checklist:


includes/classes/DistributorPost.php
```diff
@@ -0,0 +1,154 @@
...
+			$this->connection_id   = get_post_meta( $post->ID, 'dt_original_blog_id', true );
+		} elseif ( get_post_meta( $post->ID, 'dt_original_source_id', true ) ) {
+			// Post was pulled from an external connection.
+			$this->connection_type = 'external';
+			$this->connection_id   = get_post_meta( $post->ID, 'dt_original_source_id', true );
+		} elseif ( get_post_meta( $post->ID, 'dt_subscription_signature', true ) ) {
+			// Post was pushed from an external connection.
+			$this->connection_type = 'pushed';
+			$this->connection_id   = get_post_meta( $post->ID, 'dt_subscription_signature', true );
+		}
```
peterwilsoncc: This is broken...

`dt_original_source_id` is included in both pushed and pulled posts.

In pushed posts, `dt_original_source_id` is stored in the distributed copy as the connection ID on the source server. IE, the distributed copy does not have access to the post data.

In pulled posts, `dt_original_source_id` is stored in the distrubuted copy as the connection on on the source server, IE the distributed copy does have access to the post data.

peterwilsoncc:
This is how I've been testing, it var_dumps the object in place of the posts content on the front end of the site:

```php
function xu_dist_nope_the_content( $content ) {
	if (
		is_admin()
		|| ( defined( 'REST_REQUEST' ) && REST_REQUEST )
	) {
		return $content;
	}




	if ( is_singular() && in_the_loop() && is_main_query() ) {
		remove_filter( 'the_content', 'xu_dist_nope_the_content' );

		$post = get_post();
		$dt = new Distributor\DistributorPost( $post );
		$t = $dt->original_post_url;
		// $t = $dt_clone->source_site;

		// This is just to get a smaller dump.
		$dt_clone = clone $dt;
		unset( $dt_clone->post->post_content );

		ob_start();
		var_dump( get_post()->ID, $dt_clone );
		var_dump( isset( $dt_clone->original_post_url ) );


		return ob_get_clean();
	}

	return $content;
}

add_filter( 'the_content', 'xu_dist_nope_the_content' );
```

peterwilsoncc:
I think this is in a good spot for an initial review and (hopefully) merge so I can proceed to use it in the connectors in follow up PRs. 

I'm sure it will change/expand slightly during implimentation but I think getting it in earlier is probably better than a mega PR.

Prepare media is pretty substantial in that it now parses all media rather than just those attached to the post directly.

includes/classes/DistributorPost.php
```diff
@@ -0,0 +1,563 @@
...
+	 * Get the post ID.
+	 *
+	 * @return int Post ID.
+	 */
+	public function get_the_id() {
+		return $this->post->ID;
+	}
+
+	/**
+	 * get the post permalink.
```
dkotter:
```suggestion
	 * Get the post permalink.
```

includes/classes/DistributorPost.php
```diff
@@ -0,0 +1,563 @@
...
+		 *
+		 * The Distributor defaults are {
+		 *    'core/image' => 'id',
+		 *    'core/audio' => 'id',
+		 *    'core/video' => 'id',
+		 * }
+		 *
+		 * @param array $media_blocks Array of media blocks.
+		 */
+		$media_blocks = apply_filters( 'dt_parse_media_blocks', $media_blocks );
```
dkotter, line 485: We'll want to ensure we follow the [JSDoc standards](https://github.com/10up/actions-wordpress/blob/stable/hookdocs-workflow.md#docblock-example) here so this filter ends up in our hook docs.

includes/utils.php
```diff
@@ -438,30 +440,12 @@ function prepare_meta( $post_id ) {
...
-		$featured_image['featured'] = true;
-
-		$media_array[] = $featured_image;
+	$dt_post = new DistributorPost( $post_id );
+	if ( ! $dt_post ) {
+		return array();
 	}
 
-	return $media_array;
+	return $dt_post->prepare_media();
```
dkotter: Looks like the `prepare_media` method is called as part of the `constructor` for `DistributorPost`. Can we just do `$dt_post->media` at this point?

includes/classes/DistributorPost.php
```diff
@@ -0,0 +1,563 @@
...
+		if ( ! $post ) {
+			return;
+		}
+
+		$this->post = $post;
+
+		// Set up the distributable data.
+		$this->meta  = Utils\prepare_meta( $post->ID );
+		$this->terms = Utils\prepare_taxonomy_terms( $post->ID );
+		$this->media = $this->prepare_media();
```
dkotter: Wondering if we should set this data in the `constructor` or if it should be set when requested? I guess I'm not sure if this data is needed every time we create an instance of `DistributorPost` or if sometimes this isn't necessary, resulting in potential impacts to performance.

I know this is still a work in progress so maybe this data is always needed but figured I'd call it out

includes/classes/DistributorPost.php
```diff
@@ -0,0 +1,563 @@
...
+			/*
+			 * Internal connections store the original blog's ID.
+			 *
+			 * Pushed and pulled posts are indistinguishable from each other.
+			 */
+			$this->connection_type = 'internal';
+			$this->connection_id   = get_post_meta( $post->ID, 'dt_original_blog_id', true );
+		} elseif ( get_post_meta( $post->ID, 'dt_full_connection', true ) ) {
+			// This connection was pushed from an external connection.
+			$this->connection_type = 'pushed';
```
dkotter: Wondering if we need another property to store if something is pushed or pulled? I guess just seems a little weird to me to have a `$connection_type` property but instead of storing `external` there, we store `pushed`.

I understand there's not a way to distinguish pushed or pulled on internal content, so in essence this is `external - pushed` and the other is `external - pulled` but might be more obvious to have a different property to store if it's a push or pull.

dkotter:
Left a few comments but overall this is looking good to me. Will be interesting to see if this helps simplify things / reduce code duplication once we start implementing this in the various connection classes.

dkotter:
I guess one other thing, not sure if we want to merge this into `develop` or if this is another of those that we should create a temporary `develop` to build off of (`develop-v2`?). Just so all of this work can be merged in as completed but it keeps our queue clean in case other things need to be released prior to work being done?

includes/classes/DistributorPost.php
peterwilsoncc: > Wondering if we should set this data in the `constructor` or if it should be set when requested?

Good point.

There's an option for limiting which media is distributed so I'll start with that and see if I can either lazy load (similar to source site items) or more simply switch to methods to get all the things.

peterwilsoncc: > Wondering if we should set this data in the `constructor` or if it should be set when requested? I guess I'm not sure if this data is needed every time we create an instance of `DistributorPost` or if sometimes this isn't necessary, resulting in potential impacts to performance.

I ended up doing a couple of things:

* storing the media data in a persistent cache for block posts f5d6ddb76c0c94463f75926dfde15c0299a9fe51
* removed the meta, terms and media population from the constractor
* cache them within the object but I may end up removing that to account for post updates.

includes/classes/DistributorPost.php
peterwilsoncc: See 768515f101c01b6c0e32534b4af2528a38548f40

peterwilsoncc:
> I guess one other thing, not sure if we want to merge this into `develop` or if this is another of those that we should create a temporary `develop` to build off of (`develop-v2`?). Just so all of this work can be merged in as completed but it keeps our queue clean in case other things need to be released prior to work being done?

I've got `develop-v1` for items that need to be cherry picked so I think that will be fine. If we end up with massive conflicts, patch releases can come from the v1 branch.

includes/utils.php
peterwilsoncc: Moved filling this out of the constructor.

peterwilsoncc: I've removed populating media, meta and terms in the constructor. They now use methods and rely on the object cache.

includes/settings.php
```diff
@@ -436,6 +438,32 @@ function handle_network_settings() {
...
+
+	if ( function_exists( 'wp_cache_delete_multiple' ) ) {
+		wp_cache_delete_multiple( $cache_keys, 'dt::post' );
+	} else {
+		foreach ( $cache_keys as $cache_key ) {
+			wp_cache_delete( $cache_key, 'dt::post' );
+		}
+	}
+}
+
```
peterwilsoncc, line 466: There's possibly a better spot to put this, maybe even in bootstrap.php.

peterwilsoncc:
@dkotter I think this is ready for another look now I've address the perf issues and other notes.

tests/php/UtilsTest.php
```diff
@@ -6,6 +6,23 @@
...
+	public function setUp(): void {
+		parent::setUp();
+
+
+		// Return voids.
+		\WP_Mock::userFunction( '_prime_post_caches' );
+		\WP_Mock::userFunction( 'update_object_term_cache' );
+		\WP_Mock::userFunction( 'update_postmeta_cache' );
+	}
+
```
peterwilsoncc, line 25: I tried to include these in `tests/php/includes/TestCase.php` but it wasn't been registered.  I renamed the class in that file to `IsThisEvenUsed` and phpunit didn't throw any fatal errors, so I suspect it isn't used.

dkotter (dismissed):

peterwilsoncc:
@dkotter Sorry, I managed to dismiss your review wrapping up the tests.

> I've got `develop-v1` for items that need to be cherry picked so I think that will be fine. If we end up with massive conflicts, patch releases can come from the v1 branch.

Are you happy for this to go in `develop` given the above or would you still prefer a targeted branch to avoid the potential for a mess.

dkotter:
> > I've got `develop-v1` for items that need to be cherry picked so I think that will be fine. If we end up with massive conflicts, patch releases can come from the v1 branch.
> 
> Are you happy for this to go in `develop` given the above or would you still prefer a targeted branch to avoid the potential for a mess.

Yeah, that sounds good to me

dkotter (approved):

PR #1003, opened 2023-01-12
Add view all post types filter to External Connections.

### LINKED ISSUE #861: Add ability to view all post types when Pulling from an External Connection

ankitrox:
### Description of the Change

Introduce a "View All" post type view to the pull screen for external connections for parity with the internal connections. The new view is the default when visiting Distributor > Pull Content > [External connection].

Closes #861 

### How to test the Change

1. Create two separate sites using insta wp.
2. Install distributor plugin on both the sites. Connect the sites.
3. Create a custom post type "Book" on both the sites.
4. On pull screen ensure that "View All" option is available.
5. Pull content from CPT and it should pull the posts from external sites.

### Changelog Entry

> Added - Introduce "View all" post type view when pulling content from external connections.

### Credits

Props @ankitrox, @Sidsector9, @ravinderk, @jeffpaul, @peterwilsoncc 

### Checklist:


jeffpaul:
Tagging Max on this one as he's already reviewed a separate PR #1002 for the same issue, so worth checking here to see which approach is most optimal (and can close the lessor of the two)

Sidsector9:
@ankitrox can you please fill up the description and testing instructions for clarity?

ankitrox:
> @ankitrox can you please fill up the description and testing instructions for clarity?

Added the steps to test

Sidsector9 (changes requested):
Hi @ankitrox 

I followed your instructions and tested this branch, I faced some issues. The Posts filter shows Pages as well.

I tested by resolving the conflicts locally as well and still faced the same issue.

Please watch the video below:

https://github.com/10up/distributor/assets/17757960/7bfb0b6a-6caf-43ed-9277-54f3fdb70d09

Let me know if I'm missing any steps.

peterwilsoncc:
The changes in `develop` to the `includes/rest-api.php` file are causing a few conflicts for this. 

However, the good news is that the changes allow for requesting `any` as the post type to get all the post types so you won't need to make any changes to the file -- the other files will need to specify the post type as `any` rather than `all`.

peterwilsoncc (commented):
I've pushed a few changes to account for updates to the REST API endpoint in Distributor 2.0.0

For external connections, I am wondering if it would be best to send the local post types individually when requesting `all` post types: this will allow for situations in which the external connection has post types that are not supported locally.

However, I think that might be best left fro another ticket.

ravinderk (changes requested):
@peterwilsoncc I am not getting a correct result for the `View all` filter.

I noticed that in the final SQL query, WordPress adds `post_author` to the current user id (rest API) when doing a query for multiple posts (i.e. `all`), which means I am not getting the list of all editable posts from all post types.

Can you check?
 
image
image

peterwilsoncc (dismissed):
> I noticed that in the final SQL query, WordPress adds `post_author` to the current user id (rest API) when doing a query for multiple posts (i.e. `all`), which means I am not getting the list of all editable posts from all post types.

I'm not seeing this issue when connected as an administrator. If Ravinder had connected the sites as a non-admin it may be an effect of the security fixes in https://github.com/10up/distributor/security/advisories/GHSA-q43c-v867-4cfp 

I think this code is good or merging.

jeffpaul:
Closing in favor of #1002.

PR #1004, opened 2023-01-13
Consider multisite for posts abstraction.

### LINKED ISSUE #1007: Consider multisite for post abstraction.
### Describe the bug

Network connections make heavy use of the `switch_to_blog()` function. With regard to the post abstraction, this can be problematic if a helper method is called after switching blogs as the returned data can be a mix of content from the original site's post and the switched site's post.

For the methods within the post abstraction, it would be good to ensure the data always comes from the site used when the post abstraction was created.

### Steps to Reproduce

1. Create a fresh multisite install of WordPress (example below uses subdirectory mode)
2. Network activate distributor using the `develop` branch.
3. On the main site, edit the hello world's post's title to `Edited sample posts title`
4. Create a subsite (it should have the site ID 2)
5. Run the following in the wp cli shell command.
   ```
   wp> $dt_post = new Distributor\DistributorPost( 1 );
   wp> $dt_post->get_permalink();
   string "http://ms-distributor.local/?p=1"
   wp> $dt_post->post->post_title
   string "Edited sample posts title" 
   
   wp> switch_to_blog( 2 );
   wp> $dt_post->get_permalink();
   http://ms-distributor.local/two/?p=1
   wp> $dt_post->post->post_title
   string "Edited sample posts title"   
   ```

Note that after switching to site two, the `$dt_post` variable is returning a mix of data from the primary and secondary site

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

peterwilsoncc:
### Description of the Change

Accounts for multisite in the post abstraction helper methods.

Ensures the site switched correctly when using accessing post data. This prevents incorrect data during post ID collisions, or missing data if the post does not exist on the switched site.

Uses a lot of statics to avoid needless switching. This might not be worth it.

Closes #1007

### Credits

Props @peterwilsoncc 

### Checklist:

### DIFF
--- a/includes/classes/DistributorPost.php
+++ b/includes/classes/DistributorPost.php
@@ -111,6 +111,13 @@ class DistributorPost {
 	private $source_site = [];
 
 	/**
+	 * The site ID for the current site.
+	 *
+	 * @var int
+	 */
+	public $site_id;
+
+	/**
 	 * Initialize the DistributorPost object.
 	 *
 	 * @param WP_Post|int $post WordPress post object or post ID.
@@ -176,6 +183,20 @@ class DistributorPost {
 			$this->connection_direction = 'pulled';
 			$this->connection_id        = (int) get_post_meta( $post->ID, 'dt_original_source_id', true );
 		}
+
+		$this->site_id = get_current_blog_id();
+
+		// Pre-populate the post data to account for switching sites.
+		if ( ! is_multisite() ) {
+			return;
+		}
+
+		$this->get_permalink();
+		$this->get_post_thumbnail_id();
+		$this->get_post_thumbnail_url();
+		$this->get_the_post_thumbnail();
+		$this->get_meta();
+		$this->get_terms();
 	}
 
 	/**
@@ -314,7 +335,12 @@ class DistributorPost {
 	 * @return string Post permalink.
 	 */
 	public function get_permalink() {
-		return get_permalink( $this->post );
+		static $permalink = null;
+		if ( null === $permalink ) {
+			$permalink = get_permalink( $this->post );
+		}
+
+		return $permalink;
 	}
 
 	/**
@@ -323,7 +349,7 @@ class DistributorPost {
 	 * @return string Post type.
 	 */
 	public function get_post_type() {
-		return get_post_type( $this->post );
+		return $this->post->post_type;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/classes/DistributorPost.php
```diff
@@ -353,6 +389,23 @@ public function get_post_thumbnail_url( $size = 'post-thumbnail' ) {
...
+			}
+			$thumbnail[ $hash ] = get_the_post_thumbnail( $this->post, $size, $attr );
+
+			if ( $switched ) {
+				restore_current_blog();
+			}
+		}
+
+		return $thumbnail[ $hash ];
 		return get_the_post_thumbnail( $this->post, $size, $attr );
```
dkotter: Need to remove this return statement

includes/classes/DistributorPost.php
```diff
@@ -176,6 +183,20 @@ public function __construct( $post ) {
 			$this->connection_direction = 'pulled';
 			$this->connection_id        = (int) get_post_meta( $post->ID, 'dt_original_source_id', true );
 		}
+
+		$this->site_id = get_current_blog_id();
+
+		// Pre-populate the post data to account for switching sites.
```
dkotter: Can you expound on what the problem is? I guess I'm not quite understanding why all the data below needs to be set when we are on a multisite. I'm assuming it has something to do with ensuring we have the right information from the right site but I guess not completely following without being able to see in what situations this comes into play

includes/classes/DistributorPost.php
```diff
@@ -314,7 +335,12 @@ public function get_the_id() {
 	 * @return string Post permalink.
 	 */
 	public function get_permalink() {
-		return get_permalink( $this->post );
+		static $permalink = null;
```
dkotter: Not opposed to using statics here but does this actually help in practice? Are we calling these methods multiple times in the same execution?

includes/classes/DistributorPost.php
peterwilsoncc: > ensuring we have the right information from the right site

This is the goal.

We we distribute network sites, we run `switch_to_blog()` fairly early on (I'll probably try to delay it) and calling the helper methods was returning either `false` or the wrong post's information if there was a post ID collision.

In the `DistributorPost::post_data()` method, we currently call `get_the_title()` and the `::get_media()`, `::get_terms()` and `::get_meta()` methods. 

I'm wanting to make sure they always return data related to the post and site we set up the object with. 

I guess I need to improve the comment.

includes/classes/DistributorPost.php
peterwilsoncc: We may... some of the canonical filters call `get_permalink`.

I'm not sure what's better or worse: making sure the site matches and switching if needs be or using statics. The former is slow but the latter may hit order of operations issues if other plugins add or remove filters.

peterwilsoncc:
Closing this off as an interesting experiment.

Replaced with #1010

PR #1006, opened 2023-01-20
Update `unlink` warning message copy

jeffpaul:
### Description of the Change
The editor notice that displays on remote, distributed posts has some odd spacing (line break where a single space should be, no space after a period) and is missing trailing periods in some instances.  This PR attempts to clean up that spacing and missing periods as well as further clarify the terms of "origin" and "remote" posts when referring to distributed content.

See the [screenshot](https://webtrainingwheels.com/wp-content/uploads/2016/06/unlink.jpg) from a [sample how-to post](https://webtrainingwheels.com/syncing-content-between-wordpress-sites-wp-site-sync/):

unlink

While I've attempted various searches through the codebase to try and catch all the instances, its possible I missed some so having someone else run a spot-check on this branch to ensure things are **all** properly updated would be wonderful.

### How to test the Change
Ask ChatGPT, I'm curious if it knows.

### Changelog Entry

> Changed - Descriptive warning message copy on remote, distributed posts.

### Credits

Props @jeffpaul, @peterwilsoncc, @cadic.

### Checklist:

### DIFF
--- a/assets/js/gutenberg-syndicated-post.js
+++ b/assets/js/gutenberg-syndicated-post.js
@@ -14,7 +14,7 @@ if (
 		message = sprintf(
 			/* translators: 1) Distributor post type singular name, 2) Source of content. */
 			__(
-				'This %1$s was distributed from %2$s. However, the origin post has been deleted.'
+				'This %1$s was distributed from %2$s. However, the origin %1$s has been deleted.'
 			),
 			dtGutenberg.postTypeSingular,
 			dtGutenberg.originalLocationName
@@ -23,7 +23,7 @@ if (
 		message = sprintf(
 			/* translators: 1) Source of content, 2) Distributor post type singular name. */
 			__(
-				'Distributed from %1$s. This %2$s is linked to the origin post. Edits to the origin post will update this remote version.',
+				'Distributed from %1$s. This %2$s is linked to the origin %2$s. Edits to the origin %2$s will update this remote version.',
 				'distributor'
 			),
 			dtGutenberg.originalLocationName,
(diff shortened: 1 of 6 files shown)


assets/js/gutenberg-syndicated-post.js
```diff
@@ -23,40 +23,40 @@ if (
 		message = sprintf(
 			/* translators: 1) Source of content, 2) Distributor post type singular name. */
 			__(
-				'Distributed from %1$s. This %2$s is linked to the original. Edits to the original will update this version.',
+				'Distributed from %1$s. This %2$s is linked to the origin post. Edits to the origin post will update this remote version.',
```
peterwilsoncc:
```suggestion
				'Distributed from %1$s. This %2$s is linked to the origin %2$s. Edits to the origin %2$s will update this remote version.',
```

This is so the correct post type is displayed, eg podcasts will read:

> Distributed from Distributor Test Site. This Podcast is linked to the origin Podcast. Edits to the origin Podcast will update this remote version.

There is a catch, the `dtGutenberg.postTypeSingular` uses the case for the singular post name label in `register_post_type()`, most commonly with an upper case first letter.

We could convert to lowercase before use but I am not sure how this would work with non-english language sites.

peterwilsoncc (commented):
Just one note for now while I figure out what to do with upper case post type labels. 

It applies throughout though.

assets/js/gutenberg-syndicated-post.js
jeffpaul: I think I'm fine with casing being a spot off on this.  Are there other places throughout this PR that should similarly be updated @peterwilsoncc?

peterwilsoncc: @jeffpaul A few more spots, I'll push some commits so I can make sure the string is available in each location.

peterwilsoncc:
What I hadn't noticed was that post type strings were cast to lowercase in the PHP but not in the JavaScript so there was an inconsistency already. I've pushed 3fc57c08e854d165bd1c5e07830e36a42bde5c00 to make it consistently lowercase.

peterwilsoncc:
I think this is good to go, the strings are consistent across their various uses, the post type names are consistently cased.

Of course, each of the changed lines of code are mine now, so I'll need someone else to review.

cadic (approved):
I've tested new messages with the workflow of:

1. Creating the original
2. Distributing it
3. Linking and unlinking

The code is LGTM

I've discovered a new issue #1021 during my tests, was able to reproduce it on the `develop` branch, so it's not related to the current PR.

PR #1010, opened 2023-01-30
Ensure post abstraction methods use correct site ID.

### LINKED ISSUE #1007: Consider multisite for post abstraction.

peterwilsoncc:
### Description of the Change

This resolves an issue in which the post abstraction can return a mix of data on Multisite networks when the site is switched.

It introduces the `DistributorPost::__call()` magic method to ensure the current site (`get_current_blog_id()`) matches the site ID used to instantiate the object.

To match the PHP behavior without the magic method, a fatal error is thrown if the method name does not exist.

To discourage developers from calling the helper methods while using a switched site, a notice is thrown if the magic method needs to switch sites.

To avoid site switching when possible, an in memory cache is used to bypass the site switch the second or subsequent time a method is called. The cached data is only used if the object is called from the wrong site to reduce the risk of errors caused by the addition/removal of hooks during runtime. Basically it's a performance compromise.

Closes #1007.
Supersedes #1004.

### TODO

* [x] Test interaction of magic methods: ensure `$dt_post->source_site` doesn't result in multiple switches

### Changelog Entry

> Fixed - Inconsistent data returned from post abstraction after site switch on multisite.

### Credits

Props @peterwilsoncc, @dkotter  

### Checklist:

### DIFF
--- a/includes/classes/DistributorPost.php
+++ b/includes/classes/DistributorPost.php
@@ -17,6 +17,13 @@ use WP_Post;
  * includes the phrase Distributor to make it clear to developers `use`ing
  * the class that they are not using the `WP_Post` object.
  *
+ * Developer note: This class uses the `__call()` magic method to ensure the
+ * post data is from the correct site on multisite installs. This is to avoid
+ * repeating code to determine if `switch_to_blog()` is required.
+ *
+ * When adding new methods to this class, please ensure that the method is
+ * protected to ensure the magic method is used.
+ *
  * @since x.x.x
  */
 class DistributorPost {
--- a/includes/classes/DistributorPost.php
+++ b/includes/classes/DistributorPost.php
@@ -95,6 +95,13 @@ class DistributorPost {
 	public $connection_id = 0;
 
 	/**
+	 * The site ID of this post.
+	 *
+	 * @var int
+	 */
+	public $site_id = 0;
+
+	/**
 	 * The source site data for internal connections.
 	 *
 	 * This is an array of site data for the source site. This is set by
@@ -122,7 +129,8 @@ class DistributorPost {
 			return;
 		}
 
-		$this->post = $post;
+		$this->post    = $post;
+		$this->site_id = get_current_blog_id();
 
 		/*
 		 * The original post ID is listed as excluded post meta and therefore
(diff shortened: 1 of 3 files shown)


includes/classes/DistributorPost.php
```diff
@@ -178,6 +186,74 @@ public function __construct( $post ) {
...
+			$switched = true;
+		}
+		$result = call_user_func_array( array( $this, $name ), $arguments );
+		if ( $switched ) {
+			restore_current_blog();
+		}
+		return $result;
+	}
+
+	/**
```
peterwilsoncc, line 279: `::__get()` and `::__isset()` methods are unchanged but have been relocated to group all the magic methods together.

peterwilsoncc:
In the revised PR #1010 (in draft), I went with a `__call` magic method to avoid statics. I'll see if I can use a cache to avoid switching sites when the method is used a second time.

peterwilsoncc:
[Mini plugin I've been using for testing and checking error logs](https://github.com/10up/distributor/files/10564592/nope-call-magic-method.php.txt)

dkotter (approved):

PR #1022, opened 2023-03-02
Toggle Distributor controls when post status changes

### LINKED ISSUE #141: Gutenberg: Extra Refresh to Push Content In Gutenberg

ggutenberg:
### Description of the Change

Renders the Distributor element in the admin bar whether the post status is distributable or not. If it's not, the element is hidden by CSS. When the post status becomes distributable, the element is displayed.

Also adds a `watch` command to the `npm` scripts.

Closes #141 

### How to test the Change
1. Edit a Draft post. Note that the Distributor button in the admin bar is not available, and the Distributor element in the Inspector says "Distribution options available once published".
2. Publish the post. Note that the Distributor button in the admin bar is now available, and the Distributor element has a button to "Distribute page".
3. Reload the editor page. Note that the Distributor button in the admin bar is still available, as is the "Distribute page" button.
4. Click `Switch to draft` and unpublish the post. Note that the admin bar and Distributor Inspector element are as they were in step 1.

### Credits

Props @ggutenberg 

### Checklist:


package.json
```diff
@@ -37,6 +37,7 @@
 	},
 	"scripts": {
 		"build": "wp-scripts build",
+		"watch": "wp-scripts build --watch",
```
peterwilsoncc: The `start` script on line 54 already does this (`wp-scripts start` being an alias).

Did you add this because `watch` is a more common name than `start`? I've been considering the same because it's a more common name.

includes/push-ui.php
```diff
@@ -83,10 +83,6 @@ function syndicatable() {
 		return;
 	}
 
-	if ( ! in_array( $post->post_status, \Distributor\Utils\distributable_post_statuses(), true ) ) {
-		return false;
-	}
-
```
peterwilsoncc, line 89: The syndicate function needs a few more changes. At the moment, an error is thrown on the new post screen as the `$pagenow` sniff is incorrect.

I think something like this in the `is_admin` section is what you need but it probably needs a little fine tuning.

```php
		$current_screen = get_current_screen();

		if ( empty( $current_screen ) || empty( $current_screen->post_type ) ) {
			return false;
		}

		if ( ! in_array( $current_screen->post_type, $distributable_post_types, true ) ) {
			return false;
		}
```

Screen Shot 2023-03-03 at 1 26 26 pm

includes/push-ui.php
```diff
@@ -83,10 +83,6 @@ function syndicatable() {
...
 	if ( ! in_array( get_post_type(), $distributable_post_types, true ) || ( ! empty( $_GET['post_type'] ) && 'dt_ext_connection' === $_GET['post_type'] ) ) { // @codingStandardsIgnoreLine Nonce not required
 		return false;
 	}
```
peterwilsoncc: Not entirely related but more readable. 

```suggestion
	$distributable_post_types = array_diff( $distributable_post_types, array( 'dt_ext_connection' ) );
	if ( ! in_array( get_post_type(), $distributable_post_types, true ) ) {
		return false;
	}
```


It's outside the diff but change `global $post` to `$post = get_post();` above this too. It's a much nicer way of getting the current global post object.

assets/js/gutenberg-plugin.js
```diff
@@ -179,11 +179,16 @@ const DistributorPlugin = () => {
...
+	const distributorTopMenu = document.querySelector(
+		'#wp-admin-bar-distributor'
+	);
+
 	// If we are on a non-supported post status, change what we show
 	if (
 		dtGutenberg.supportedPostStati &&
 		! dtGutenberg.supportedPostStati.includes( postStatus )
 	) {
+		distributorTopMenu.classList.add( 'hide' );
```
peterwilsoncc: Untested but this should avoid errors in the JS if the element is not defined for whatever reason.

```suggestion
		distributorTopMenu && distributorTopMenu.classList.add( 'hide' );
```

assets/js/gutenberg-plugin.js
```diff
@@ -196,6 +201,7 @@ const DistributorPlugin = () => {
 		);
 	}
 
+	distributorTopMenu.classList.remove( 'hide' );
```
peterwilsoncc:
```suggestion
	distributorTopMenu && distributorTopMenu.classList.remove( 'hide' );
```

peterwilsoncc (commented):
I've added a few notes. the main item that needs improvement is to prevent the errors on the new post screen.

When hacking the `syndicatable()` function to test this out (by adding return true at the start ;) I noticed that the post title needs to be updated via JavaScript too.

The UI shows the post title on page load rather than the post title at publication. In this screen shot it shows as "Auto Draft" as that's the default title WP uses. 

Screen Shot 2023-03-03 at 2 15 10 pm

package.json
ggutenberg: Ah, I didn't realize `start` does the same thing. I did a search in `package.json` for the word `watch` and couldn't find anything, so I added it.

Do you want me to remove this or alias it to `npm run start`?

includes/push-ui.php
ggutenberg: I did some debugging and it looks like `$pagenow` is `post-new.php` on a new post. `get_post_type()` and `$current_screen->post_type` both return `page`, `post`, or the slug of a custom post type, and there's already a check happening against `get_post_type()` further down. It looks like the only addition I need here is to check `$pagenow` against `post-new.php` as well, but maybe I'm overlooking something.

assets/js/gutenberg-plugin.js
ggutenberg: eslint doesn't like this syntax, so I used the optional chaining operator.

assets/js/gutenberg-plugin.js
ggutenberg: Same as above - optional chaining operator.

assets/js/gutenberg-plugin.js
peterwilsoncc: > eslint doesn't like this syntax, so I used the optional chaining operator.

NP, works for me.

includes/push-ui.php
peterwilsoncc: > It looks like the only addition I need here is to check `$pagenow` against `post-new.php` as well, but maybe I'm overlooking something.

I think this works just as well, too.

assets/js/gutenberg-plugin.js
```diff
@@ -179,11 +179,23 @@ const DistributorPlugin = () => {
...
+	const distributorTopMenu = document.querySelector(
+		'#wp-admin-bar-distributor'
+	);
+
+	// eslint-disable-next-line no-shadow, react-hooks/rules-of-hooks -- permission checks are needed.
+	const post = useSelect( ( select ) =>
+		select( 'core/editor' ).getCurrentPost()
+	);
+	// Make the post title available to the top menu.
+	dtGutenberg.postTitle = post.title;
```
peterwilsoncc:
```suggestion
	dt.postTitle = post.title;
```

assets/js/push.js
```diff
@@ -344,9 +344,14 @@ jQuery( window ).on( 'load', () => {
 		// Determine if we need to hide the admin bar
 		maybeHideAdminBar();
 
-		if ( distributorPushWrapper.classList.contains( 'loaded' ) ) {
+		// If the post title has changed, we need to reload the template.
+		if (
+			distributorPushWrapper.classList.contains( 'loaded' ) &&
+			dtGutenberg.postTitle === dtGutenberg.previousPostTitle
```
peterwilsoncc:
```suggestion
			dt.postTitle === dt.previousPostTitle
```

assets/js/push.js
```diff
@@ -344,9 +344,14 @@ jQuery( window ).on( 'load', () => {
...
+		) {
 			return;
 		}
+		dtGutenberg.previousPostTitle = dtGutenberg.postTitle;
```
peterwilsoncc:
```suggestion
		dt.previousPostTitle = dt.postTitle;
```

assets/js/push.js
```diff
@@ -394,13 +399,15 @@ jQuery( window ).on( 'load', () => {
 							connections: mustacheData.connections,
 							foundConnections: mustacheData.connections.length,
 							showSearch: 5 < mustacheData.connections.length,
+							postTitle: dtGutenberg.postTitle,
```
peterwilsoncc:
```suggestion
							postTitle: dt.postTitle,
```

assets/js/push.js
```diff
@@ -394,13 +399,15 @@ jQuery( window ).on( 'load', () => {
...
 						}
 					);
 
 					setDisabledConnections();
 				} else {
 					distributorPushWrapper.innerHTML = template( {
 						connections: dtConnections,
+						postTitle: dtGutenberg.postTitle,
```
peterwilsoncc:
```suggestion
						postTitle: dt.postTitle,
```

peterwilsoncc (commented):
I think this is really close, just a few things I've noticed.

Instead of using `dtGutenberg` to store the title, use the `dt` object. That's available wherever the admin bar is: classic editor, block editor and on the front end.

I've pushed 33679559cece65e2f0969d62318f72c2ef3fd104 to include the page title at load in the `dt` object, that will ensure it's present for the front end and the classic editor.

I noticed on the classic editor, the menu bar shows and allows a user to distribute a draft.

dist-new-post

For the classic editor, I think we'll need to do something to prevent the Distributor menu from displaying on unpublished posts (the classic editor uses a full refresh) so it can be hidden. Maybe check `use_block_editor_for_post()`. You'll need to confirm the `classic-editor-remember` meta data has been added to the auto-draft at that point.

I'm trying to figure out if we'll need to protect against users unhiding the menu bar in Gutenberg and pushing a draft/non-distributable status to another site. Unfortunately, I haven't been able to figure out how we might go about this.

Thanks for the progress so far, it's looking really nice.

package.json
peterwilsoncc: > Do you want me to remove this or alias it to `npm run start`?

Let's remove it for now. I'll review our other open-source plugins for the common names, whether it be `dev`, `watch` or something else.

assets/js/gutenberg-plugin.js
ggutenberg: This is strange. If I modify the `const { document, dtGutenberg, MouseEvent } = window;` line and change it to `const { document, dtGutenberg, MouseEvent, dt } = window;`, `dt` is undefined. However if I don't make that modification I'm able to reference it in code directly via `dt.postTitle` or `window.dt.postTitle`. Going with `window.dt.postTitle` for now as it's more explicit and doesn't require an eslint exception, but please let me know if you have a better suggestion.

ggutenberg:
@peterwilsoncc 

> You'll need to confirm the `classic-editor-remember` meta data has been added to the auto-draft at that point.

Are you referring to the draft created on the source side, or on the target side? I've done some digging and familiarized myself with the Classic Editor plugin and how it uses this metadata, but am still unsure what the goal of this requirement is.

peterwilsoncc:
> Are you referring to the draft created on the source side, or on the target side? I've done some digging and familiarized myself with the Classic Editor plugin and how it uses this metadata, but am still unsure what the goal of this requirement is.

I'm referring to the source site, so the check you pushed to `syndicatable()` looks correct but it could do with a comment.

> This is strange. If I modify the `const { document, dtGutenberg, MouseEvent } = window;` line and change it to `const { document, dtGutenberg, MouseEvent, dt } = window;`, `dt` is undefined. However if I don't make that modification I'm able to reference it in code directly via `dt.postTitle` or `window.dt.postTitle`.

It looks like the `dt-push` script is enqueued after the `dt-gutenberg-syndicated-post` script. You might need to force the order by adding `dt-push` as a dependency if it doesn't cause any problems:

```php
$asset_file = DT_PLUGIN_PATH . '/dist/js/gutenberg-syndicated-post.min.asset.php';
// Fallback asset data.
$asset_data = array(
	'version'      => DT_VERSION,
	'dependencies' => array(),
);
if ( file_exists( $asset_file ) ) {
	$asset_data = require $asset_file;
}
$asset_data['dependencies'][] = 'dt-push'; // NEW LINE.

wp_enqueue_script( 'dt-gutenberg-syndicated-post', plugins_url( '/dist/js/gutenberg-syndicated-post.min.js', __DIR__ ), $asset_data['dependencies'], $asset_data['version'], true );
```

ggutenberg:
@peterwilsoncc Thanks. PR updated.

peterwilsoncc (dismissed):
Looks good to me, thanks so much for your patience over the last few days.

I took the liberty of pushing an E2E test in d9f355a1776c723c67bf6b5341c818fd83243b03 to ensure the distributor panel works as expected.

As the block editors selectors change from time-to-time, it may fail in which case I will revert and merge the existing work as is.

peterwilsoncc (approved):

PR #1043, opened 2023-04-06
feat(#524): Add hooks for subscription(s) post push

### LINKED ISSUE #524: Return response from distribution request
We're using distributor in background and calling **\Distributor\Subscriptions\send_notifications** directly. As the function doesn't return any value, we're not able to get information about the distributed post. Is it possible to return the response of the request from this function, so we can know whether it was distributed successfully or not ?
https://github.com/10up/distributor/blob/develop/includes/subscriptions.php#L280

aaronware:
### Description of the Change

This feature adds to hooks for to subscriptions.php `dt_subscription_after_post_push` and `dt_subscriptions_after_post_push` to be able to take action after a post has been pushed to a subscription and after the entire set of subscriptions have been pushed regardless of success or failure.

These hooks are not nearly as robust as `dt_push_post` as I tried to keep the change as slim as possible.

Closes #524 (as potentially stale issue or could be considered a separate issue)

### How to test the Change

utilize 

```
add_action( 'dt_subscriptions_after_post_push' ...
add_action( 'dt_subscription_after_post_push' ...
```

### Changelog Entry

> Added - New hooks to allow developers to extend post push subscriptions

### Credits

Props @aaronware,

### Checklist:

- [x] I have updated the documentation accordingly. **I added docblocks, that should generate accordingly**
- [ ] I have added tests to cover my change. **I need to research what I need to do here**

### DIFF
--- a/includes/subscriptions.php
+++ b/includes/subscriptions.php
@@ -335,13 +335,39 @@ function send_notifications( $post ) {
 				$update_subscriptions = true;
 
 				wp_delete_post( $subscription_id, true );
+
+				/**
+				 * Fires after Distributor pushes a post subscription
+				 *
+				 * @since TBD
+				 * @hook dt_after_set_meta
+				 *
+				 * @param {int}   $post_id         Post ID
+				 * @param {int}   $subscription_id Subscription ID
+				 * @param {array} $request         Request array
+				 */
+				do_action( 'dt_subscription_after_post_push', $post_id, $subscription_id, $request );
 			}
+
+
 		}
+
 	}
 
 	if ( $update_subscriptions ) {
 		update_post_meta( $post_id, 'dt_subscriptions', $subscriptions );
 	}
+
+	/**
+	 * Fires after Distributor completes all subscriptions regardless of success/failure
+	 *
+	 * @since TBD
+	 * @hook dt_after_set_meta
+	 *
+	 * @param {int}   $post_id       Post ID
+	 * @param {array} $subscriptions Array of Subscriptions
+	 */
+	do_action( 'dt_subscriptions_after_post_push', $post_id, $subscriptions );
 }
 
 /**


aaronware:
I can review this draft request and the 2.0 project to make sure this is still accurate. It's been a while since I looked at this

jeffpaul:
@aaronware are you interested in continuing the extension work here (assuming primarily for deeper Woo support)?

aaronware:
@jeffpaul yes definitely. Now that 2.x has been out in the wild I can revisit. My apprehension before and why I drafted this was adding to the 1.9.x branch when so much was changing in 2.x. There are some quick wins to add hooks like what I drafted, however, I wasn't sure if you all wanted something more structured when integrating with Distributor for other plugins similar to Steam for example

includes/subscriptions.php
```diff
@@ -335,13 +335,39 @@ function send_notifications( $post ) {
...
+	/**
+	 * Fires after Distributor completes all subscriptions regardless of success/failure
+	 *
+	 * @since TBD
+	 * @hook dt_after_set_meta
+	 *
+	 * @param {int}   $post_id       Post ID
+	 * @param {array} $subscriptions Array of Subscriptions
+	 */
+	do_action( 'dt_subscriptions_after_post_push', $post_id, $subscriptions );
```
faisal-alvi, line 370: Should we also include `$update_subscriptions` as a third parameter so users can be informed of the success or failure status?

includes/subscriptions.php
```diff
@@ -335,13 +335,39 @@ function send_notifications( $post ) {
...
+				 * Fires after Distributor pushes a post subscription
+				 *
+				 * @since TBD
+				 * @hook dt_after_set_meta
+				 *
+				 * @param {int}   $post_id         Post ID
+				 * @param {int}   $subscription_id Subscription ID
+				 * @param {array} $request         Request array
+				 */
+				do_action( 'dt_subscription_after_post_push', $post_id, $subscription_id, $request );
```
faisal-alvi, line 349: Since the status is currently failure at this point, should we consider tweaking the filter name to reflect this?

faisal-alvi (changes requested):

PR #1063, opened 2023-05-30
Add snippet for setting meta data automatically.

### LINKED ISSUE #971: Provide example for setting custom meta data in snippits.

The `dt_after_set_meta` action can be used for extenders to set custom meta data on distributed posts. This is useful for both extenders requiring custom meta data or wishing to alter a Distributor default value.

Adding an example for this to the code snippets could be helpful.

Follow up to #965.

### Designs

_No response_

_No response_

### Code of Conduct

peterwilsoncc:
### Description of the Change

Provides snippet for setting both Distributor and custom meta data using the `dt_after_set_meta` hook.

Closes #971

### Changelog Entry

> Added - Snippets for setting meta data automatically.

### Credits

Props @peterwilsoncc, @iamdharmesh. 

### Checklist:

### DIFF
--- a/includes/classes/ExternalConnections/WordPressExternalConnection.php
+++ b/includes/classes/ExternalConnections/WordPressExternalConnection.php
@@ -428,7 +428,7 @@ class WordPressExternalConnection extends ExternalConnection {
 				update_post_meta( $new_post, 'dt_full_connection', true );
 			}
 
-			if ( ! empty( $post_array['meta'] ) ) {
+			if ( isset( $post_array['meta'] ) && is_array( $post_array['meta'] ) ) {
 				// Filter documented in includes/classes/InternalConnections/NetworkSiteConnection.php.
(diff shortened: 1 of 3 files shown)


includes/classes/ExternalConnections/WordPressExternalConnection.php
```diff
@@ -428,7 +428,7 @@ public function pull( $items ) {
 				update_post_meta( $new_post, 'dt_full_connection', true );
 			}
 
-			if ( ! empty( $post_array['meta'] ) ) {
+			if ( isset( $post_array['meta'] ) && is_array( $post_array['meta'] ) ) {
```
peterwilsoncc, line 431: The empty check was preventing the hook from running if there was no meta because `empty( [] ) === true`. My understanding is that the hook should fire regardless of whether there is or is not meta so I've changed it to fire when the array is set.

iamdharmesh (approved):
Thanks for adding this @peterwilsoncc. Snippets working fine. LGTM 🚀

Issue #975, opened 2022-11-22
External connections do not distribute media


peterwilsoncc:
### Describe the bug

When pushing or pulling content from an external connection, no media is distributed regardless of the settings in the admin.

This occurs regardless of whether the featured image is attached to the post or not.

Internal/network connections do distribute the featured images.

**Updates:**
* Please note that it seems to be more of an issue with in-content images and not featured images.
* See the test results in https://github.com/10up/distributor/issues/975#issuecomment-1382566549 and work on their fixes.

### Steps to Reproduce

1. Create two sites single sites running Distributor
2. Connect them bidirectionally (ie, so each site has the other as an external connection)
3. Create a post on one site
4. Add and in-content image(s)
5. publish the post
6. Either
   a. Push the post from the first sites screen
   b. Pull the post from the second sites admin
9. Open the pulled post in the editor, and observe there is no in-content image(s).

### Screenshots, screen recording, code snippet

**Source site**
![Screen Shot 2022-11-22 at 2 09 03 pm](https://user-images.githubusercontent.com/519727/203212356-24704be1-c967-4bc4-918c-2159d69f3b2a.png)

**Pulled content**
![Screen Shot 2022-11-22 at 2 10 15 pm](https://user-images.githubusercontent.com/519727/203212373-155c6160-f304-47d6-b07f-9bacbf0cf9ca.png)

**Pushed content**
![Screen Shot 2022-11-22 at 2 11 10 pm](https://user-images.githubusercontent.com/519727/203212383-b582e062-2190-4d94-8edf-013213d77a46.png)

### WordPress information

- Distributor Develop (at 3bb3e3f0f73573b6e7586478530b97568c4db430 )
- WordPress 6.1.1

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gleysen:
Hope this gets fixed soon.. really need this as well...

faisal-alvi:
@peterwilsoncc I've performed the steps you suggested above, and I've seen that media is distributed between external connections without any problems (Tested on the "develop" branch). Please have a look at the video and let me know if there is anything I missed.

https://user-images.githubusercontent.com/25176325/212026447-bad4f4b7-30c7-4591-a2c9-d0df4f49a5b1.mp4

gleysen:
Hey Faisal,
[...]

PR #1110, opened 2023-07-06
Set scheme before downloading relative scheme URLs.

peterwilsoncc:
### Description of the Change

Before calling the WordPress function `download_url()` in `Utils\process_media()` check for relative scheme URLs and set a scheme if required.

WordPress isn't a browser so will throw an error if a request is made to download a relative URL. Defaulting to `http` will ensure the most likelihood of success, if the site redirects to an HTTPS URL then WordPress will follow the redirect.

See #975

### How to test the Change

1. Set up Distributor on two WordPress sites
2. In the wp-config.php file for each site, define `WP_CONTENT_URL` to use a relative scheme URL: `define( 'WP_CONTENT_URL', '//' . $_SERVER['HTTP_HOST'] . '/wp-content' );`
3. Connect each site as an external connection to the other
4. On site A, publish a post with a featured image.
5. On site B, pull the post from site A using Distributor

On this branch the post should include the featured image, on `develop` the featured image should be missing.

### Changelog Entry

> Fixed - Account for relative scheme URLs when processing media.

### Credits

Props @peterwilsoncc

### Checklist:

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -924,6 +924,11 @@ function process_media( $url, $post_id, $args = [] ) {
 	// Default for external or if a local file copy failed.
 	if ( $download_url ) {
 
+		// Set the scheme to http: if a relative URL is specified.
+		if ( str_starts_with( $url, '//' ) ) {
+			$url = 'http:' . $url;
+		}
+
 		// Allows to pull media from local IP addresses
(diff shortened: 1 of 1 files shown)


ravinderk (approved):
@peterwilsoncc I could not reproduce the issue but did not see the side effect of a minor change. For this reason, I am approving the pull request.

PR #1124, opened 2023-08-22
Update `yahnis-elsts/plugin-update-checker` from 4.13 to 5.1

dkotter:
### Description of the Change

The Plugin Update Checker library we use is also used on a few other 10up projects.  We need to ensure we are using the same version across all of these or you can end up with conflicts (especially when using composer to manage all dependencies of a project). This PR updates this library to the latest version, which will also be done in other repos that are using this library.

### How to test the Change

1. Pull these changes down
2. Within the `distributor.php` file, set the version number to something lower, like 1.8.0
3. Ensure you have a proper license in place
4. Go to the plugins section and ensure you see an update notice for Distributor (may need to click the `Check for updates` link; also ensure the plugin is active)
5. Click on the `View version X.X.X details` link and ensure you see proper details
6. Go to the WordPress Updates page and ensure you see Distributor there as well, with the icon
7. Click on the `View version X.X.X details` link and ensure you see proper details
8. If desired, click the `update now` link and ensure an update works (either from the plugins page or the main updates page). Note though that this will wipe out your version controlled `distributor` directory, so proceed with caution

### Changelog Entry

> Changed - Upgrade the Plugin Update Checker library from 4.13 to 5.1

### Credits

Props @dkotter

### Checklist:


peterwilsoncc (commented):
@dkotter After running `composer install`, I am seeing modifications for the files:

* vendor/yahnis-elsts/plugin-update-checker/.gitattributes
* vendor/yahnis-elsts/plugin-update-checker/phpcs.xml

They appear to be white space/EOL type related -- are able to check if you get the same result as it would be good to avoid having to revert the changes each time if possible. 

Based on manual testing, the changes seem to be working.

dkotter:
@peterwilsoncc 

> They appear to be white space/EOL type related -- are able to check if you get the same result as it would be good to avoid having to revert the changes each time if possible.

I don't get this myself, if it is just EOL things, wondering if there's a difference between our setups in regards to that? 

Edit: I guess also to add, looking at those files directly in the source repo, they appear to match what's in this PR. [phpcs.xml](https://github.com/YahnisElsts/plugin-update-checker/blob/v5.1/phpcs.xml) and [.gitattributes](https://github.com/YahnisElsts/plugin-update-checker/blob/v5.1/.gitattributes)

peterwilsoncc:
> don't get this myself, if it is just EOL things, wondering if there's a difference between our setups in regards to that?

I suspect it's this -- it's something WordPress has needed to handle in the past in either the editor config or gitattributes files.

peterwilsoncc:
I think we can resolve the differences between systems by enforcing the line ending types via the gitattributes file:

```
# Text files should have LF line endings.
* text eol=lf

# Binary files should not be modified.
*.mo binary
*.png binary
*.jpg binary
```

It may be worth implimenting an editor config file similar to the [one used in wordpress-develop](https://github.com/WordPress/wordpress-develop/blob/71d2050014676ea16e873991acb0f1e9d0444f69/.editorconfig#L0-L1).

dkotter:
> I think we can resolve the differences between systems by enforcing the line ending types via the gitattributes file:

I've added this now, though I still didn't get any differences in line endings in the files you saw, so not sure if this fixed things.

> It may be worth implimenting an editor config file similar to the [one used in wordpress-develop](https://github.com/WordPress/wordpress-develop/blob/71d2050014676ea16e873991acb0f1e9d0444f69/.editorconfig#L0-L1).

We already had an `.editorconfig` in place but it was missing the last line from `wordpress-develop`, which I've added now (this shouldn't make a difference for this particular PR though, as that only applies to .txt files and the wp-config-sample.php file).

peterwilsoncc:
composer install is still causing dirty files for me.

Let's merge we have now and I'll remember to do a checkout of the committed files each time I run composer install. I often have to do it for package-lock files so am used to it by now.

peterwilsoncc (approved):

PR #1143, opened 2023-10-11
Copy .htaccess file when initilizing multisite on test container.

peterwilsoncc:
### Description of the Change

Ensures the htaccess file is copied when the test container's Docker environment is converted to multisite.

Closes #

### How to test the Change

1. Start docker
2. `npm run env:destroy` to reset the environment
3. `npm run env:start`
4. Visit http://localhost/second/ to ensure you can access the second site on the test environment.

### Changelog Entry

> Fixed - Copy the htaccess when initilizing the E2E test environment.

### Credits

Props @peterwilsoncc, @theskinnyghost, @dkotter.

### Checklist:

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -96,9 +96,6 @@ jobs:
         npx wp-env run cli "wp core version"
         npx wp-env run cli "php --version"
 
-    - name: Copy .htaccess
-      run: npm run copy-htaccess
-
     - name: Test
       run: npm run cypress:run
 
--- a/tests/bin/initialize.sh
+++ b/tests/bin/initialize.sh
@@ -17,4 +17,5 @@ else
 	npm run env run tests-cli "wp site create --slug=second --title='Second Site' --email='second@admin.local'"
 	npm run env run tests-cli "wp theme enable twentytwentyone --activate"
 	npm run env run tests-cli "wp theme enable twentytwentyone --url=localhost/second --activate"
+	npm run env run tests-cli "cp wp-content/plugins/distributor/tests/cypress/.htaccess .htaccess"
 fi


theskinnyghost (dismissed):

dkotter (commented):
Curious on why this is needed? I know we already have an npm script that copies the htaccess file before tests are run (see https://github.com/10up/distributor/blob/develop/.github/workflows/cypress.yml#L100 and https://github.com/10up/distributor/blob/develop/package.json#L63) so wondering if that's not working?

peterwilsoncc:
@dkotter It's for running the tests locally rather than in CI. It allows devs to run `npm run env:start; npm run cypress:run` locally.

At the moment `env:start` does the conversion to multisite but misses the step for copying the htaccess file.

dkotter:
> @dkotter It's for running the tests locally rather than in CI. It allows devs to run `npm run env:start; npm run cypress:run` locally.
> 
> At the moment `env:start` does the conversion to multisite but misses the step for copying the htaccess file.

Okay, I guess the question I have then is should we remove the copying of the htaccess file from the `cypress.yml` file then? I believe at the moment that is happening twice, once as part of the `postenv:start` command and then again in the `cypress.yml` file when it runs `npm run copy-htaccess`. I haven't tested this locally but seems like we can get rid of that second one now.

.github/workflows/cypress.yml
```diff
@@ -99,9 +99,6 @@ jobs:
     - name: Copy .htaccess
       run: npm run copy-htaccess
 
-    - name: Test
```
dkotter, line 102: Looks like the wrong step was removed. This test step should stay and I think the step above should be removed

dkotter (changes requested):

.github/workflows/cypress.yml
peterwilsoncc: Fixed in 9b7c613c9e14fea43efad76196010d5e7c4054d5, I force pushed the incorrect commit to oblivion to keep the commit history clean.

dkotter (approved):

PR #1156, opened 2023-11-21
Account for unpublished post statuses.

### LINKED ISSUE #1149: Filter to allow push/pull of draft/scheduled posts in network/external connected sites
[Edited 2023-11-29 by @jeffpaul]
I'm updating the title and the description in this area here to note we should review the push/pull options to allow for a filter that is disabled/false by default but could be set to true that would enable drafts and scheduled posts to be able to be pushed/pulled in a network and external connected sites.

### Describe the bug

In previous versions (for example 1.9.1 for sure) we were able to view both published and draft posts in the Pull Content list on the recipient website.   In version 2.0.1 we can no longer see any posts that are 'draft' status in the Pull Content list.  Posts only become available to pull once they are published.

Is this feature change intended?  Or a bug.
It's an issue for us, since we sometimes wish to pull posts from one source to another before they are published.
If there's a workaround (or a hook) that would be nice. 

### Steps to Reproduce

On the receiving site go to Distributor > Pull Content > Pull Content from (choose your site) > New.
We see only posts that have either been published on the source website or pulled posts.  No posts in draft status are available to pull. 

_No response_

### Environment information

_No response_

### WordPress information

WordPress 6.3.2
Distributor 2.0.1 (on all sites)

### Code of Conduct

peterwilsoncc:
### Description of the Change

Fixes various issues relating to post statuses:

* (network) when pulling scheduled posts, set the post date per original site
* (network) when pulling draft posts, ensure the post is not published when "pull as draft" is unchecked

> **Note**
>
> End-to-end tests are passing but the code to push the summary to the GitHub action's summary page is failing. This is reporting a false negative for successful tests.

See #1149

### How to test the Change

1. Install Distributor on a Multisite Install
2. Create a sub site
3. Generate some draft posts on the main site: `wp post generate --count=5 --post_title="MS1 Draft" --post_author=1 --post_status=draft`
4. Generate some scheduled posts on the main site: `wp post generate --count=5 --post_title="MS1 Scheduled" --post_author=1 --post_status=future --post_date="2123-11-20 21:46:18"`
5. Open the sub site's dashboard
6. Go to Distributor > Pull Content
7. Uncheck "pull as draft"
8. Pull a scheduled post via the post's actions
9. Return to new posts
10. Uncheck "pull as draft"
11. Pull a draft post via the post's action
12. Return to new posts
13. Uncheck "pull as draft"
14. Check multiple scheduled posts' checkbox
15. Select "pull" from the bulk action drop down
16. Click "Apply"
17. Return to new post
18. Uncheck "pull as draft"
19. Check multiple draft posts' checkbox
20. Select "pull" from the bulk action drop down
21. Click "Apply"
22. Go the the All posts screen
23. Ensure draft posts remain in draft
24. Ensure scheduled posts are scheduled

### Changelog Entry

> Fixed - Prevent early publishing of scheduled posts when pulled internally.

### Credits

Props @peterwilsoncc, @jeffpaul, @dkotter, @maxledoux, @pcrumm 

### Checklist:

### DIFF
--- a/includes/classes/DistributorPost.php
+++ b/includes/classes/DistributorPost.php
@@ -860,14 +860,16 @@ class DistributorPost {
 		if ( ! empty( $args['remote_post_id'] ) ) {
 			// Updating an existing post.
 			$insert['ID'] = (int) $args['remote_post_id'];
+			// Never update the post status when updating a post.
 			unset( $insert['post_status'] );
-		}
-
-		if ( ! empty( $args['post_status'] ) ) {
+		} elseif ( ! empty( $args['post_status'] ) ) {
 			$insert['post_status'] = $args['post_status'];
 		}
 
-		if ( 'future' === $insert['post_status'] ) {
+		if (
+			isset( $insert['post_status'] )
+			&& 'future' === $insert['post_status']
+		) {
 			// Set the post date to the future date.
 			$insert['post_date']     = $post_data['date'];
 			$insert['post_date_gmt'] = $post_data['date_gmt'];
--- a/includes/classes/DistributorPost.php
+++ b/includes/classes/DistributorPost.php
@@ -782,6 +782,9 @@ class DistributorPost {
 			'excerpt'                        => $this->post->post_excerpt,
 			'parent'                         => ! empty( $this->post->post_parent ) ? (int) $this->post->post_parent : 0,
 			'status'                         => $this->post->post_status,
+			'date'                           => $this->post->post_date,
+			'date_gmt'                       => $this->post->post_date_gmt,
+
 			'distributor_media'              => $this->get_media(),
 			'distributor_terms'              => $this->get_terms(),
 			'distributor_meta'               => $this->get_meta(),
@@ -862,6 +865,12 @@ class DistributorPost {
 
 		if ( ! empty( $args['post_status'] ) ) {
 			$insert['post_status'] = $args['post_status'];
+		}
+
+		if ( 'future' === $insert['post_status'] ) {
+			// Set the post date to the future date.
+			$insert['post_date']     = $post_data['date'];
+			$insert['post_date_gmt'] = $post_data['date_gmt'];
 		}
 
 		// Post meta used by wp_insert_post, wp_update_post.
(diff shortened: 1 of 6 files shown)


jeffpaul:
Let's also credit `@maxledoux` and `@pcrumm` for reporting the issue here as well.

dkotter:
@peterwilsoncc This looks good to me and seems to fix the issues mentioned.

That said, one of the reported issues was around scheduled posts, in particular this was the reported problem:

1. Within a multisite, create a scheduled post on Site A
2. Either Push (though by default you can only Push published items) or Pull this content into another site on the network (Site B)
3. On Site B, this post comes over in a published state, which is not desired
4. If you then change this post on Site B to be a draft and then later update the original post on Site A (but keep it in a scheduled state), the post on Site B becomes published again

I tested these steps both on the `develop` branch and on this PR branch. In both cases, I couldn't reproduce the first reported problem. On both branches, Pushing or Pulling scheduled content brought that content over in a draft state, not a published state.

But I was able to reproduce the second problem both on `develop` and on this branch, so that part still needs addressed. I Pulled a scheduled post from Site A into Site B, it came over as a draft, I then updated the post on Site A and the post on Site B changed to published.

dkotter:
Sorry, just to add more details to ^ above as I tested this a bit more.

If when Pulling or Pushing you leave the `Pull/Push as Draft` setting on, it will bring the item in a draft state, both on this PR and on `develop` (which is correct behavior). If you turn that setting off, on `develop`, it brings over scheduled items as published. With this PR, it brings over scheduled items as scheduled, which is correct.

So the remaining issue here is if you Pull a scheduled item in as a draft (which is the default) and you then update the original scheduled item, it turns the status from draft to published. I think ideal would be keeping it as draft but even changing it to scheduled would be better

peterwilsoncc:
> So the remaining issue here is if you Pull a scheduled item in as a draft (which is the default) and you then update the original scheduled item, it turns the status from draft to published. I think ideal would be keeping it as draft but even changing it to scheduled would be better

In ee56b99dcc587641641a2aabd13b1a897e645bcd I've pushed a change to prevent the post status from being updated when updating a post. The post status will always need to be changed on the distributed copy of the content.

Testing notes:

1. Pull scheduled post as draft
2. Update scheduled post on origin site
3. Ensure post remains in draft on destination site
4. Push published post as draft
5. Ensure post is in draft state on destination site
6. Update post on origin site
7. Ensure post remains in draft on destination site.

dkotter (approved):

PR #1159, opened 2023-11-21
Avoid fatal error when Pulling an item that has been Pulled already on a different site

### LINKED ISSUE #1158: Fatal error due to using `array_key_exists` on a non-array value
### Describe the bug

In a recent demo of Distributor, a PHP fatal error occurred when trying to Pull content from one site into another (within a multisite). The error we got was `Fatal error: array_key_exists() expects parameter 2 to be array, null given`, pointing to this line of code: https://github.com/10up/distributor/blob/8e46f93f35130624819044d1a825a9c905fc4858/includes/classes/InternalConnections/NetworkSiteConnection.php#L285

I originally was not able to reproduce this, either in the demo environment or locally. But in testing something else I ran across the same issue and I think I can reliably reproduce now.

If you're using Internal Connections and you Pull a piece of content from one site to another (say from Site A into Site B), if you then try and Pull that same piece of content into another site (say from Site A into Site C), this error occurs.

In looking at the structure of the `dt_connection_map` meta, this is what I see:

```
array(
	'internal' => array(
		1 => array(
			'post_id' => 123,
			'time'    => 123456789,
		),
	),
	'external' => array(),
)
```

In looking at/around the mentioned code, we iterate over the meta values and look for a key of `internal`. But that key doesn't exist within the inner array and thus the error occurs.

I don't know if this is due to a structure change within this meta as part of the v2 release or if this is a long standing bug that just hasn't been noticed. EDIT: seems to be a PHP 8 thing, whereas previously this would throw a PHP warning but now throws a fatal.

I think we can fix this by doing something like this:

```php
foreach ( $post_props['meta']['dt_connection_map'] as $connection_type => $distributed ) {
	$distributed = maybe_unserialize( $distributed );

	if ( 'internal' === $connection_type && array_key_exists( $current_blog_id, $distributed ) ) {
		$dt_pull_messages['duplicated'] = 1;
		continue 2;
	}
}
```

### Steps to Reproduce

Not sure. The issue happened when Pulling content from one site into another, within a multisite environment. But have not been able to reproduce locally

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

dkotter:
### Description of the Change

Within a multisite instance, if you Pull an item from one site to another (say from Site A to Site B), the original item has the connection details stored in the `dt_connection_map` meta.

If you then pull that same item into a different site (say from Site A into Site C) we check if the original item has that meta and if so, we check to see if the current site ID exists within that meta (which would indicate a duplicate item).

The problem here is the code we run looks for a different structure than what we actually have. Unsure if this is a new bug coming out of the v2 release or if this is a long standing issue that just hasn't been found. **EDIT**: Tested further and seems to be something that was always an issue but PHP 8.1 (maybe 8.0 but didn't test that) started throwing a fatal error instead of just a warning. So anyone running PHP 7.4 or less wouldn't see this.

This PR fixes this by using the correct data structure.

Closes #1158 

### How to test the Change

1. Set up a WordPress multisite with Distributor active and at least three sites in place
2. Create a piece of content on the first site
3. On the latest released version of Distributor, Pull this content into the second site
4. Then try pulling this content into the third site. A PHP fatal error should happen
5. Check out this PR, create a new piece of content on the first site and run through the steps again
6. No PHP fatals should happen

### Changelog Entry

> Fixed - Avoid a PHP fatal error when pulling content that has previously been pulled into a different network site

### Credits

Props @dkotter, @jeffpaul 

### Checklist:

### DIFF
--- a/includes/classes/InternalConnections/NetworkSiteConnection.php
+++ b/includes/classes/InternalConnections/NetworkSiteConnection.php
@@ -279,10 +279,10 @@ class NetworkSiteConnection extends Connection {
 			$current_blog_id = get_current_blog_id();
 
 			if ( ! empty( $post_props['meta']['dt_connection_map'] ) ) {
-				foreach ( $post_props['meta']['dt_connection_map'] as $distributed ) {
+				foreach ( $post_props['meta']['dt_connection_map'] as $connection_type => $distributed ) {
 					$distributed = maybe_unserialize( $distributed );
 
-					if ( array_key_exists( $current_blog_id, $distributed['internal'] ) ) {
+					if ( 'internal' === $connection_type && array_key_exists( $current_blog_id, $distributed ) ) {
 						$dt_pull_messages['duplicated'] = 1;
 						continue 2;
 					}


peterwilsoncc (approved):
This looks good to me and resolves the error listed in the ticket.

While testing, I did notice that these lines in the same method can also trigger a fatal error

https://github.com/10up/distributor/blob/4fe15ccd41ac4733d1676caae41eaadbc918a832/includes/classes/InternalConnections/NetworkSiteConnection.php#L270-L275

`remote_get()` can return `false` if the post is an invalid ID and the unexpected value isn't caught by the `is_wp_error()` condition. This subsequently passes the value to a number of functions that do not handle booleans, triggers a few warnings and finally a fatal error.

I'm happy if you want to merge this and follow up in another PR for the issue of an invalid post ID.

PR #1160, opened 2023-11-24
Fix and update "keep original publish date" snippet in the docs

leogermani:
Updating the docs for the publication date snippet.

* Fixes the current snippet: The argument expected by the core API is `date` and not `post_date`. Also adding `date_gmt` for completion
* Adds the counterpart snippet for Pulls

### Description of the Change

Fixes one of the snippets in the docs

### Credits

Props @leogermani 

### Checklist:


peterwilsoncc (commented):
Good catch but I think it's a little more complicated.

The same filter is used for both internal connections (via `wp_insert_post`) and external connections (via the REST API).

The forth parameter sent to the filter is the connection, so the code snippet probably needs to be conditional with an `is_a()` check for whether it's an internal or external push connection.

The pull post snippet looks good.

leogermani:
Thanks for the clarification @peterwilsoncc , fixed the snippet

docs/snippets.md
```diff
@@ -94,11 +94,27 @@ add_action( 'plugins_loaded', function() {
  *
  * This filter is used to filter the arguments sent to the remote server during a push. The below code snippet passes the original published date to the new pushed post and sets the same published date instead of setting it as per the current time.
  */
-add_filter( 'dt_push_post_args', function( $post_body, $post ) {
-    $post_body['post_date'] = $post->post_date;
+add_filter( 'dt_push_post_args', function( $post_body, $post, $args, $connection ) {
+
+    // When pushing to an external connection, we use the REST API, so the name of the field is `date`.
+    // But when pushing to an internal connection, the attributes are sent to wp_insert_post, which expects `post_date`.
+    $field_prefix = is_a( $connection, 'Distributor\ExternalConnections\WordPressExternalConnection' ) ? '' : 'post_';
```
peterwilsoncc:
```suggestion
    $field_prefix =( $connection instanceof \Distributor\ExternalConnections\WordPressExternalConnection ) ? '' : 'post_';
```

This is a bit of a nitpick but [PHP have tried to deprecate `is_a()` a couple of times](https://github.com/PHPCSStandards/PHPCSExtra/issues/260) so it's nice to avoid it in case another attempt is made.

peterwilsoncc (commented):
Thanks @leogermani 

This looks good to me, I've added a bit of a nitpick. I'll commit that and merge the PR.

peterwilsoncc (approved):

docs/snippets.md
peterwilsoncc: BTW: I know I gave you the bad steer on this -- sorry about that. I had a dumb moment.

PR #1179, opened 2024-01-15
For non-super-admins, use get_blogs_of_user() to determine authroized sites.

### LINKED ISSUE #1176: Method for assembling list of internal connections does not work well on large installations
### Describe the bug

The mechanism used by Distributor to assemble a list of internal connections (used on the 'Pull Content' dropdown and elsewhere) uses `get_sites( [ 'number' => 1000 ] )` to pull a list of all sites on the network, then iterates through to see whether the current user has posting permissions on any of those sites. This strategy does not work well on the sorts of installations where there are thousands of sites, since any site beyond the initial 1000 will never appear in the list. (This came up in the context of a client site where the network has many thousands of sites, but the vast majority of users are members of only a few handful of them.)

By contrast, if one uses `get_blogs_of_user()` or some other WP mechanism that only identifies the sites on which a user has a role, the list will contain all of a user's sites. (As a bonus, it performs far better.) 

This change would mean, however, that a super admin would not see sites where the super admin does not have a role. This might not work well for some use cases of the plugin.

For the time being, I've effected the `get_blogs_of_user()` change using the 'dt_pre_get_authorized_sites' filter. But I thought I'd open the ticket, because I suspect that the use case I described (large networks where users are members of many sites) is perhaps more common than the situation best served by the current strategy (a super admin who needs access to all sites). In any case, I would argue that a dropdown containing 1000 items is not particularly useful, and in the case where `wp_is_large_network()` (or something similar) this interface ought to be an AJAX-powered type-ahead or something like that.

Thanks for the plugin :-D

### Steps to Reproduce

1. Have a large network
2. Log in as a user who owns at least two sites. At least one of them (site A) should have ID > 1000
3. Go to Site B > Dashboard > Distributor > Pull Content
4. Site A will not be in the dropdown

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

peterwilsoncc:
### Description of the Change

Modifies the query used to determine a user's authorized sites to use `get_blogs_of_user()`.

As super admins have access to all sites without being a member, the existing `get_sites()` query is used for super admins.

Closes #1176

### How to test the Change

1. Install Distributor on a Multisite Install in sub-directory mode
2. Make it big `export XDEBUG_MODE=off; for i in {1..1050}; do wp site create --slug=$i --title="MS $i" --url=https://ms-distributor.local; done;` (change the URL if needs be)
3. Create a non super-admin user
4. Add the new user as an administrator on sites 2, 3 and 1050
5. Log in/switch to the new user
6. Visit the pull screen of either site 2 or 3.
7. The last site should be displayed (1050)

Creating the sites may take some time so you should leave the command going and switch tasks for a while.

### Changelog Entry

> Fixed - Improved database query for getting a users authorized sites on very large networks.

### Credits

Props @jeffpaul, @boonebgorges, @peterwilsoncc.

### Checklist:


kirtangajjar:
@peterwilsoncc Tested and ensured that the PR is working for me.

kirtangajjar:
@peterwilsoncc It seems that the 1000 site limit problem will still be there for super admins, so why not query for number of sites like this and store it in transients instead of passing a static number:
```
SELECT COUNT(blog_id) FROM {$wpdb->blogs} WHERE archived = '0' AND spam = '0' AND deleted = '0'
```

peterwilsoncc:
Thanks for looking at this @kirtangajjar.

The limit of 1000 sites was added in 3077cf9ae6647fc0a15ceda9b4f02813aed9a728. Based on the commit message it was a coding standards fix to avoid a warning for an unlimited query.

I'd be surprised if many users were hitting the 1000 site limit but do you think it's worth adding a filter to allow developers to increase the size of the dropdown?

jeffpaul:
Fwiw filtering for that seems like a good handling to me.

vikrampm1:
@Sidsector9 checking if you can help with the review here?

includes/classes/InternalConnections/NetworkSiteConnection.php
```diff
@@ -892,12 +892,38 @@ public static function build_available_authorized_sites( $user_id = false, $cont
...
-				array(
-					'number' => 1000,
-				)
-			);
-			$current_blog_id  = (int) get_current_blog_id();
+
+			if ( is_super_admin() ) {
+				$sites = get_sites(
+					array(
+						'number' => 1000,
```
dkotter: Since the original issue mentions a site that has more than 1000 sites on the network, I do think we should add a filter around this so people can increase this limit if needed

peterwilsoncc: `dt_max_authorized_sites` filter added in 97f71ca4e0c654f0d311aa306bca51ea0b9ad568

dkotter (approved):

PR #1181, opened 2024-01-17
add filter to DistributorPost::to_pull_list

leogermani:
### Description of the Change

Adds a new hook to `DistributorPost::to_pull_list` method to allow third party plugins to add additional data to the post being pulled by another site.

This is a useful alternative to extend the post data when pulling posts. Apart from that, the only other way plugins can add data to posts is by writing data to the post meta table, which is not always convenient.

### How to test the Change

In the site where the content is going to be pulled from, add a snippet to add custom data to the post:
```PHP
add_filter( 'dt_post_to_pull', function( $post_data ) {
    $post_data['custom_field'] = 'custom_value';
} );
```

Now in the site where you are pulling the content from, add a snippet to inspect the contents of the pulled post and make sure the custom data is there:
```PHP
add_action( 'dt_pull_post', function ( $new_post_id, $connection, $post_array ) {

    if ( isset( $post_array['custom_field']) && $post_array['custom_field'] == 'custom_value') {
        error_log( 'It worked!' );
    }

} );
```

### Changelog Entry

> Added - `dt_post_to_pull` filter that allows plugins to add custom data to a post being pulled by another site.

### Credits

Props @leogermani

### Checklist:

I suppose the documentation is automatically generated in this case. Doc block looks good.

Also suppose this does not require new tests.

### DIFF
--- a/includes/classes/DistributorPost.php
+++ b/includes/classes/DistributorPost.php
@@ -918,7 +918,17 @@ class DistributorPost {
 		$display_data['distributor_original_site_name'] = $this->source_site['name'];
 		$display_data['distributor_original_site_url']  = $this->source_site['home_url'];
 
-		return $display_data;
+		/**
+		 * Filters the post data for when they are being formated for a pull
+		 *
+		 * @since 2.0.3
+		 * @hook dt_post_to_pull
+		 *
+		 * @param {array} $display_data The post data.
+		 *
+		 * @return {array} Modified post data.
+		 */
+		return apply_filters( 'dt_post_to_pull', $display_data );
 	}
 
 	/**


peterwilsoncc:
@leogermani Thanks for the pull request.

Is the background to this request to ensure the custom distributor endpoints account for data added via the [`register_rest_field()`](https://developer.wordpress.org/reference/functions/register_rest_field/) function in WordPress?

I realise that pulling from the custom endpoint fails to consider that for both posts and terms so we may wish to include all such data in the distributor endpoint now we are using it to pull content.

leogermani:
> @leogermani Thanks for the pull request.
> 
> Is the background to this request to ensure the custom distributor endpoints account for data added via the [`register_rest_field()`](https://developer.wordpress.org/reference/functions/register_rest_field/) function in WordPress?
> 
> I realise that pulling from the custom endpoint fails to consider that for both posts and terms so we may wish to include all such data in the distributor endpoint now we are using it to pull content.

Hi @peterwilsoncc,

No, not really.

It's just a way to be able to add custom data to a post that is being pulled. I realize I could have data sent if I added them as post meta, but that's not very convenient when you are running the plugin in a site with hundreds or thousands of existing posts. I want to be able to add data to the distribution on the fly, only when the post is distributed.

Another way to look at it is as this hook being the counterpart of `dt_push_post_args` and `dt_subscription_post_args`. Those hooks can be used to add data to pushed posts. I want to achieve the same thing, but for pulled posts.

leogermani:
Sharing some additional context:

Before 2.0, I was using [core's `rest_prepare_post`](https://github.com/Automattic/newspack-network/blob/9e370127caef28382ef9811ea5b26fe241719037/includes/distributor-customizations/class-author-distribution.php#L30) and it took us way too long to realize that this integration broke in 2.0, since Distributor no longer pull posts via the core REST API endpoints, but uses the same endpoint to list posts to be pulled and to actually pull them.

dkotter (approved):

dkotter:
I'm fine with including this new filter.

As @peterwilsoncc mentions, there may be a better way for us to handle this in Distributor that doesn't require others to utilize a filter. But for now, I think adding this is fine.

PR #1184, opened 2024-01-18
Fix typos

szepeviktor:
### Description of the Change

Found few misspellings.

### How to test the Change

Read the diff.

### Checklist:


.github/workflows/dependency-review.yml
```diff
@@ -1,6 +1,6 @@
 # Dependency Review Action
 #
-# This Action will scan dependency manifest files that change as part of a Pull Reqest, surfacing known-vulnerable versions of the packages declared or updated in the PR. Once installed, if the workflow run is marked as required, PRs introducing known-vulnerable packages will be blocked from merging.
+# This Action will scan dependency manifest files that change as part of a Pull Request, surfacing known-vulnerable versions of the packages declared or updated in the PR. Once installed, if the workflow run is marked as required, PRs introducing known-vulnerable packages will be blocked from merging.
```
jeffpaul, line 3: This probably exists across all our repos 😬

szepeviktor: Oh no!

# 😲

szepeviktor: Introduce `typos` to your organization.

jeffpaul:
@10up/open-source-practice tagging the team as there are some code-related typos that I'm not quickly certain aren't referenced elsewhere / might have implications in changing

includes/debug-info.php
```diff
@@ -146,7 +146,7 @@ function add_debug_info( $info ) {
  *
  * @return array
  */
-function get_formatted_internal_connnections() {
+function get_formatted_internal_connections() {
```
peterwilsoncc, line 159: In case an external plugin is using these functions, it might be worth retaining the old names as wrappers for the new names and marking them as deprecated.

```php
get_formatted_internal_connnections() {
    ___deprecated_function( __FUNCTION__, 'x.x.x', __NAMESPACE__ . '\\get_formatted_internal_connections' );
    return get_formatted_internal_connections();
}
```

The same applies for the external getter below.

includes/settings.php
```diff
@@ -195,7 +195,7 @@ function setup_fields_sections() {
 	add_settings_field( 'media_handling', esc_html__( 'Media Handling', 'distributor' ), __NAMESPACE__ . '\media_handling_callback', 'distributor', 'dt-section-1' );
 
 	if ( false === DT_IS_NETWORK ) {
-		add_settings_field( 'registation_key', esc_html__( 'Registration Key', 'distributor' ), __NAMESPACE__ . '\license_key_callback', 'distributor', 'dt-section-1' );
+		add_settings_field( 'registration_key', esc_html__( 'Registration Key', 'distributor' ), __NAMESPACE__ . '\license_key_callback', 'distributor', 'dt-section-1' );
```
peterwilsoncc, line 198: I've confirmed this doesn't affect the setting itself as it uses a different string.

(Mainly a note to myself and future reviewers so we don't need to check it each time.)

peterwilsoncc (commented):
The code changes look fine to me, I've confirmed the registration key is retained with the settings change.

The one thought I have is to properly deprecate the function names as they're available externally so it's possible they are used elsewhere. I couldn't find anything on GitHub but as Distributor tends to be used by enterprise clients, often their repositories are private.

includes/debug-info.php
szepeviktor: Users of this plugin cannot interact with these two functions.
- not possible to override
- not possible to filter

peterwilsoncc: I'm thinking of the case in which a plugin extending the functionality of this plugin calls these functions. 

As a simple example, a plugin that adds a WP CLI command to get the external connections:

```php
if ( ! class_exists( 'WP_CLI' ) ) {
	return;
}

/**
 * WP CLI Command to list external connections.
 */
if ( ! class_exists( 'WP_CLI' ) ) {
	return;
}

/**
 * WP CLI Command to list external connections.
 */
function external_connections() {
	// Typo in Distributor function name.
	$connections = Distributor\DebugInfo\get_formatted_external_connnections();
	$format_connections = [];
	foreach ( $connections as $name => $connection ) {
		$data = explode( "\n", $connection );

		$format_connection['Name'] = $name;

		foreach ( $data as $cell ) {
			$cell = explode( ': ', $cell, 2 );
			$format_connection[substr( $cell[0], 2 )] = $cell[1];
		}

		$format_connections[] = $format_connection;
	}

	WP_CLI\Utils\format_items( 'table', $format_connections, [ 'Name', 'URL', 'Version', 'Roles Allowed to Push' ] );
}
WP_CLI::add_command( 'dist', 'external_connections' );
```

szepeviktor: I see!

szepeviktor: Added.

peterwilsoncc (approved):
Thanks @szepeviktor, this looks good to me.

I've taken the liberty of pushing a few coding standards fixes to your branch in d98248e04424da7553a4fe562478f279cd32f561 and fixed a few more typos in 95d8118068cf1211726290da2edda6801d3f8480.

This looks good to me.

szepeviktor:
Glad to contribute.

PR #1185, opened 2024-01-18
Fix post cache

leogermani:
### Description of the Change

Fixes a typo in `DistributorPost::parse_media_blocks` that was not correctly parsing the post ID into the cache key.

This was causing the cache for all posts to be stored into the same key and override the cache from a previous cached post.

In environments that use persistent cache, this would cause distributed images to send the wrong images out sites.

### How to test the Change

This will most likely only affect environments using persistent cache. But by adding some debugging before and after this change you can confirm that the cache is not being saved with the correct key, in which the post ID is parsed.

### Changelog Entry

> Fixed - Caching Post images

### Credits

Props @leogermani

### Checklist:

### DIFF
--- a/includes/classes/DistributorPost.php
+++ b/includes/classes/DistributorPost.php
@@ -666,7 +666,9 @@ class DistributorPost {
 		$found = false;
 
 		// Note: changes to the cache key or group should be reflected in `includes/settings.php`
-		$media = wp_cache_get( 'dt_media::{$post_id}', 'dt::post', false, $found );
+		$cache_key   = "dt_media::{$this->post->ID}";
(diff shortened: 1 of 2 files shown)


peterwilsoncc (approved):
Thanks for catching this @leogermani.

This looks good to me and tests well.

leogermani:
> Thanks for catching this @leogermani.
> 
> This looks good to me and tests well.

@peterwilsoncc Thank you.

This creates a huge mess when using persistent cache in the network of sites. We are being able to workaround it by clearing this cache entry on every request, but it would be great to have this released as soon as you think it's convenient.

Thanks again.

PR #1237, opened 2024-06-23
Add categories to pull screen - 1

kirtangajjar:
### Description of the Change

Add categories to the pull screen. This is a part 1 of 2  of this issue. This PR only lists categories from posts. 

Closes #428

### How to test the Change

1. Open a pull screen and select any site
2. Ensure that you see "Categories" column on it.
3. Ensure that the categories link to their original site.
4. Ensure step 1-3 for both internal and external connections.

### Credits

Props @kirtangajjar @sanketio 

### Checklist:
- [x] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my change.
- [x] All new and existing tests pass.

PR #1242, opened 2024-07-05
Fix cypress tests

### LINKED ISSUE #1241: Fix E2E tests
### Describe the bug

E2E tests are failing on the trunk. This is likely happening because of an old version of wp-env. The same issue has been present in other libraries as well i.e. [PMK#186](https://github.com/10up/publisher-media-kit/pull/186/files#diff-7e81eeb0896ab9399d73b90a3d7a92724f5db73930baa17cb68489a82061e3f9R15) and https://github.com/10up/eight-day-week/pull/150. We tried to fix tests in https://github.com/10up/distributor/pull/1237#issuecomment-2209395488, but it seems that this needed a separate issue.

### Steps to Reproduce

1. Create any MR
2. Watch E2E tests fail on it.

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

kirtangajjar:
### Description of the Change

Fixes E2E tests

Closes #1241

### Changelog Entry

> Fixed - E2E tests

### Credits

Props @kirtangajjar

### Checklist:

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -93,8 +93,8 @@ jobs:
 
     - name: Log WP environment versions
       run: |
-        npx wp-env run cli wp core version
-        npx wp-env run cli php --version
+        npm run env run cli -- wp core version
+        npm run env run cli -- php --version
 
     - name: Test
       run: npm run cypress:run
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -93,8 +93,8 @@ jobs:
 
     - name: Log WP environment versions
       run: |
-        npx wp-env run cli "wp core version"
-        npx wp-env run cli "php --version"
+        npx wp-env run cli wp core version
+        npx wp-env run cli php --version
 
     - name: Test
       run: npm run cypress:run
--- a/tests/bin/initialize.sh
+++ b/tests/bin/initialize.sh
@@ -1,21 +1,21 @@
 #!/bin/bash
 set -e
 
-npm run env run tests-wordpress "chmod -c ugo+w /var/www/html"
-npm run env run tests-cli "wp rewrite structure '/%postname%/' --hard"
+npx wp-env run tests-wordpress chmod -c ugo+w /var/www/html
+npx wp-env run tests-cli wp rewrite structure '/%postname%/' --hard
 
 status=0
-npm run env run tests-cli "wp site list" || status=$?
+npx wp-env run tests-cli wp site list || status=$?
 
 if [ $status -eq 0 ]
 then
 	echo "Multisite already initialized"
 else
 	echo "Converting to multisite"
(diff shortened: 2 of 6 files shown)


.github/workflows/cypress.yml
```diff
@@ -93,8 +93,8 @@ jobs:
 
     - name: Log WP environment versions
       run: |
-        npx wp-env run cli "wp core version"
```
kirtangajjar, line 96: We needed to remove quotes because, with quotes, arguments were not being passed properly. The container was looking for a "wp core version" binary.

tests/bin/initialize.sh
```diff
@@ -1,21 +1,21 @@
 #!/bin/bash
 set -e
 
-npm run env run tests-wordpress "chmod -c ugo+w /var/www/html"
-npm run env run tests-cli "wp rewrite structure '/%postname%/' --hard"
+npx wp-env run tests-wordpress chmod -c ugo+w /var/www/html
```
kirtangajjar: We needed to change from `npm run env run` to `npx wp-env run` as after removing the quotes, flags like `-c ugo+w` and `--slug=second` were being omitted while passing the command to wp-env. My best guess is that those flags were passed to npm instead.

kirtangajjar:
The pending work here is described in https://github.com/10up/distributor/pull/1237#issuecomment-2209395488. Basically as we have removed quotes around command here, we need to do the same in  https://github.com/10up/cypress-wp-utils/blob/develop/src/commands/wp-cli.ts#L21 unless we find a different solution.

tests/bin/initialize.sh
iamdharmesh: @kirtangajjar As this shell script is triggered by the npm after script, I think we can directly use `wp-env` and avoid using `npx`.

iamdharmesh:
@kirtangajjar PR looks good to me just added one note to discuss it. 

> we need to do the same in https://github.com/10up/cypress-wp-utils/blob/develop/src/commands/wp-cli.ts#L21 unless we find a different solution.

I have already raised PR https://github.com/10up/cypress-wp-utils/pull/128 to handle this in https://github.com/10up/cypress-wp-utils/

Just curious on why E2E test GH action check is not triggered on this PR 🤔 

Thank you.

jeffpaul:
@iamdharmesh looks like e2e tests are running with latest commit 🤷🏼

kirtangajjar:
@jeffpaul I think the code is good to merge. I'm observing some tests are failing for `Should push oEmbed content` on some envs, which I think is because the tests might be flakey. I cannot think of other reasons why they're failing. 

cc @iamdharmesh any thoughts on why the tests could be failing otherwise?

iamdharmesh (approved):
Yes @kirtangajjar, It seems like flaky tests. I tried re-run on my commit and now it is passing. Approving. Thanks for all your great work here.

PR #1244, opened 2024-07-09
Add categories to pull screen - 2

### LINKED ISSUE #428: add Category to bulk pull screen
I received the following feedback in one of my email conversations with someone who's tested Distributor.

> I have been doing some testing here and see that I can ‘bulk’ pull posts which is great. What would really be brilliant and make the plugin very useful is if it was possible to view the category of the posts and select which categories to pull. As currently if you have thousands of posts without seeing the category they are in it makes it difficult to pull what you need.

**Designs**
n/a

none.

**Additional context**
Related to #392 (in that it impacts the same screen).

kirtangajjar:
### Description of the Change

Add categories to the pull screen. This is part 2 of 2  of this issue. This PR filters posts by categories. 

Closes #428

### How to test the Change

1. Open post pull screen
2. You should be able to see post categories in the filter.
3. Select a category and filter. The posts should be filtered by categories.
4. Pull the posts and verify only posts from that category have been pulled.
5. Verify it on both internal and external connections.
6. Verify it in conjunction with post type filter as well.

### Changelog Entry

> Added - Category filter in pull screen

### Credits

Props @kirtangajjar 

### Checklist:


faisal-alvi (changes requested):
@kirtangajjar thanks for working on this. I tested and found the category filter does not work. Please check the GIF:



When I select `test03`, 0 posts should return as results, but it still shows the post with the `Uncategorised` category.

sanketio:
@faisal-alvi The issue mentioned in https://github.com/10up/distributor/pull/1244#pullrequestreview-2175094375, should be fixed now. In addition to that, I have removed outputting the filters from the bottom, as it is being rendered with the same HTML id attribute, which is an issue.

One more thing we need to work on, is we will have to update the taxonomy list based on the supported post-types, I think, instead of just outputting all the taxonomies for all the post-types. I will work on it in this PR.

cc: @jeffpaul

sanketio:
@jeffpaul I think we should also not generalize this only for the category, but give an option (maybe a filter) to allow other taxonomies as well, just a thought! What you say?

sanketio:
For example, if the `page` post-type does not supports category or any other taxonomies, we shouldn't display the filter. So I think that category (taxonomy in general) filter should be based on post-type supports, instead of just the category or taxonomy.

In addition to that, it should respect the remote site's post-type and not the post-type on the current site.

sanketio:
Just adding a note about https://github.com/10up/distributor/pull/1244#issuecomment-3174971249, that it's going to be tricky because we will have to generate the taxonomy terms again and again when someone changes the post-type filter. But I think we can get all the terms when creating a post-type filter, so trying it out as of now.

jeffpaul:
> I think we should also not generalize this only for the category, but give an option (maybe a filter) to allow other taxonomies as well, just a thought! What you say?

Making this filterable so folks with custom taxonomies sounds good to me.

> In addition to that, it should respect the remote site's post-type and not the post-type on the current site.

Since you're filtering terms from a separate site, we should render those in the taxonomy dropdown.  If that term is missing on the current site then it should get added as part of the pull distribution

sanketio:
After thorough testing I found that the `pull` feature with this PR breaks, and I tried debugging it but it keeps taking more time.

I'm creating new PR to generalize the taxonomy terms filter. :)

sanketio:
@jeffpaul Just an update, I started working on generalizing taxonomy filter, and it's going pretty good. I think I will be able to complete it by tomorrow. :)

sanketio:
Created a new PR: https://github.com/10up/distributor/pull/1327 for this, I will continue working on this in that PR.

sanketio:
@jeffpaul @dkotter This can be closed, as I have merged #1327 to the main PR #1237

PR #1256, opened 2024-08-04
Pull screen search feature

### LINKED ISSUE #1235: Add ability to search for a site on the Pull Content screen

For collection of sites using Distributor that have a large number of connected sites (e.g. >1000) there's a good option for searching through that listing on the Push menu to try and find a site to push content to.  However the Pull Content menu lacks a similar search feature, so adding search there for the list of connected sites would be helpful in similar scenarios (e.g. >1000 connected sites).

### Designs

Here's an example of the Push menu with ability to search for a connected site to push content to:

push-menu-search

...and here's an example of how the Pull Content screen has no ability to search on connected sites, having something available would help for set ups that have 100s+ of connected sites to more quickly find the site to pull content from:

pull-content-no-search

n/a

### Code of Conduct

kirtangajjar:
### Description of the Change

Add ability to search for a site on the Pull Content screen

Closes #1235

### Changelog Entry

> Added - Ability to search for a site on pull content screen

### Credits

Props @kirtangajjar

### Checklist:


assets/js/admin-pull.js
```diff
@@ -98,3 +88,115 @@ const getURL = () => {
...
+	const inputContainer = container.querySelector(
+		'.searchable-select__input-container'
+	);
+	const input = container.querySelector( '.searchable-select__input' );
+	const icon = container.querySelector(
+		'.searchable-select__input-container > .dashicons-arrow-down'
+	);
+	const dropdown = container.querySelector( '.searchable-select__dropdown' );
+
+	const itemss = await fetch( '/wp-admin/admin-ajax.php', {
```
peterwilsoncc: Typeo? If it's meant to be searchResults or similar then use the longer name for clarity.

```suggestion
	const items = await fetch( '/wp-admin/admin-ajax.php', {
```

assets/js/admin-pull.js
```diff
@@ -98,3 +88,115 @@ const getURL = () => {
...
+		dropdown.style.display = 'block';
+	}
+
+	function hideDropdown() {
+		dropdown.style.display = 'none';
+	}
+
+	function filterItems( searchTerm ) {
+		const filteredItems = itemss.filter( ( item ) =>
+			item.toLowerCase().includes( searchTerm.toLowerCase() )
```
peterwilsoncc: I'm getting an error on this line as `item` is an object:

```
{
  "id": "2",
  "name": "ms-distributor.local/two",
  "url": "ms-distributor.local/two",
  "pull_url": "http://ms-distributor.local/wp-admin/admin.php?page=pull&#038;connection_type=internal&#038;connection_id=2",
  "type": "internal"
}
```

includes/pull-ui.php
```diff
@@ -624,3 +588,63 @@ function output_pull_errors() {
...
+	if ( ! empty( \Distributor\Connections::factory()->get_registered()['networkblog'] ) ) {
+		$sites = \Distributor\InternalConnections\NetworkSiteConnection::get_available_authorized_sites( 'pull' );
+
+		foreach ( $sites as $site_array ) {
+			$internal_connection = new \Distributor\InternalConnections\NetworkSiteConnection( $site_array['site'] );
+
+			$connections[] = [
+				'id'   => $internal_connection->site->blog_id,
+				'name' => untrailingslashit( $internal_connection->site->domain . $internal_connection->site->path ),
+				'url'  => untrailingslashit( preg_replace( '#(https?:\/\/|www\.)#i', '', get_site_url( $internal_connection->site->blog_id ) ) ),
```
peterwilsoncc: I'm a little unclear why name and URL are using the same (or at least similar ) values. `site->domain . site->path` ought to resolved to `get_site_url()`.

Although we don't display the name, it may be helpful to allow people to search by the name, so a search for `Make WordPress Core` would return `make.wordpress.org/core`

includes/push-ui.php
```diff
@@ -23,7 +23,7 @@ function() {
 			add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_scripts' );
 			add_filter( 'amp_dev_mode_element_xpaths', __NAMESPACE__ . '\add_element_xpaths' );
 			add_filter( 'script_loader_tag', __NAMESPACE__ . '\add_dev_mode_to_assets', 10, 2 );
-			add_action( 'wp_ajax_dt_load_connections', __NAMESPACE__ . '\get_connections' );
+			add_action( 'wp_ajax_dt_load_connections_push', __NAMESPACE__ . '\get_connections' );
```
peterwilsoncc: I'm assuming this is for clarity from `wp_ajax_dt_load_connections_pull`? 

We're probably stuck with the old name to ensure backward compatibility with the third party extensions and between different versions of the plugin. Otherwise we could go through a soft deprecation.

peterwilsoncc (changes requested):
I've added some notes inline.

The UI looks good but I'm hitting a bug with the search. Otherwise the notes a minor.

assets/js/admin-pull.js
kirtangajjar: Improved the name of var

assets/js/admin-pull.js
kirtangajjar: Yupp, fixed it now.

includes/pull-ui.php
kirtangajjar: Good idea, added name and URL as separate things and showed them both on FE as well.

includes/push-ui.php
kirtangajjar: Thanks for pointing it out. I've added the old method back with soft deprecation. LMK if I did it correctly(this is my first time doing it)

kirtangajjar:
@peterwilsoncc I've addressed your feedback. LMK if you find anything else.

peterwilsoncc:
These changes look good, thank you.

This is looking good, I'm seeing a bug on the pull screen when there are two connections with the same ID (ie an external connection's POST ID matches the network connection's SITE ID). I'm seeing that selecting the external site is showing the internal site in the drop down following a refresh:

pulled-from-incorrect

In the gif you'll see I am selecting `XU-Distributor` but upon refresh the dropdown is showing `10 Dist thirty-four`.

PR #1296, opened 2025-01-24
Check for an existing image before processing media on the target site.

### LINKED ISSUE #765: Media files duplicated after distribution
Distributing multiple posts that use the same image file as post thumbnail, the media file gets duplicated into the receiving site's media library.

**Steps to Reproduce**
1. Assign the same image file to multiple posts in origin site
2. Distribute these posts to a subsite
3. Check subsite's media library
4. See myimage.jpg, myimage-1.jpg, myimage-2.jpg, myimage-3.jpg etc.

**Expected behavior**
Distributor should check if image file is already present in target site, and then (re)use this image.

WordPress 5.7.2
Distributor 1.6.4-dev

iamdharmesh:
### Description of the Change
As reported in #765, when the same image is used in multiple posts and these posts are distributed to connected sites, the media files are imported for each post, creating duplicate copies of the images on the target sites. This PR fixes the media duplication issue by checking for existing media on the target site before creating new media.

Closes #765

### How to test the Change
1. Assign the same image file to multiple posts on the origin site.
2. Distribute these posts to a subsite (network connection) or external connection.
3. Verify that there are no duplicate images on the target site and that images are reused in multiple posts.

### Credits
Props @philbraun @jeffpaul @iamdharmesh 

### Checklist:

### DIFF
--- a/tests/php/UtilsTest.php
+++ b/tests/php/UtilsTest.php
@@ -777,12 +777,11 @@ class UtilsTest extends TestCase {
 	 * @since 1.0
 	 * @group Utils
 	 * @runInSeparateProcess
+	 * @dataProvider data_test_set_media
 	 */
-	public function test_set_media() {
+	public function test_set_media( $new_image_id, $source_url, $existing_media_id ) {
 		$post_id    = 1;
 		$media_item = $this->test_format_media_featured();
-
-		$new_image_id = 5;
 
 		$attached_media_post                 = new \stdClass();
 		$attached_media_post->ID             = 3;
@@ -827,17 +826,10 @@ class UtilsTest extends TestCase {
 		);
 
 		\WP_Mock::userFunction(
-			'wp_delete_attachment', [
-				'times' => 1,
-				'args'  => [ $attached_media_post->ID, true ],
-			]
-		);
-
-		\WP_Mock::userFunction(
 			'get_post_meta', [
 				'times'  => 1,
 				'args'   => [ $attached_media_post->ID, 'dt_original_media_url', true ],
-				'return' => 'http://mediaitem.com',
+				'return' => $source_url,
 			]
 		);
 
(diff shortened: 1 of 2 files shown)


tests/php/UtilsTest.php
```diff
@@ -777,13 +777,12 @@ public function test_format_media_no_attachment_meta() {
 	 * @since 1.0
 	 * @group Utils
 	 * @runInSeparateProcess
+	 * @dataProvider data_test_set_media
 	 */
-	public function test_set_media() {
+	public function test_set_media( $new_image_id, $source_url, $existing_media_id ) {
```
peterwilsoncc, line 786: Add `@param` attributes to docblock.

tests/php/UtilsTest.php
```diff
@@ -938,6 +936,20 @@ public function test_set_media() {
 		Utils\set_media( $post_id, [ $media_item ], [ 'use_filesystem' => false ] );
 	}
 
+	/**
+	 * Data provider for test_set_media
+	 *
+	 * @since 1.0
+	 * @group Utils
+	 */
```
peterwilsoncc, line 949:
```suggestion
	 *
	 * @return array[] Data provider.
	 */
```

tests/php/UtilsTest.php
```diff
@@ -938,6 +936,20 @@ public function test_set_media() {
...
+	public function data_test_set_media() {
```
peterwilsoncc: Nit: I recommend removing the `test_`, that way we have a test prefix for tests and a data prefix for data providers.

I also don't have super strong feeling so ignore if you have your own preferences. This is the first data provider so we can use this to set a convention.

peterwilsoncc (commented):
This looks good to me and tests well, thanks.

Just a minor note inline to include the data provider param annotations in the test's docblock. Otherwise I think it's good to go.

tests/php/UtilsTest.php
iamdharmesh: Thanks for the note.

tests/php/UtilsTest.php
iamdharmesh: Updated in https://github.com/10up/distributor/pull/1296/commits/264bd4b0e881722cb941fe39c30a709025d07c78. Forgot to remove `test_` after the copy function name. Thanks for the note.

peterwilsoncc (approved):
Thanks for the follow up, this looks good to merge.

PR #1302, opened 2025-03-03
Stored ID handling - Phase 1

iamdharmesh:
### Description of the Change
This PR introduces a new helper function, `distributor_register_data`, which enables developers to register stored data references, such as IDs stored in post meta, shortcodes, or block attributes and define custom pre-distribution and post-distribution callbacks to process these references.

#### Function Definition

**`distributor_register_data`**  
This function registers a piece of data by providing details about its location and the callbacks needed to process it during distribution.

**Parameters**

- **`$data_name`** (`string`): The unique identifier for the data.
- **`$args`** (`array`): An array of parameters that describe:
  - **`location`** (`string`): Where the data is stored. Accepted values are `'post_meta'` or `'post_content'`.
  - **`attributes`** (`array`): Additional details depending on the location:
    - **For post meta:**
      - `meta_key` (`string`): The post meta key. **(Required)**
    - **For shortcodes:**
      - `shortcode` (`string`): The shortcode tag. **(Required)**
      - `shortcode_attribute` (`string|array`): The attribute(s) containing the data. **(Required)**
    - **For blocks:**
      - `block_name` (`string`): The block name. **(Required)**
      - `block_attribute` (`string|array`): The attribute(s) containing the data. **(Required)**
  - **`pre_distribute_cb`** (`callable`): A callback function that prepares the data on the source site before distribution. It should return extra data required for processing on the target site.
  - **`post_distribute_cb`** (`callable`): A callback function that processes the extra data on the target site and returns the updated ID to replace the source reference.
  
I have added detailed information on how this function works along with examples. You can view them [here](https://github.com/10up/distributor/blob/enhancement/stored-id-handling/docs/stored-id-handling.md).

> [!NOTE]
> - For the Classic editor, the post content is distributed in its rendered form. Therefore, shortcode data does not work as expected because the output of the shortcode is distributed instead of the shortcode tag. Let's discuss how we can handle this. I need feedback on this from someone who has a better understanding of this.
> - This PR is Phase 1 of stored ID handling, which aims to add support for popular plugins like ACF and Gravity Forms in upcoming phases. However, to fully support these plugins, additional implementation will be required beyond this newly added helper function.

### How to test the Change
[...]

PR #1305, opened 2025-03-18
Update all actions we use in our workflows to pull from specific pinned commits

dkotter:
### Description of the Change

In order to help protect against compromised actions, instead of including actions based on their major version (like v4), this PR switches all the actions we use to pull based on the commit hash from the latest release.

While this does impact maintenance a bit going forward, it ensures that we we're always using actions that we (hopefully) trust and if an action gets compromised (which [happens](https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/)) we don't have to worry that we're using a compromised action. This also goes along with what [GitHub suggests](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

### How to test the Change

Ensure all our workflows still run as expected

### Changelog Entry

> Developer - Update all third-party actions our workflows rely on to use versions based on specific commit hashes.

### Credits

Props @dkotter, @jeffpaul 

### Checklist:

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,31 +18,38 @@ jobs:
     name: eslint
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
     - name: Use desired version of NodeJS
-      uses: actions/setup-node@v4
+      uses: actions/setup-node@cdca7365b2dadb8aad0a33bc7601856ffabcc48e # v4.3.0
       with:
           node-version-file: '.nvmrc'
+
     - name: Cache Node
-      uses: actions/cache@v4
+      uses: actions/cache@d4323d4df104b026a6aa633fdb11d772146be0bf # v4.2.2
       with:
         path: |
           node_modules
           ~/.cache
           ~/.npm
         key: ${{ hashFiles('package-lock.json') }}
+
     - name: Check versions
       run: npm -v; node -v
+
     - name: npm install
       run: npm install
+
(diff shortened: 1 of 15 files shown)


.github/workflows/lint.yml
```diff
@@ -18,31 +18,38 @@ jobs:
...
     - name: npm install
       run: npm install
+
     - name: Generate linting report
       run: npm run lint:js -- --output-file eslint-report.json --format json
       continue-on-error: true
+
     - name: Annotate code linting results
-      uses: ataylorme/eslint-annotate-action@1.2.0
+      uses: ataylorme/eslint-annotate-action@5f4dc2e3af8d3c21b727edb597e5503510b1dc9c # v2.2.0
```
jeffpaul, line 48: Looks like some interesting updates in the 3.0.0 release, but definitely a concern for a different day/PR

jeffpaul (dismissed):
Looks good, thanks for handling these so quickly!

.github/workflows/lint.yml
dkotter: Yeah... I debated going to 3.0.0 but with the noted breaking changes there, decided it might be better to handle that separately (though it may just work without any needed changes on our end)

dkotter:
Some E2E failures here but does not seem related to the changes in this PR. I think I may wait until #1302 gets merged in (as that PR fixes some existing failures) to see if that fixes things here

faisal-alvi (approved):

dkotter:
Decided to go ahead and merge this in instead of waiting on #1302 so we can have these changes in place if anyone asks about it. Nothing here should impact if E2E tests pass or not so no real need to wait

PR #1316, opened 2025-05-21
Potential fix for code scanning alert no. 3: DOM text reinterpreted as HTML

Sidsector9:
Potential fix for [https://github.com/10up/distributor/security/code-scanning/3](https://github.com/10up/distributor/security/code-scanning/3)

To fix this issue, we need to ensure that the user input from `searchField.value` is properly sanitized or encoded before being used in the URL. The best way to do this is to use a library that provides functions for escaping or encoding user input to prevent XSS attacks.

In this case, we can use the `encodeURIComponent` function, which is a built-in JavaScript function that encodes a URI component by replacing each instance of certain characters by one, two, three, or four escape sequences representing the UTF-8 encoding of the character.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/assets/js/admin-pull.js
+++ b/assets/js/admin-pull.js
@@ -39,7 +39,7 @@ if ( chooseConnection && choosePostType && form ) {
 		jQuery( searchBtn ).on( 'click', ( event ) => {
 			event.preventDefault();
 
-			const search = searchField.value;
+			const search = encodeURIComponent(searchField.value);
 
 			document.location = `${ getURL() }&s=${ search }`;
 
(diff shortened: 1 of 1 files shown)


assets/js/admin-pull.js
```diff
@@ -39,7 +39,7 @@ if ( chooseConnection && choosePostType && form ) {
 		jQuery( searchBtn ).on( 'click', ( event ) => {
 			event.preventDefault();
 
-			const search = searchField.value;
+			const search = encodeURIComponent(searchField.value);
```
dkotter:
```suggestion
			const search = encodeURIComponent( searchField.value );
```

dkotter (approved):

Issue #1317, opened 2025-05-29
Implementation of automatic push of a post


maxim-wla:
### Describe your question

I am trying to implement automatic push of a post from one site to another when publishing it, I use external connection from the Distributor plugin. I get a connection, then all the sites that are connected and I try to push a post to a specific site, but I can't do it. Can you tell me if this is really possible to implement and if so, how can I do it? Thanks

```
add_action( 'save_post_products', 'auto_distribute_product_on_save', 99, 3 );
public function auto_distribute_product_on_save( $post_id, $post, $update ) {
		if ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) return;
		if ( wp_is_post_revision( $post_id ) ) return;
		if ( ! class_exists( '\Distributor\Connections' ) ) return;

		$factory = Connections::factory();
		$registered = $factory->get_registered( 'external' );
		if(!empty($registered['networkblog'])){
			$blogs = get_sites();
			$current_site = get_current_site();
			foreach ($blogs as $blog){
				if($blog->id !== $current_site->id){
					$network_connection = new $registered['networkblog']($blog);
					if(method_exists($network_connection, 'push')){
						$res = $network_connection->push($post);
					}
				}
			}
		}
	}
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1321, opened 2025-06-16
Auto Distribution feature.

### LINKED ISSUE #456: Add ability to auto-distribute to all connected sites (internal and external)
**Describe your question**

I have a large multisite and would like to distribute a post to all my subsites, Is there a way to do this without having to click select 350 times? I can add filters or hooks as needed, I wanted to ask before our dev team got into trying to "make" it work.

Thanks
ML

peterwilsoncc:
### Description of the Change

Allows developers to enable auto-distribution of posts via the use of a filter.

Todo:

- [x] Limit allowed post types
- [x] Limit allowed connections by type
- [x] Limit allowed connections by type + ID
- [x] Filter to modify default status to `draft`
- [ ] Maybe hide push/pull distribution UI when enabled

Closes #456

### How to test the Change

1. Create a multisite network
2. Add several sub-sites
3. Create a single site
4. Connect the single site as an external connection to the main site of the MS network. If using a local environment, ensure that the two can communicate if using a virtual machine
5. Enable the feature by adding the following as an MU-plugin `add_filter( 'dt_auto_distribution_enabled', '__return_true' );`
6. Publish a post on the main site of the MS network
7. A few moments after pressing publish, ensure the post has been distributed to all connections. You can do this via the distributor drop-down in the admin bar if you wait
8. Ensure the site is published on the destination sites.
9. Add the following to your test plugin `add_filter( 'dt_auto_distribution_default_status', function() { return 'draft'; } );`
10. Publish a post on the main site of the MS network
11. A few moments after pressing publish, ensure the post has been distributed to all connections
12. Ensure the post is a draft on each of the connections
13. Remove the draft filter created above from your test plugin
14. Add the following to the test plugin:
   ```
   add_filter( 'dt_auto_distribute_post', function() {
	// Randomise the result to true or false for testing purposes.
	return (bool) rand( 0, 1 );
   }, 10, 3 );
   ```
15. Publish a post on the main site of the MS network
16. A few moments after pressing publish, ensure the post has been distributed to approximately half of the connections. _Note: This test will work better with more sites, as it reduces the chance of dumb luck distributing to all of the sites/none of the sites_
17. Remove the test plugin from your site to avoid confusion at a later date.

### Changelog Entry

> Added - Auto-distribution of posts via a filter

### Credits

Props @peterwilsoncc

### Checklist:


includes/auto-distribute.php
```diff
@@ -0,0 +1,440 @@
...
+ */
+function get_external_connections( $post_id = 0, $user_id = 0 ) {
+	$connection_objects = [];
+	$connections        = new \WP_Query(
+		[
+			'post_type'      => 'dt_ext_connection',
+			'fields'         => 'ids',
+			'no_found_rows'  => true,
+			// phpcs:ignore WordPress.WP.PostsPerPage.posts_per_page_posts_per_page -- All connections need to be fetched.
+			'posts_per_page' => 550,
```
faisal-alvi: Just a quick check – hoping this doesn’t introduce any performance overhead with a large number of connections. Also, curious if there’s a specific reason for hardcoding 550 here?

I noticed we’ve used 200 as the limit in the other file on the following line, so wondering if it might be worth aligning both for consistency?

https://github.com/10up/distributor/blob/ac1204b85ed103c3958ab410886bc1a9ab4020f3/includes/push-ui.php#L131

Also, would it make sense to wrap this in a filter for more flexibility going forward?

includes/auto-distribute.php
```diff
@@ -0,0 +1,440 @@
...
+}
+
+/**
+ * Get all connections
+ *
+ * @param integer $post_id Post ID.
+ * @param integer $user_id User ID.
+ * @return array
+ */
+function get_connections( $post_id = 0, $user_id = 0 ) {
```
faisal-alvi, line 294: Just wondering – do we already have some of these common functions in a helper class or utility file that we could leverage here? That might help keep things consistent and reusable.

Same question for the other functions added below, like `get_external_connections` and `get_internal_connections`.

faisal-alvi (changes requested):
@peterwilsoncc Thanks so much for the work on this!

I tested the feature locally and noticed that two copies are created when a post is auto-distributed. I assume only one copy should be created. Could you please take a look?

image

I’ve also shared a couple of additional points below for your consideration.

Lastly, it might be helpful to add some test coverage for this feature to ensure long-term stability.

includes/auto-distribute.php
peterwilsoncc: We have similar code in `Distributor\PushUI\get_connections()` that is behind a nonce and capability check, so not appropriate in this instance as it needs to run on cron.

Do you want me to see if I can share the functionality between this and the admin-ajax endpoint and move this to the utils namespace?

includes/auto-distribute.php
peterwilsoncc: Honestly, it was a copy-paste from the existing auto-distribute plugin.

Looks like we use different numbers in a few different places, I'll look at unifying and setting up a filter for getting external and internal sites.

peterwilsoncc:
@faisal-alvi I'm not seeing the double posting. Is there any chance you have a site on a multisite instance set up as both an internal and external connection? 

I'll try to figure out how to test this as the unit tests use mocks. I'll check the reliability of wp-cron in wp-env.

includes/auto-distribute.php
faisal-alvi: Yeah, if you believe separating those common functions and moving them to the utils namespace would be beneficial for future enhancements (and it does not require too much efforts), then we should go ahead with that. Otherwise, I’m also fine keeping it as is for now; happy to go with your judgment on this.

faisal-alvi:
@peterwilsoncc Regarding the double posting: I have a multisite setup (subdirectory structure) with one main site and one subsite. I had set up an external connection from the subsite by adding the main site URL; that was the only external connection configured.

However, it seems that in a multisite network, the sites are already internally connected by default, and we might not need to set up an external connection manually. Is that correct?

When I published a post from the subsite, two copies were auto-distributed to the main site. But when I removed the external connection and tested again, only one copy was distributed; so it looks like the feature works as expected, and the duplicate was likely due to the external + internal connection both being active.

If my assumption is correct that sites in a multisite are automatically connected internally, could you clarify how we can limit distribution to specific subsites, rather than distributing to all?

As for the tests, I’d lean towards adding an E2E test rather than a PHP unit test. If you think it’s feasible and worth the effort, then I think adding an E2E test would be a solid approach.

dkotter (commented):
Overall this looks good to me, noting Faisal has a number of comments he's already added. One thing that may be nice to add is a filter around the connections, making it easier to remove some of those from auto-distribution. Right now you can filter the posts that get auto-distributed and the post types that get auto-distributed but not seeing an easy way to filter the connections that get auto-distributed to

peterwilsoncc:
> However, it seems that in a multisite network, the sites are already internally connected by default, and we might not need to set up an external connection manually. Is that correct?

@faisal-alvi Yes, that's correct. Without auto-distribution it's fine to do so but with auto-distribution it will cause the double posting you're seeing. I'll have to figure out how to handle this in the tests.

> Overall this looks good to me, noting Faisal has a number of comments he's already added. One thing that may be nice to add is a filter around the connections, making it easier to remove some of those from auto-distribution. Right now you can filter the posts that get auto-distributed and the post types that get auto-distributed but not seeing an easy way to filter the connections that get auto-distributed to

@dkotter I was thinking a developer might be able to use the same filter for the purpose:

```php
add_filter( 'dt_auto_distribute_post', 'no_external', 10, 4 );

function no_external( $should_distribute, $post, $user_id, $connection_type ) {
	if ( $connection_type === 'external' ) {
		return false;
	}
	return $should_distribute;
}
```

Is that adequate or would you prefer another filter so we can bypass the relevant code in `get_connections()`?

dkotter:
> I was thinking a developer might be able to use the same filter for the purpose:

Yeah, I think that works fine. Probably worth adding this to our documentation snippets

jeffpaul:
@peterwilsoncc thoughts on looking to cover #1317 here as well?

includes/auto-distribute.php
peterwilsoncc: IN 0a63d09a264467dcee919b6b9e459ef1095f399f I've set the number to a consistent `200` throughout. That was the highest number used (for the push screen). The value can be changed via the new filter: `dt_external_connections_per_page`.

faisal-alvi (dismissed):

dkotter:
@peterwilsoncc Is there anything else you were wanting to change/add here or is this good for merge?

peterwilsoncc:
> Is there anything else you were wanting to change/add here or is this good for merge?

I pushed 2dbe340f7d495a7e908325f67f691a175e16f91e as a one final thing: it modifies when the auto-distribution setup runs to wait until `plugins_loaded, 20` to give a little more room for feature plugins to run.

Otherwise, I'm happy if you are.

PR #1337, opened 2025-09-12
Replace hookdoc with WP Hooks Documentor

### LINKED ISSUE #1336: Replace hookdoc with WP Hooks Documentor
Replace hookdoc with WP Hooks Documentor to improve the overall dev docs. Follow what we did on ClassifAI: https://github.com/10up/classifai/pull/966

iamdharmesh:
### Description of the Change
PR replace the existing hookdoc package with the WP Hooks Documentor package which is used for generate developer docs. The new package is uses the docusaurus and provide good flexibility on extending docs site.

screencapture-localhost-3000-distributor-2025-09-12-16_26_09

> [!NOTE]
> The Distributor's yellow color is a bit light for the light theme, so I used a slightly darker shade here. Please let me know if we should use a different color instead. Dark mode looks fine, no changes there.

Closes #1336 

### How to test the Change
1. Run `npm i && npm run build:docs`
2. Run `cd ./docs-build && npm run serve`
3. Visit http://localhost:3000/distributor to check generated doc site.

### Credits

Props @jeffpaul, @iamdharmesh 

### Checklist:


README.md
```diff
@@ -171,9 +171,38 @@ By default, drafts are the preferred status and can't be changed at the source s
...
+The Distributor documentation site is built using [WP Hooks Documentor](https://github.com/10up/wp-hooks-documentor). Follow these steps to build and run the documentation site locally:
+
+### 1. Build Documentation
+
+```bash
+# Install dependencies and build the plugin
+npm i && npm run build:docs
+```
+
+This will:
```
dkotter, line 182: May be worth mentioning somewhere that PHP 8.3+ is required (I ran into that when testing locally)

dkotter (commented):
Code looks good here but I agree on the note around the yellow color.

This is very hard to read and I expect does not pass color contrast accessibility. This color works on the [Distributor site](https://distributorplugin.com/) because we only use it when the background is black. But on a white background, you can barely read the yellow text.

There's a few approaches I could see here:

1. Only have the dark mode theme and don't allow switching to light mode. That matches how our plugin site looks so may be fine
2. Anywhere we have yellow text, ensure the background is dark. This would be similar to always forcing dark mode but would still retain some lighter elements
3. (easiest option probably) Don't use that yellow color for text if the text is on a light background. Not sure what color to replace it with though, could just always use black but not as fun visually then

Open to other suggestions / opinions on this though, cc @jeffpaul

README.md
iamdharmesh: Good catch!! Added in https://github.com/10up/distributor/pull/1337/commits/ae136b6a3d19bc0316fa0e0bb7d2805830b5c5c0

iamdharmesh:
I deployed this on my fork repo here: https://iamdharmesh.github.io/distributor/ to get feedback on color in light mode.

cc: @jeffpaul

dkotter:
@iamdharmesh One other thing we'll need to take care of here that I just realized we [missed in ClassifAI](https://github.com/10up/classifai/pull/1008/commits/4a9dbbbf14ad5e3e7772acb9b8aec5a54866a138) is to update any existing links within our code that points to the old documentation site. I know there's some in the changelog and looks like some in the documentation itself

wp-hooks-docs/src/theme/Footer/index.js
```diff
@@ -0,0 +1,162 @@
...
+				<div className="container">
+					<div className={ styles.footerContainer }>
+						<div className={ styles.footerCopyright }>
+							<div className={ styles.footerCopyrightLinks }>
+								<FooterCopyright
+									copyright={ `Finely crafted by Fueled, ©${ currentYear }` }
+								/>
+								<a href="https://github.com/10up/distributor/issues/">
+									Support
+								</a>
```
peterwilsoncc, line 65: This kind of reads as a donation request as is.

```suggestion
								<a href="https://github.com/10up/distributor/issues/">
									Issues
								</a>
```

wp-hooks-docs/src/theme/Footer/index.js
```diff
@@ -0,0 +1,162 @@
...
+								</div>
+							</div>
+						</div>
+						<div className={ styles.footerSocial }>
+							<a
+								href="https://www.facebook.com/10up.agency/?ref=distributor"
+								title="10up on Facebook"
+								aria-hidden="true"
+								target="_blank"
+								rel="noopener nofollow noreferrer"
```
peterwilsoncc: `target=_blank` is really bad for accessibility, let's give users the choice.

```suggestion
```

🔢 Applies to the other social links too.

wp-hooks-docs/src/theme/Footer/index.js
```diff
@@ -0,0 +1,162 @@
...
+										WordPress Practice
+									</a>
+								</div>
+							</div>
+						</div>
+						<div className={ styles.footerSocial }>
+							<a
+								href="https://www.facebook.com/10up.agency/?ref=distributor"
+								title="10up on Facebook"
+								aria-hidden="true"
```
peterwilsoncc: At the moment these don't appear in the accessibility tree (checked with firefox dev tools)

```suggestion
								aria-label="10up on Facebook"
```

:1234: Applies to the other social links too

peterwilsoncc (commented):
Thanks for working on this, in addition to the accessibility issues I've got a couple of notes:

* defined in lines in the source section, eg on [this page](https://iamdharmesh.github.io/distributor/hooks/Actions/dt_after_set_meta#source): 
   * would it be possible to link to the permalink rather than `trunk`, eg `https://github.com/10up/distributor/blob/[commit-hash]/includes/utils.php#L318`
   * would it be possible to highlight the start of the docblock through to the `apply_filters|do_actions`, eg `https://github.com/10up/distributor/blob/4831cdcc01ac23f32f28b5bb51c683a610d7c34c/includes/utils.php#L311-L322`

jeffpaul:
@peterwilsoncc will probably need your help finishing off the changes needed here before getting a final review from @dkotter

peterwilsoncc:
@jeffpaul @dkotter What do you think of using the default colour scheme for the initial merge and we can then tidy it up to look more like distributorplugin.com once the code goes in?

Screenshot 2026-01-22 at 14-57-17 Distributor Developer Documentation

dkotter:
> @jeffpaul @dkotter What do you think of using the default colour scheme for the initial merge and we can then tidy it up to look more like distributorplugin.com once the code goes in?

Fine by me, I think this is way better than what we have now so would be nice to get this in

peterwilsoncc:
I've pushed a couple of commits to restore the default styles and make some slight colour tweaks in order to meet WCAG AA contrast minimums.

jeffpaul (approved):

jeffpaul:
@peterwilsoncc once we land this update here it looks like we'll also want to over in https://github.com/10up/safe-redirect-manager and https://github.com/10up/simple-google-news-sitemap

peterwilsoncc (approved):
I'm happy with these changes and Jeff has approved the minor changes I made on top of the original work.

I think this is good to merge in to the main branch and then trunk so the docs are generated by the new system.

dkotter (approved):

PR #1396, opened 2026-09-09
WordPress Integration Tests

peterwilsoncc:
### Description of the Change

Adds WP Integration tests using the WP test suite rather than using WP_Mock.

This is to begin the process of using the WP test suite rather than mocks to ensure the plugin works as expected in a range of circumstances.

To begin with I have:

* Moved the debug info tests to the integration test suite
* Duplicated the utilities tests in the integration test suite

For the utilities tests, I used AI for the initial pass and worked on the results. It would be good to get a logic check on the test suite to make sure I haven't missed anything.

### How to test the Change

1. Ensure the new tests pass
2. Otherwise it's a manual review to make sure that I haven't broken anything in the translation.

### Credits

Props @peterwilsoncc 

### Checklist:


.github/workflows/php-wp-integration-reusable.yml
```diff
@@ -0,0 +1,62 @@
...
+  phpunit:
+    name: WP ${{ inputs.wp-version }} / PHP ${{ inputs.php-version }}
+    runs-on: ubuntu-latest
+    env:
+      DB_DATABASE: wordpress_test
+      DB_USER: root
+      DB_PASSWORD: root
+    steps:
+    - name: Checkout
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
dkotter: I think these versions are all out of date at this point, I'm assuming this was copied from other workflow files but may want to bump these to their latest

.github/workflows/php-wp-integration-reusable.yml
```diff
@@ -0,0 +1,62 @@
...
+      uses: shivammathur/setup-php@9e72090525849c5e82e596468b86eb55e9cc5401 # v2.32.0
+      with:
+        php-version: ${{ inputs.php-version }}
+        coverage: none
+        tools: phpunit-polyfills, composer:v2
+
+    - name: Install dependencies
+      run: composer update -W
+
+    - name: Install WordPress
```
dkotter, line 53: If desired, we could utilize `wp-env` to spin up a test environment which would simplify things here (and remove the need for the bin script). If helpful, can see how we run PHPUnit on the AI plugin: https://github.com/WordPress/ai/blob/813c81ddfa7807f628dda5fcf3df75314ff8602c/.github/workflows/test.yml#L169

.github/workflows/php-wp-integration-reusable.yml
peterwilsoncc: Action versions updated in 65282953c9fb9cbaa51d72704e449718f381a89d.

.github/workflows/php-wp-integration-reusable.yml
peterwilsoncc: While the initial PR is more complex, I was hoping to keep the time taken for each test run down by avoiding a Docker pull, etc. I'm happy to switch if you have a strong preference.

dkotter: Nope, works for me

dkotter (dismissed):

peterwilsoncc:
@dkotter I've wrapped up what I think is a good start.

dkotter (approved):

